PR #21, opened 2019-02-12
Fix the MSAuthHelper on macOS to trace ADAL logs and return correct exit codes

### LINKED ISSUE #20: Mac Microsoft Authentication helper should capture ADAL logs
The Windows Microsoft Authentication helper currently captures the ADAL logging information when `GCM_TRACE=1`. The macOS version currently does not. We should ensure parity here.

mjcheetham:
Enable the Microsoft Authentication helper on macOS to include ADAL logging information in the trace output when GCM_TRACE is set. Also respects PII or sensitive information tracing setting GCM_TRACE_SECRETS.

Upgrade to ADAL 4.0.0 which has now fixed the problem of being noisy and writing to NSLog/stderr.

Update the AHLogger trace format output to be consistent with that of the main GCM application and Git's own output.

Correctly return a non-zero exit code and error message on failure.

Fixes #20


nickgra (approved):

jrbriggs (approved):
Looks good, thanks!

PR #54, opened 2019-06-26
Add ability to read Git config via libgit2

mjcheetham:
Use libgit2 to read Git configuration files and call other helpful methods such as locating config files and root repository directories.

LibGit2Sharp was not used as we only need a minimal subset of the APIs. P/Invoke wrapping the native libgit2 C functions was created instead.

The motivation for reading from Git configuration is to be able to extract things such as the http.proxy setting, coming in a later PR.


src/shared/Microsoft.Git.CredentialManager/Microsoft.Git.CredentialManager.csproj
```diff
@@ -7,10 +7,15 @@
...
     <LangVersion>latest</LangVersion>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'net461'">
     <Reference Include="System.Net.Http" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="LibGit2Sharp.NativeBinaries" Version="2.0.278" />
```
mjcheetham: @jrbriggs we'll want to clear this dependency with compliance.
We may need to include the license file and a copy of the source?

src/shared/Microsoft.Git.CredentialManager.Tests/Interop/U8StringConverterTests.cs
```diff
@@ -0,0 +1,121 @@
...
+         *
+         * Snowman without snow
+         *   ⛄
+         *   Codepoint : U+26C4
+         *   UTF-16    : 0x26C4
+         *   UTF-8     : 0xE2 0x9B 0x84
+         */
+
+        // "Unicode 😅 ıs awesome! Ω ⛄";
+        private const string ComplexString = "Unicode \uD83D\uDE05 \u0131s awesome! \u03A9 \u26C4";
```
derrickstolee: Is there something about .NET that prevents you from putting the emoji literally as part of the string contents?

src/shared/Microsoft.Git.CredentialManager/IGit.cs
```diff
@@ -0,0 +1,85 @@
...
+    public interface IGitConfiguration : IDisposable
+    {
+        /// <summary>
+        /// Path to repository if this configuration object is also scoped to a repository, otherwise null.
+        /// </summary>
+        string RepositoryPath { get; }
+
+        /// <summary>
+        /// Get the value of a configuration entry as a string, or null if not found.
+        /// Equivalent to <seealso cref="GetString"/>.
```
derrickstolee: If this is an equivalent, then can we make one of them an extension method so we don't need to duplicate the redirection in every implementation?

derrickstolee: Since you already have extension methods below, perhaps `.GetString(string name)` can be in there, too.

src/shared/Microsoft.Git.CredentialManager/Interop/LibGit2.cs
```diff
@@ -0,0 +1,154 @@
...
+            GC.SuppressFinalize(this);
+        }
+
+        ~LibGit2()
+        {
+            Dispose(false);
+        }
+    }
+
+    internal class LibGit2Configuration : IGitConfiguration
```
derrickstolee: Is it part of your style to have multiple (non-nested) classes in a single file?

src/shared/Microsoft.Git.CredentialManager/Interop/LibGit2.cs
```diff
@@ -0,0 +1,154 @@
...
+            _config = config;
+            RepositoryPath = repositoryPath;
+
+            // Create snapshot for reading values
+            _snapshot = CreateSnapshot(_config);
+        }
+
+        public string RepositoryPath { get; }
+
+        public string this[string name] => GetString(name);
```
derrickstolee: With my suggestion to make `GetString()` an extension method, the redirection would be reversed here.

src/shared/Microsoft.Git.CredentialManager/Interop/LibGit2.cs
```diff
@@ -0,0 +1,154 @@
...
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        ~LibGit2Configuration()
+        {
+            Dispose(false);
+        }
+
+        private static unsafe git_config* CreateSnapshot(git_config* config)
```
derrickstolee: Do we need protection from calling this multiple times? It appears we do not keep a list or stack of snapshots.

src/shared/Microsoft.Git.CredentialManager/Interop/LibGit2.cs
```diff
@@ -0,0 +1,154 @@
...
+        public string this[string name] => GetString(name);
+
+        public unsafe string GetString(string name)
+        {
+            int result = git_config_get_string(out string value, _snapshot, name);
+
+            switch (result)
+            {
+                case GIT_ENOTFOUND:
+                    return null;
```
derrickstolee: I appreciate that `config["key"]` will not throw a `KeyNotFoundException` like a dictionary would. We'll need to be careful about assuming default values when we see a null or empty on read.

src/shared/Microsoft.Git.CredentialManager.Tests/Interop/U8StringConverterTests.cs
derrickstolee: I do appreciate that you have these unit tests!

src/shared/Microsoft.Git.CredentialManager/Interop/Posix/Native/Stdlib.cs
```diff
@@ -0,0 +1,11 @@
+using System;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.Git.CredentialManager.Interop.Posix.Native
+{
+    public static class Stdlib
+    {
+        [DllImport("libc", CallingConvention = CallingConvention.Cdecl, CharSet = CharSet.Ansi)]
+        public static extern unsafe IntPtr realpath(byte* path, byte* resolved_path);
```
jamill: How are we doing error handling here? I only made it through the test code, but it looks like we are not checking the actual error? (Either with `SetLastError` or `Marshal.GetLastWin32Error`)?

src/shared/Microsoft.Git.CredentialManager.Tests/Interop/LibGit2Tests.cs
```diff
@@ -0,0 +1,201 @@
...
+
+            Assert.Equal(realExpected, realActual);
+        }
+
+        private static string RealPath(string path)
+        {
+            if (PlatformUtils.IsPosix())
+            {
+                byte[] pathBytes = Encoding.UTF8.GetBytes(path);
+                byte[] resolvedBytes = new byte[64 * 1024];
```
jamill: Is there a `PATH_MAX` we can use here for a maximum value (from a quick look at `realpath` documentation, it looks like it might be platform specific?)

src/shared/Microsoft.Git.CredentialManager.Tests/Interop/LibGit2Tests.cs
```diff
@@ -0,0 +1,201 @@
...
+                            return null;
+                        }
+                    }
+                }
+
+                int nullByteIndex = Array.IndexOf(resolvedBytes, (byte) '\0');
+                return Encoding.UTF8.GetString(resolvedBytes, 0, nullByteIndex);
+            }
+
+            if (PlatformUtils.IsWindows())
```
jamill: nit: `else if`?

src/shared/Microsoft.Git.CredentialManager.Tests/Interop/U8StringConverterTests.cs
mjcheetham: > Is there something about .NET that prevents you from putting the emoji literally as part of the string contents?

Nothing about .NET prevents me, but this is a precaution against _certain_ IDEs, editors, and version control systems from messing up the file encoding. At least this way the tests will be working with stable data.

src/shared/Microsoft.Git.CredentialManager/Interop/LibGit2.cs
```diff
@@ -0,0 +1,154 @@
...
+            }
+            finally
+            {
+                git_buf_dispose(buf);
+            }
+        }
+
+        private void Dispose(bool disposing)
+        {
+            git_libgit2_shutdown();
```
jeschu1: Should we be checking the value of disposing?

src/shared/Microsoft.Git.CredentialManager.Tests/Interop/U8StringConverterTests.cs
```diff
@@ -0,0 +1,121 @@
...
+            unsafe
+            {
+                byte* actual = (byte*) U8StringConverter.ToNative(ComplexString);
+
+                AssertCStringEqual(ComplexUtf8,  actual);
+            }
+        }
+
+        [Fact]
+        public void U8StringConverter_ToManaged_SimpleString_ReturnsExpectedString()
```
jeschu1: Is it work validating the behavior when you send in U8StringConverter.ToManager(null) or U8StringConverter.ToNative(null)?

src/shared/Microsoft.Git.CredentialManager/Interop/LibGit2.cs
mjcheetham: Currently most of the reason for the existence of `class LibGit2` this file is for getting configuration. If and once the class gets larger in scope, then breaking it up would be worth it 😃

src/shared/Microsoft.Git.CredentialManager/Interop/LibGit2.cs
mjcheetham: I'll just remove the indexer I think.

src/shared/Microsoft.Git.CredentialManager/Interop/LibGit2.cs
mjcheetham: I've changed the design here. Instead we now have a `TryGetString` method which is on the interface, and extension methods for `GetString` (and the overloads).

src/shared/Microsoft.Git.CredentialManager/IGit.cs
mjcheetham: I've changed the design here. Instead we now have a `TryGetString` method which is on the interface, and extension methods for `GetString` (and the overloads).

src/shared/Microsoft.Git.CredentialManager.Tests/Interop/LibGit2Tests.cs
mjcheetham: Not required since all code paths from the above `if` block `return`. I actually get a warning that the `else` here is redundant.

src/shared/Microsoft.Git.CredentialManager/Interop/LibGit2.cs
mjcheetham: `disposing` is only useful for disposal of managed code.
The pattern is:

```csharp
protected virtual void Dispose(bool disposing)
{
    if (disposing)
    {
        FreeManagedResources();
    }

    // This is always run, whether disposing or finalising.
    FreeUnmanagedResources();
}

public void Dispose()
{
    Dispose(true);
    GC.SuppressFinalize(this);
}

~ClassName()
{
    Dispose(false);
}
```

I.e., if _disposing_ then clean up everything. If _finalising_ then only clean up unmanaged resources.

mjcheetham:
> I see us adding the ability to read, but I don't see where we actually use the config.
> 
> Did I miss something or will that be in a followup PR?

See the PR description: https://github.com/microsoft/Git-Credential-Manager-Core/pull/54#issue-291918125

> The motivation for reading from Git configuration is to be able to extract things such as the http.proxy setting, coming in a later PR.

We will use it when I bring the proxy support PR; reading http.proxy.

src/shared/Microsoft.Git.CredentialManager.Tests/Interop/U8StringConverterTests.cs
mjcheetham: Not much _work_, and definitely _worth_ it! Adding...

src/shared/Microsoft.Git.CredentialManager/Interop/LibGit2.cs
mjcheetham: This method is only used from the constructor, we always create one snapshot and store it in the object as a field (and clean it up after).

To avoid this from being introduced, I'll just inline the `git_config_snapshot` call.

src/shared/Microsoft.Git.CredentialManager/Interop/Posix/Native/Stdlib.cs
mjcheetham: We are not. Not sure `Marshal.GetLastWin32Error` will work here on this POSIX call.

This is really only used in tests, and then only to resolve the symlinked temporary directory on macOS because.. Apple.. and there is no .NET API exposing 'realpath' because.. Microsoft. 😢

```
lrwxr-xr-x@ 1 root  wheel  11 31 Jan 23:02 /var -> private/var
```

I'm tempted to just remove this, the tests it supports (testing of `LibGit2.GetRepositoryPath` which just wraps `git_repository_discover`) because of this mess with symlinks 😢

src/shared/Microsoft.Git.CredentialManager.Tests/Interop/LibGit2Tests.cs
mjcheetham: There is, and yes it's platform specific.

This method exists only in the specific LibGit2 tests and only used in here. In practice it's only going to be `<tmp-base>/repo-12345678/<1-10 long>` in total length.

If someone decides to move the temporary location to more than 64k characters deep, then they can fix the test. 😄

.azure-pipelines/templates/windows/compile.yml
```diff
@@ -31,7 +31,6 @@ steps:
     configuration: 'Windows$(configuration)'
     otherConsoleOptions: '/Framework:.NETCoreApp,Version=2.1'
     testRunTitle: 'Unit tests - common (Windows)'
-  continueOnError: true
```
mjcheetham: This was hiding test failures!

src/shared/Microsoft.Git.CredentialManager.Tests/Interop/LibGit2Tests.cs
mjcheetham: I decided to use the returned pointer from `realpath` rather than the other 'output' mechanism via user allocated memory.

`realpath` returns a null pointer on error, or the resolved `char *` path. On success it also writes the resolved path to the user supplied `char *resolved_path`.

PR #55, opened 2019-06-26
Cache environment variables dictionary

mjcheetham:
Cache the environment variables in a dictionary after first use and don't always reconstruct it per access.

This is also work which will be used for upcoming PRs to combine the environment variables and Git configuration layers into a 'settings' layer in GCM.

### DIFF
--- a/src/shared/Microsoft.Git.CredentialManager/CommandContext.cs
+++ b/src/shared/Microsoft.Git.CredentialManager/CommandContext.cs
@@ -3,6 +3,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Collections.ObjectModel;
 using System.IO;
 using System.Text;
 using Microsoft.Git.CredentialManager.Interop.MacOS;
@@ -55,7 +56,7 @@ namespace Microsoft.Git.CredentialManager
         /// Access the environment variables for the current GCM process.
         /// </summary>
         /// <returns>Set of all current environment variables.</returns>
-        IReadOnlyDictionary<string, string> GetEnvironmentVariables();
+        IReadOnlyDictionary<string, string> EnvironmentVariables { get; }
     }
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


src/shared/Microsoft.Git.CredentialManager/CommandContext.cs
```diff
@@ -132,33 +134,39 @@ public ITerminal Terminal
...
+                        if (key is string name && variables[key] is string value)
+                        {
+                            _envars[name] = value;
+                        }
+                    }
                 }
-            }
 
-            return result;
+                return new ReadOnlyDictionary<string, string>(_envars);
```
derrickstolee: You seem to be populating a new `ReadOnlyDictionary` with every call, even though `_envars` is not changing. Do you want to cache the `ReadOnlyDictionary` instead?

src/shared/Microsoft.Git.CredentialManager/CommandContext.cs
```diff
@@ -132,33 +134,39 @@ public ITerminal Terminal
...
+                    // See this issue for more information: https://github.com/dotnet/corefx/issues/13146
 
-            // If we're on the Windows platform we should de-duplicate by setting the string
-            // comparer to OrdinalIgnoreCase.
-            var comparer = PlatformUtils.IsWindows()
-                         ? StringComparer.OrdinalIgnoreCase
-                         : StringComparer.Ordinal;
+                    // If we're on the Windows platform we should de-duplicate by setting the string
+                    // comparer to OrdinalIgnoreCase.
+                    var comparer = PlatformUtils.IsWindows()
```
jeschu1: Is there an option to make Mac non-case sensitive for env variables?

src/shared/Microsoft.Git.CredentialManager/CommandContext.cs
mjcheetham: Good point! Fixed.

src/shared/Microsoft.Git.CredentialManager/CommandContext.cs
mjcheetham: As far as I know.. there is no way.

PR #60, opened 2019-07-10
Introduce ISettings component to consolidate envar and Git cfg settings

mjcheetham:
Introduce a new component `ISettings` that provides an interface for looking up the value of various GCM settings that can be specified from either the environment or Git configuration.

### DIFF
--- a/src/shared/Microsoft.Git.CredentialManager.Tests/InputArgumentsTests.cs
+++ b/src/shared/Microsoft.Git.CredentialManager.Tests/InputArgumentsTests.cs
@@ -62,5 +62,80 @@ namespace Microsoft.Git.CredentialManager.Tests
             Assert.Equal("bar", inputArgs["foo"]);
             Assert.Equal("bar", inputArgs.GetArgumentOrDefault("foo"));
         }
+
+        [Fact]
+        public void InputArguments_GetRemoteUri_NoAuthority_ReturnsNull()
+        {
+            var dict = new Dictionary<string, string>();
+
+            var inputArgs = new InputArguments(dict);
+
+            Uri actualUri = inputArgs.GetRemoteUri();
+
+            Assert.Null(actualUri);
+        }
+
+        [Fact]
+        public void InputArguments_GetRemoteUri_Authority_ReturnsUriWithAuthority()
+        {
+            var expectedUri = new Uri("https://example.com/");
+
+            var dict = new Dictionary<string, string>
+            {
+                ["protocol"] = "https",
+                ["host"]     = "example.com"
+            };
+
+            var inputArgs = new InputArguments(dict);
+
+            Uri actualUri = inputArgs.GetRemoteUri();
+
+            Assert.NotNull(actualUri);
+            Assert.Equal(expectedUri, actualUri);
+        }
+
+        [Fact]
+        public void InputArguments_GetRemoteUri_AuthorityPath_ReturnsUriWithAuthorityAndPath()
+        {
+            var expectedUri = new Uri("https://example.com/an/example/path");
+
+            var dict = new Dictionary<string, string>
+            {
+                ["protocol"] = "https",
+                ["host"]     = "example.com",
+                ["path"]     = "an/example/path"
+            };
+
+            var inputArgs = new InputArguments(dict);
+
+            Uri actualUri = inputArgs.GetRemoteUri();
+
+            Assert.NotNull(actualUri);
+            Assert.Equal(expectedUri, actualUri);
+        }
+
+        [Fact]
+        public void InputArguments_GetRemoteUri_AuthorityPathUserInfo_ReturnsUriWithAuthorityAndPath()
+        {
+            var expectedUri = new Uri("https://example.com/an/example/path");
+
+            var dict = new Dictionary<string, string>
+            {
+                ["protocol"] = "https",
+                ["host"]     = "example.com",
+                ["path"]     = "an/example/path",
+
+                // Username and password are not expected to appear in the returned URI
+                ["username"] = "john.doe",
+                ["password"] = "password123"
+            };
+
+            var inputArgs = new InputArguments(dict);
+
+            Uri actualUri = inputArgs.GetRemoteUri();
+
+            Assert.NotNull(actualUri);
+            Assert.Equal(expectedUri, actualUri);
+        }
     }
 }
--- a/src/shared/Microsoft.Git.CredentialManager.Tests/UriExtensionsTests.cs
+++ b/src/shared/Microsoft.Git.CredentialManager.Tests/UriExtensionsTests.cs
@@ -1,0 +1,32 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+using System;
+using System.Linq;
+using Xunit;
+
+namespace Microsoft.Git.CredentialManager.Tests
+{
+    public class UriExtensionsTests
+    {
+        [Theory]
+        [InlineData("http://com")]
+        [InlineData("http://example.com",
+            "http://example.com")]
+        [InlineData("http://foo.example.com",
+            "http://foo.example.com", "http://example.com")]
+        [InlineData("http://example.com/foo",
+            "http://example.com/foo", "http://example.com")]
+        [InlineData("http://example.com/foo/",
+            "http://example.com/foo", "http://example.com")]
+        [InlineData("http://example.com/foo?query=true#fragment",
+            "http://example.com/foo", "http://example.com")]
+        [InlineData("http://buzz.foo.example.com/bar/baz",
+            "http://buzz.foo.example.com/bar/baz", "http://buzz.foo.example.com/bar", "http://buzz.foo.example.com", "http://foo.example.com", "http://example.com")]
+        public void UriExtensions_GetGitConfigurationScopes(string start, params string[] expected)
+        {
+            var startUri = new Uri(start);
+            string[] actual = UriExtensions.GetGitConfigurationScopes(startUri).ToArray();
+            Assert.Equal(expected, actual);
+        }
+    }
+}
--- a/src/shared/Microsoft.Git.CredentialManager/EnvironmentVariables.cs
+++ b/src/shared/Microsoft.Git.CredentialManager/EnvironmentVariables.cs
@@ -1,0 +1,109 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+
+namespace Microsoft.Git.CredentialManager
+{
+    /// <summary>
+    /// Simple wrapper dictionary around environment variables.
+    /// </summary>
+    public interface IEnvironmentVariables : IReadOnlyDictionary<string, string> { }
+
+    public class EnvironmentVariables : IEnvironmentVariables
+    {
+        private readonly IReadOnlyDictionary<string, string> _envars;
+
+        public EnvironmentVariables(IDictionary variables)
+        {
+            EnsureArgument.NotNull(variables, nameof(variables));
+
+            // On Windows it is technically possible to get env vars which differ only by case
+            // even though the general assumption is that they are case insensitive on Windows.
+            // For example, some of the standard .NET types like System.Diagnostics.Process
+            // will fail to start a process on Windows if given duplicate environment variables.
+            // See this issue for more information: https://github.com/dotnet/corefx/issues/13146
+
+            // If we're on the Windows platform we should de-duplicate by setting the string
+            // comparer to OrdinalIgnoreCase.
+            var comparer = PlatformUtils.IsWindows()
+                ? StringComparer.OrdinalIgnoreCase
+                : StringComparer.Ordinal;
+
+            var dict = new Dictionary<string, string>(comparer);
+
+            foreach (var key in variables.Keys)
+            {
+                if (key is string name && variables[key] is string value)
+                {
+                    dict[name] = value;
+                }
+            }
+
+            _envars = dict;
+        }
+
+        #region IReadOnlyDictionary<string, string>
+
+        IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
+        public IEnumerator<KeyValuePair<string, string>> GetEnumerator() => _envars.GetEnumerator();
+        public int Count => _envars.Count;
+        public bool ContainsKey(string key) => _envars.ContainsKey(key);
+
+        /// <summary>
+        /// Try and get the value of an environment variable as a string.
+        /// </summary>
+        /// <param name="name">Environment variable name.</param>
+        /// <param name="value">Environment variable value.</param>
+        /// <returns>True if the value was found, false otherwise.</returns>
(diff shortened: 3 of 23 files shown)


src/shared/Microsoft.Git.CredentialManager/EnvironmentVariables.cs
```diff
@@ -0,0 +1,109 @@
...
+
+            foreach (var key in variables.Keys)
+            {
+                if (key is string name && variables[key] is string value)
+                {
+                    dict[name] = value;
+                }
+            }
+
+            _envars = new ReadOnlyDictionary<string, string>(dict);
```
jrbriggs: Could this just be
```suggestion
            _envars = dict
```
since `Dictionary` implements `IReadOnlyDictionary`?

src/shared/Microsoft.Git.CredentialManager/EnvironmentVariables.cs
```diff
@@ -0,0 +1,109 @@
...
+    public static class EnvironmentVariablesExtensions
+    {
+        /// <summary>
+        /// Get the value of an environment variable as a string.
+        /// </summary>
+        /// <exception cref="System.Collections.Generic.KeyNotFoundException">An environment variable with the specified name was not found.</exception>
+        /// <param name="envars">Environment variables.</param>
+        /// <param name="name">Environment variable name.</param>
+        /// <returns>Environment variable value.</returns>
+        public static string GetValue(this IEnvironmentVariables envars, string name)
```
jrbriggs: The docs say this throws `KeyNotFoundException` when there is no value for `name` but the implementation _looks_ like it returns `null` and doesn't throw. Which is the desired behavior?

src/shared/Microsoft.Git.CredentialManager/StringExtensions.cs
```diff
@@ -25,5 +25,108 @@ public static bool IsTruthy(this string str)
...
+            return ToBooleany(str) ?? defaultValue;
+        }
+
+        /// <summary>
+        /// Trim all characters at the end of the string from the last index of the given character.
+        /// </summary>
+        /// <param name="str">String to trim.</param>
+        /// <param name="c">Character to locate the index of.</param>
+        /// <returns>Trimmed string.</returns>
+        public static string TrimEndIndexOf(this string str, char c)
```
jrbriggs: This and the subsequent `TrimStartIndexOf` don't make it explicit exactly what they do. Maybe `TruncateFromLastIndexOf` and `TruncateUntilFirstIndexOf` or something similar?

jrbriggs: Also add to the docstring whether the character is included or excluded from the result?

src/shared/TestInfrastructure/Objects/TestGitConfiguration.cs
```diff
@@ -1,22 +1,58 @@
...
-        bool IGitConfiguration.TryGetString(string name, out string value) => Values.TryGetValue(name, out value);
+        bool IGitConfiguration.TryGetValue(string name, out string value) => _dict.TryGetValue(name, out value);
 
         void IDisposable.Dispose() { }
 
         #endregion
+
+        #region IDictionary<string, string>
+
+        public IEnumerator<KeyValuePair<string, string>> GetEnumerator() => _dict.GetEnumerator();
```
jrbriggs: I'm not sure I love needing to duplicate the IDictionary interface. Is there a simpler option here?

src/shared/Microsoft.Git.CredentialManager/EnvironmentVariables.cs
mjcheetham: The exception, to be consistent with `IGitConfiguration`. Will fix.

src/shared/Microsoft.Git.CredentialManager/EnvironmentVariables.cs
mjcheetham: 👍

src/shared/Microsoft.Git.CredentialManager/StringExtensions.cs
mjcheetham: :+1:

src/shared/TestInfrastructure/Objects/TestGitConfiguration.cs
mjcheetham: This is just a testing connivence more than anything, to avoid having the `.Dictionary` everywhere.
I could just keep the commonly used methods (`TryGetValue`, `this[string key]`) instead and also expose the internal dictionary for more complicated usages.

src/shared/Microsoft.Git.CredentialManager/Settings.cs
```diff
@@ -0,0 +1,176 @@
...
+            {
+                using (var config = _git.GetConfiguration(repositoryPath))
+                {
+                    if (remoteUri != null)
+                    {
+                        /*
+                         * Look for URL scoped "section" configuration entries, starting from the most specific
+                         * down to the least specific (stopping before the TLD).
+                         *
+                         * In a divergence from standard Git configuration rules, we also consider matching URL scopes
```
mjcheetham: @jrbriggs, calling out this part of the code again in the latest push.
I realised that GCM Core was doing something 'not consistent with Git', as does GCM Windows.

Git really only expects `[credential]` configuration entries to have URLs that _include the scheme_ such as:

```ini
    useHttpPath = true
    proxy = https://user:pass@proxy.example.com
```

..but the initial code was only looking for entries without the scheme.. i.e,

```ini
[credential "dev.azure.com"]
    useHttpPath = true
    proxy = https://user:pass@proxy.example.com
```

GCM Windows seems to have introduced the idea of omitting the scheme, probably so that you can specify options for multiple protocols at the same time, such as "http" and "https".

I have adjusted (and called out the divergence) the code here to look for both the scheme- and scheme-less- variants so that we can be consistent with Git's logic first, and then also pick up the behaviour of GCM Windows second.

---
**Side note:**
Looks like GCM Windows introduced the `credential.proxy` configuration entry and the `GCM_HTTP_PROXY` environment variable. Git/cURL have their own configuration of `http.proxy` and `HTTP_PROXY`. GCM Windows seems to support `GCM_HTTP_PROXY` but emit a deprecation warning message.

This is something I'll need to keep in mind in the next PR when reading proxy config, and trying to stay compat with existing GCM Windows users' setups.

src/shared/Microsoft.Git.CredentialManager.Tests/InputArgumentsTests.cs
```diff
@@ -62,5 +62,78 @@ public void InputArguments_OtherArguments()
...
+            {
+                ["protocol"] = "https",
+                ["host"]     = "example.com"
+            };
+
+            var inputArgs = new InputArguments(dict);
+
+            Uri actualUri = inputArgs.GetRemoteUri();
+
+            Assert.NotNull(actualUri);
```
jeschu1: Curious here, why do a NotNull then an Equal check.  Would the equal check be sufficient?

src/shared/Microsoft.Git.CredentialManager.Tests/InputArgumentsTests.cs
```diff
@@ -62,5 +62,78 @@ public void InputArguments_OtherArguments()
...
+            Assert.Equal(expectedUri, actualUri);
+        }
+
+        [Fact]
+        public void InputArguments_GetRemoteUri_AuthorityPathUserInfo_ReturnsUriWithAuthorityAndPath()
+        {
+            var expectedUri = new Uri("https://example.com/an/example/path");
```
jeschu1: Maybe a small comment saying username and password should be ignored in remoteUri construction?  I beleive thats the only difference between this test and the previous one.

src/shared/Microsoft.Git.CredentialManager.Tests/InputArgumentsTests.cs
mjcheetham: For me it makes the error reporting slightly nicer in case of failure. :)

src/shared/Microsoft.Git.CredentialManager/Settings.cs
jrbriggs: @mjcheetham thanks for calling that out. I think having back-compat makes sense. Should we add a unit test for `GetGitConfigurationScopes` with scheme-less URIs?

src/shared/Microsoft.Git.CredentialManager/EnvironmentVariables.cs
```diff
@@ -0,0 +1,109 @@
...
+            // even though the general assumption is that they are case insensitive on Windows.
+            // For example, some of the standard .NET types like System.Diagnostics.Process
+            // will fail to start a process on Windows if given duplicate environment variables.
+            // See this issue for more information: https://github.com/dotnet/corefx/issues/13146
+
+            // If we're on the Windows platform we should de-duplicate by setting the string
+            // comparer to OrdinalIgnoreCase.
+            var comparer = PlatformUtils.IsWindows()
+                ? StringComparer.OrdinalIgnoreCase
+                : StringComparer.Ordinal;
```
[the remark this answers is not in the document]
mjcheetham: That is correct. This is the best we can do on Windows to avoid this situation, which only happens in obscure circumstances and only if invoked from Git (which can play fast-and-loose with the PEB).

src/shared/Microsoft.Git.CredentialManager/IGit.cs
```diff
@@ -18,7 +18,7 @@ public interface IGitConfiguration : IDisposable
         /// <param name="name">Configuration entry name.</param>
         /// <param name="value">Configuration entry value.</param>
         /// <returns>True if the value was found, false otherwise.</returns>
-        bool TryGetString(string name, out string value);
+        bool TryGetValue(string name, out string value);
```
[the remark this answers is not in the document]
mjcheetham: Just for consistency with `IEnvironmentVariables` and all the dictionary-like types :)

src/shared/Microsoft.Git.CredentialManager.Tests/UriExtensionsTests.cs
```diff
@@ -0,0 +1,30 @@
...
+            "http://example.com")]
+        [InlineData("http://foo.example.com",
+            "http://foo.example.com", "http://example.com")]
+        [InlineData("http://example.com/foo",
+            "http://example.com/foo", "http://example.com")]
+        [InlineData("http://example.com/foo?query=true#fragment",
+            "http://example.com/foo", "http://example.com")]
+        [InlineData("http://buzz.foo.example.com/bar/baz",
+            "http://buzz.foo.example.com/bar/baz", "http://buzz.foo.example.com/bar", "http://buzz.foo.example.com", "http://foo.example.com", "http://example.com")]
+        public void UriExtensions_GetGitConfigurationScopes(string start, params string[] expected)
```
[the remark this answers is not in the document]
mjcheetham: It should return the same, as if the trailing slash was not there. I will add a test and confirm.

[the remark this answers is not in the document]
mjcheetham: Good catch @jeschu1! This case fails. Found a bug. Fixing now..

src/shared/TestInfrastructure/Objects/TestGit.cs
```diff
@@ -8,36 +8,73 @@ namespace Microsoft.Git.CredentialManager.Tests.Objects
...
+
+            return new TestGitConfiguration(repositoryPath, mergedConfigDict);
+        }
 
         string IGit.GetRepositoryPath(string path) =>
             Repositories.Keys.FirstOrDefault(path.StartsWith);
 
         #endregion
+
+        private static IDictionary<string, string> MergeDictionaries(params IDictionary<string, string>[] dictionaries)
```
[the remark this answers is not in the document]
mjcheetham: By duplicate variables do you mean ones that differ only by case?
Because the default dictionary constructor here is does case-sensitive hashing, we will allow multiple entries here.

@dscho are Git configuration entries case-sensitive or case-insensitive? For example are `git config http.proxy` and `git config http.PROXY` the same?

src/shared/Microsoft.Git.CredentialManager/Settings.cs
mjcheetham: > Should we add a unit test for `GetGitConfigurationScopes` with scheme-less URIs?

Do you mean a test like `GetGitConfigurationScopes(new Uri("no-scheme.com"))`?
A `System.Uri` _must_ have a scheme.

If you mean a test of the `Settings.TryGetSetting` and `Settings.GetSettingValues` methods then that already exists with test `Settings_GetSettingValues_EnvarAndMultipleConfig_ReturnsAllWithCorrectPrecedence`.

We create a Git configuration with entries containing no scope, a scope with a scheme, and a scope without a scheme (and an environment variable).

```csharp
public void Settings_GetSettingValues_EnvarAndMultipleConfig_ReturnsAllWithCorrectPrecedence()
{
    const string repositoryPath = "/tmp/repos/foo/.git";
    const string remoteUrl = "http://example.com/foo.git";
    const string scope1 = "http://example.com";
    const string scope2 = "example.com";
    const string envarName = "GCM_TESTVAR";
    const string section = "gcmtest";
    const string property = "bar";
    var remoteUri = new Uri(remoteUrl);

    const string value1 = "First value";
    const string value2 = "Second value";
    const string value3 = "Third value";
    const string value4 = "Last value";

    string[] expectedValues = {value1, value2, value3, value4};

    var envars = new EnvironmentVariables(new Dictionary<string, string>
    {
        [envarName] = value1,
    });
    var git = new TestGit();
    git.AddRepository(repositoryPath, new Dictionary<string, string>
    {
        [$"{section}.{scope1}.{property}"] = value2,
        [$"{section}.{scope2}.{property}"] = value3,
        [$"{section}.{property}"]          = value4
    });

    var settings = new Settings(envars, git);
    string[] actualValues = settings.GetSettingValues(repositoryPath, remoteUri, envarName, section, property).ToArray();

    Assert.Equal(expectedValues, actualValues);
}
```

src/shared/Microsoft.Git.CredentialManager.Tests/UriExtensionsTests.cs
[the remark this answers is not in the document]
mjcheetham: Fixed! Needed to trim both end and start of the path first in the implementation.

src/shared/TestInfrastructure/Objects/TestGit.cs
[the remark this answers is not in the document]
dscho: Git configuration keys are case-insensitive, at least partially so. Therefore: yes, `http.proxy` and `http.PROXY` are considered equivalent.

Having said that, there is an exception to this rule: if you have more than one period in the key, the part between the first and the last one is _case-sensitive_.

For example, `http.https://GitHub.com.proxy` does _not_ refer to the same entry as `http.https://github.com.proxy`. Yes, I know. I know. Oh my, do I know...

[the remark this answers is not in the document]
mjcheetham: > Yes, I know. I know. Oh my, do I know...

I want to cry... 😢

PR #61, opened 2019-07-10
Add ability to override host provider from configuration

### LINKED ISSUE #59: Add support for overriding the selected provider
**Feature description**

In GCM Windows it is possible to set the chosen authority to any of the available options, rather than the 'auto' feature which uses the standard input dictionary from Git to determine the correct one.

This is a useful feature to be able to specify and support GitHub Enterprise customers who wish to have their servers detected as GitHub and not basic/services by the generic provider.

The existing configuration for GCM Windows users is...
```
[credential "mygheserver.com"]
    authority = GitHub
```

We should either take the same configuration values, or create our own key `credential.host/provider` whilst taking into account the legacy 'authority' values.

The `HostProviderRegistry` would be the appropriate component to perform this check and lookup.

mjcheetham:
Check the Git configuration for a specified host provider for the current repository remote URI.

We look for the `credential.authority` configuration entry (and any URL scoped variants) for backwards compatibility with older GCMs, as well as the new `credential.provider` entry.

Also provides a way for various on-prem providers to have instances specified manually (such as GHE).

Fixes #59

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -22,4 +22,73 @@ For the complete list of settings GCM Core understands, see the list below.
 
 ## Available settings
 
-_Currently there are no available settings. This is a placeholder for the settings on the backlog._
+### credential.provider
+
+Define the host provider to use when authenticating.
+
+ID|Provider
+-|-
+`auto` _(default)_|_\[automatic\]_
+`azure-repos`|Azure Repos
+`github`|GitHub
+`generic`|Generic (any other provider not listed above)
+
+Automatic provider selection is based on the remote URL.
+
+This setting is typically used with a scoped URL to map a particular set of remote URLs to providers, for example to mark a host as a GitHub Enterprise instance.
+
+#### Example
+
+```shell
+git config --global credential.ghe.contoso.com.provider github
+```
+
+**Also see: [GCM_PROVIDER](environment.md#GCM_PROVIDER)**
+
+---
+
+### credential.authority _(deprecated)_
+
+> This setting is deprecated and should be replaced by `credential.provider` with the corresponding provider ID value.
+>
+> Click [here](https://aka.ms/gcmcore-authority) for more information.
+
+Select the host provider to use when authenticating by which authority is supported by the providers.
+
+Authority|Provider(s)
+-|-
+`auto` _(default)_|_\[automatic\]_
+`msa`, `microsoft`, `microsoftaccount`,<br/>`aad`, `azure`, `azuredirectory`,</br>`live`, `liveconnect`, `liveid`|Azure Repos<br/>_(supports Microsoft Authentication)_
+`github`|GitHub<br/>_(supports GitHub Authentication)_
+`basic`, `integrated`, `windows`, `kerberos`, `ntlm`,<br/>`tfs`, `sso`|Generic<br/>_(supports Basic and Windows Integrated Authentication)_
+
+#### Example
+
+```shell
+git config --global credential.ghe.contoso.com.authority github
+```
+
+**Also see: [GCM_AUTHORITY](environment.md#GCM_AUTHORITY-deprecated)**
+
+---
+
+### credential.allowWindowsAuth
+
+Allow detection of Windows Integrated Authentication (WIA) support for generic host providers. Setting this value to `false` will prevent the use of WIA and force a basic authentication prompt when using the Generic host provider.
+
+**Note:** WIA is only supported on Windows.
+
+**Note:** WIA is an umbrella term for NTLM and Kerberos (and Negotiate).
+
+Value|WIA detection
+-|-
+`true` _(default)_|Permitted
+`false`|Not permitted
+
+#### Example
+
+```shell
+git config --global credential.tfsonprem123.allowWindowsAuth false
+```
+
+**Also see: [GCM_ALLOW_WINDOWSAUTH](environment.md#GCM_ALLOW_WINDOWSAUTH)**
--- a/src/shared/Microsoft.Git.CredentialManager.Tests/EnumerableExtensionsTests.cs
+++ b/src/shared/Microsoft.Git.CredentialManager.Tests/EnumerableExtensionsTests.cs
@@ -1,0 +1,58 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Xunit;
+
+namespace Microsoft.Git.CredentialManager.Tests
+{
+    public class EnumerableExtensionsTests
+    {
+        [Fact]
+        public void EnumerableExtensions_ConcatMany_Null_ThrowsException()
+        {
+            Assert.Throws<ArgumentNullException>(() => EnumerableExtensions.ConcatMany((IEnumerable<object>)null).ToArray());
+        }
+
+        [Fact]
+        public void EnumerableExtensions_ConcatMany_OneSequence_ReturnsSequence()
+        {
+            int[] seq1 = {0, 1, 2, 3, 4, 4, 4, 4, 5, 6, 7, 8, 9};
+            int[] expectedResult = seq1;
+
+            int[] actualResult = EnumerableExtensions.ConcatMany(seq1).ToArray();
+
+            Assert.Equal(expectedResult, actualResult);
+        }
+
+        [Fact]
+        public void EnumerableExtensions_ConcatMany_TwoSequences_ReturnsConcatenateSequences()
+        {
+            int[] seq1 = {0, 1, 2, 3, 4, 4};
+            int[] seq2 = {4, 4, 5, 6, 7, 8, 9};
+            int[] expectedResult = {0, 1, 2, 3, 4, 4, 4, 4, 5, 6, 7, 8, 9};
+
+            int[] actualResult = EnumerableExtensions.ConcatMany(seq1, seq2).ToArray();
+
+            Assert.Equal(expectedResult, actualResult);
+        }
+
+        [Fact]
+        public void EnumerableExtensions_ConcatMany_ManySequences_ReturnsConcatenateSequences()
+        {
+            int[] seq1 = {0, 1};
+            int[] seq2 = {2, 3};
+            int[] seq3 = {4, 4};
+            int[] seq4 = {4, 4};
+            int[] seq5 = {5, 6};
+            int[] seq6 = {7, 8};
+            int[] seq7 = {9};
+            int[] expectedResult = {0, 1, 2, 3, 4, 4, 4, 4, 5, 6, 7, 8, 9};
+
+            int[] actualResult = EnumerableExtensions.ConcatMany(seq1, seq2, seq3, seq4, seq5, seq6, seq7).ToArray();
+
+            Assert.Equal(expectedResult, actualResult);
+        }
+    }
+}
--- a/src/shared/Microsoft.Git.CredentialManager.Tests/GenericHostProviderTests.cs
+++ b/src/shared/Microsoft.Git.CredentialManager.Tests/GenericHostProviderTests.cs
@@ -10,7 +10,7 @@ using Xunit;
 
 namespace Microsoft.Git.CredentialManager.Tests
 {
-    public class GenericProviderTests
+    public class GenericHostProviderTests
     {
         [Theory]
         [InlineData("http", true)]
@@ -24,7 +24,7 @@ namespace Microsoft.Git.CredentialManager.Tests
         [InlineData("sSh", false)]
         [InlineData("", false)]
         [InlineData(null, false)]
-        public void GenericProvider_IsSupported(string protocol, bool expected)
+        public void GenericHostProvider_IsSupported(string protocol, bool expected)
         {
             var input = new InputArguments(new Dictionary<string, string>
             {
@@ -39,7 +39,7 @@ namespace Microsoft.Git.CredentialManager.Tests
         }
 
         [Fact]
-        public void GenericProvider_GetCredentialKey_ReturnsCorrectKey()
+        public void GenericHostProvider_GetCredentialKey_ReturnsCorrectKey()
         {
             const string expectedKey = "git:https://john.doe@example.com/foo/bar";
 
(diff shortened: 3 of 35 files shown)


src/shared/Microsoft.Git.CredentialManager.Tests/EnumerableExtensionsTests.cs
```diff
@@ -0,0 +1,54 @@
...
+        [Fact]
+        public void EnumerableExtensions_ConcatMany_Null_ThrowsException()
+        {
+            Assert.Throws<ArgumentNullException>(() => EnumerableExtensions.ConcatMany((IEnumerable<object>)null).ToArray());
+        }
+
+        [Fact]
+        public void EnumerableExtensions_ConcatMany_OneSequence_ReturnsSequence()
+        {
+            int[] seq1 = {0, 1, 2, 3, 4, 5, 6, 7, 8, 9};
```
jeschu1: You know I love unit tests :-).  Would it be worth throwing a dup number in the concated sequences for testing?

mjcheetham: Added duplicate values to tests.

src/shared/Microsoft.Git.CredentialManager.Tests/Interop/LibGit2Tests.cs
```diff
@@ -123,7 +123,7 @@ public void LibGit2Configuration_TryGetValue_SectionScopeProperty_NullScope_Retu
             var git = new LibGit2();
             using (var config = git.GetConfiguration(repoPath))
             {
-                bool result = config.TryGetValue("user", null, "name", out string value);
+                bool result = config.TryGetValue("user", (string)null, "name", out string value);
```
jeschu1: Curious why we need (string) before the null

mjcheetham: We don't (any more)!

image

I used to have an overload which wasn't a `string` which I removed now.

docs/configuration.md
```diff
@@ -22,4 +22,48 @@ For the complete list of settings GCM Core understands, see the list below.
...
+### credential.provider
+
+Define the host provider to use when authenticating.
+
+ID|Provider
+-|-
+`auto` _(default)_|_\[automatic\]_
+`azure-repos`|Azure Repos
+`github`|GitHub
+`generic`|Generic
```
jeschu1: What would the basic equivalent be here?

mjcheetham: `generic`.. ish.

GCM Core doesn't support the ability to say "Don't use WIA/NTLM, even if the server supports it"

That should probably be added as an option?

src/shared/Microsoft.Git.CredentialManager/Authentication/WindowsIntegratedAuthentication.cs
```diff
@@ -14,6 +14,12 @@ public interface IWindowsIntegratedAuthentication : IDisposable
 
     public class WindowsIntegratedAuthentication : IWindowsIntegratedAuthentication
     {
+        public static readonly string[] AuthorityIds =
+        {
+            "Integrated", "Windows", "Kerberos", "NTLM",
```
jeschu1: I'm confused what some of these AuthorityIds are.  How did you get this list?

src/shared/Microsoft.Git.CredentialManager/Authentication/MicrosoftAuthentication.cs
```diff
@@ -10,15 +10,22 @@ namespace Microsoft.Git.CredentialManager.Authentication
...
     }
 
     public class MicrosoftAuthentication : AuthenticationBase, IMicrosoftAuthentication
     {
+        public static readonly string[] AuthorityIds =
+        {
+            "MSA",  "Microsoft",   "MicrosoftAccount",
+            "AAD",  "Azure",       "AzureDirectory",
+            "Live", "LiveConnect", "LiveID",
+        };
```
jeschu1: Same question as the other comment, where did you get this list of AuthorityIds from?

src/shared/Microsoft.Git.CredentialManager/Authentication/WindowsIntegratedAuthentication.cs
mjcheetham: These values came from Windows GCM:

https://github.com/microsoft/Git-Credential-Manager-for-Windows/blob/9270c89a0ce109c7a86e9b3d8174a0af56aab468/Shared/Cli/Functions/Common.cs#L433-L474

src/shared/Microsoft.Git.CredentialManager/Authentication/MicrosoftAuthentication.cs
mjcheetham: From the Windows GCM.. these are all the supported values :(

Not sure why we didn't just allow one option per authority, since they all end up returning the same thing!

https://github.com/microsoft/Git-Credential-Manager-for-Windows/blob/9270c89a0ce109c7a86e9b3d8174a0af56aab468/Shared/Cli/Functions/Common.cs#L433-L474

docs/configuration.md
mjcheetham: Added!

docs/configuration.md
```diff
@@ -22,4 +22,71 @@ For the complete list of settings GCM Core understands, see the list below.
...
+
```
jeschu1: You may want to add a description of what we mean by Generic.  It can be something as simple as "If you are using a provider other than Azure Repos or Github, choice Generic"

docs/configuration.md
```diff
@@ -22,4 +22,71 @@ For the complete list of settings GCM Core understands, see the list below.
...
+git config --global credential.ghe.contoso.com.authority github
+```
+
+**Also see: [GCM_AUTHORITY](environment.md#GCM_AUTHORITY-deprecated)**
+
+---
+
+### credential.allowWindowsAuth
+
+Allow detection of Windows Integrated Authentication (WIA) support for generic host providers. Setting this value to `false` will prevent the use of WIA and force a basic authentication prompt when using the Generic host provider.
```
jeschu1: What is the difference between WIA and NTLM?

src/shared/Microsoft.Git.CredentialManager.Tests/GenericHostProviderTests.cs
```diff
@@ -58,20 +58,89 @@ public void GenericProvider_GetCredentialKey_ReturnsCorrectKey()
...
+            {
+                ["protocol"] = "https",
+                ["host"]     = "example.com",
+            });
+
+            const string testUserName = "basicUser";
+            const string testPassword = "basicPass";
+            var basicCredential = new GitCredential(testUserName, testPassword);
+
+            var context = new TestCommandContext
```
jeschu1: worth having a case without user?  That should be the more popular case correct?

src/shared/Microsoft.Git.CredentialManager.Tests/HostProviderRegistryTests.cs
```diff
@@ -57,5 +83,140 @@ public void HostProviderRegistry_MultipleValidProviders_ReturnsFirstRegistered()
...
+
+            var provider1Mock = new Mock<IHostProvider>();
+            var provider2Mock = new Mock<IHostProvider>();
+            var provider3Mock = new Mock<IHostProvider>();
+            provider1Mock.Setup(x => x.Id).Returns("provider1");
+            provider1Mock.Setup(x => x.IsSupported(It.IsAny<InputArguments>())).Returns(true);
+            provider2Mock.Setup(x => x.Id).Returns("provider2");
+            provider2Mock.Setup(x => x.IsSupported(It.IsAny<InputArguments>())).Returns(true);
+            provider3Mock.Setup(x => x.Id).Returns("provider3");
+            provider3Mock.Setup(x => x.IsSupported(It.IsAny<InputArguments>())).Returns(false);
```
jeschu1: Is this false, because of the ProviderOverride?

src/shared/Microsoft.Git.CredentialManager.Tests/HostProviderRegistryTests.cs
```diff
@@ -57,5 +83,140 @@ public void HostProviderRegistry_MultipleValidProviders_ReturnsFirstRegistered()
...
+
+            registry.Register(provider1Mock.Object, provider2Mock.Object, provider3Mock.Object);
+
+            IHostProvider result = registry.GetProvider(input);
```
jeschu1: Probably missing something basic, 3 providers are registered only the last returns false.  Why would we expect to get the last back?

jeschu1: Probably missing something basic, 3 providers are registered only the last returns false.  Why would we expect to get the last back?   Is it because we set provider3 in Settings, hence the false doesn't matter?

src/shared/Microsoft.Git.CredentialManager.Tests/HostProviderRegistryTests.cs
```diff
@@ -57,5 +83,140 @@ public void HostProviderRegistry_MultipleValidProviders_ReturnsFirstRegistered()
...
+            provider1Mock.Setup(x => x.Id).Returns("provider1");
+            provider1Mock.Setup(x => x.IsSupported(It.IsAny<InputArguments>())).Returns(false);
+            provider2Mock.Setup(x => x.Id).Returns("provider2");
+            provider2Mock.Setup(x => x.IsSupported(It.IsAny<InputArguments>())).Returns(true);
+            provider3Mock.Setup(x => x.Id).Returns("provider3");
+            provider3Mock.Setup(x => x.IsSupported(It.IsAny<InputArguments>())).Returns(false);
+
+            registry.Register(provider1Mock.Object, provider2Mock.Object, provider3Mock.Object);
+
+            IHostProvider result = registry.GetProvider(input);
```
jeschu1: So if they specify a provider that doesn't exist, we default to the first true?  Would there be a warning?

src/shared/Microsoft.Git.CredentialManager.Tests/HostProviderRegistryTests.cs
```diff
@@ -57,5 +83,140 @@ public void HostProviderRegistry_MultipleValidProviders_ReturnsFirstRegistered()
...
+
+            Assert.Same(provider2Mock.Object, result);
+        }
+
+        [Fact]
+        public void HostProviderRegistry_GetProvider_AutoLegacyAuthoritySpecified_ReturnsFirstSupportedProvider()
+        {
+            var context = new TestCommandContext
```
jeschu1: Conceptually it would help me if we used actual names (azure, github, etc..) instead of provider1/provider2

jeschu1:
@mjcheetham this is looking good!  I asked a few questions.  

It would be good if we had a test to simulate the popular scenario "Basic" scenario we have today.  Show that that routes to "Generic" with "WindowsAuth" off.  If there is a way to way to validate we get the warning message that would be great too!

src/shared/Microsoft.Git.CredentialManager/HostProviderRegistry.cs
```diff
@@ -46,14 +55,79 @@ public void Register(params IHostProvider[] hostProviders)
...
+            }
+
+            //
+            // Auto-detection
+            //
+            _context.Trace.WriteLine("Performing auto-detection of host provider.");
+            provider = _hostProviders.FirstOrDefault(x => x.IsSupported(input));
 
-            if (provider == null)
+            if (provider is null)
```
jeschu1: nit: Why "is"?

src/shared/Microsoft.Git.CredentialManager/Authentication/BasicAuthentication.cs
```diff
@@ -27,7 +32,7 @@ public GitCredential GetCredentials(string resource, string userName)
 
             EnsureTerminalPromptsEnabled();
 
-            Context.Terminal.WriteLine("Enter credentials for '{0}':", resource);
+            Context.Terminal.WriteLine("Enter basic credentials for '{0}':", resource);
```
jrbriggs: While I like "Enter GitHub credentials", I'm not sure what to think about "Enter basic credentials".

Did you consider simply "Log in to `{0}`" or similar?

src/shared/Microsoft.Git.CredentialManager/Authentication/MicrosoftAuthentication.cs
jrbriggs: Can/should we simplify?

mjcheetham: @jrbriggs
We have no idea which values are being used actively to know which ones to drop.

I we want to _really_ simplify we can just break back-compat and ignore these values. Users would need to manually migrate to specifying the provider.

docs/configuration.md
mjcheetham: WIA: Windows Integrated Authentication
NTLM: NT LAN Manager
[Kerberos](https://en.wikipedia.org/wiki/Kerberos_(protocol)): Name of the MIT designed authentication protocol
Negotiate: Microsoft protocol for selecting either NTLM or Kerberos

WIA refers to any authentication protocols that are integrated into Windows and use your logged in user credentials. Both Kerberos and NTLM (or the 'Negotiate' pseudo-protocol) are types of WIA. Calling it just NTLM isn't really valid since most newer Windows AD domain setups should/will be using Kerberos really.

src/shared/Microsoft.Git.CredentialManager.Tests/HostProviderRegistryTests.cs
mjcheetham: We get the last provider back because it's `provider3` and the test is checking that if we set the override to `provider3` that it will be returned, even if by the 'auto' detection logic `provider3` is NOT valid.

src/shared/Microsoft.Git.CredentialManager.Tests/HostProviderRegistryTests.cs
mjcheetham: We print the following to the trace output:

```
No host provider was found with ID '{providerId}'.. falling back to auto-detection.;
```

Perhaps we should also write to stderr...

src/shared/Microsoft.Git.CredentialManager/HostProviderRegistry.cs
mjcheetham: I've been writing GCM Core with the latest language features where possible. I missed this one.

src/shared/Microsoft.Git.CredentialManager/Authentication/BasicAuthentication.cs
mjcheetham: I did consider this.. but this is similar to the message that Git itself prompts and I wanted to make this distinct (because Git will do a prompt if the helper fails).

This raises an interesting question.. do we want GCM-based TTY prompts to start with a prefix to make it really obvious?

```
warning: Something important to tell you!
Enter your GitHub credentials:
Username: john.doe
Password: ******
```
vs
```
gcm: warning: Something important to tell you!
gcm: Enter your GitHub credentials:
gcm: Username: john.doe
gcm: Password: ******
```

docs/configuration.md
mjcheetham: WIA: Windows Integrated Authentication
NTLM: NT LAN Manager
[Kerberos](https://en.wikipedia.org/wiki/Kerberos_(protocol)): Name of the MIT designed authentication protocol
Negotiate: Microsoft protocol for selecting either NTLM or Kerberos

WIA refers to any authentication protocols that are integrated into Windows and use your logged in user credentials. Both Kerberos and NTLM (or the 'Negotiate' pseudo-protocol) are types of WIA. Calling it just NTLM isn't really valid since most newer Windows AD domain setups should/will be using Kerberos really.

src/shared/Microsoft.Git.CredentialManager.Tests/GenericHostProviderTests.cs
mjcheetham: We can add a case, yes.

> That should be the more popular case correct?

Actually no. When GCM is installed (by default with Git for Windows) it will be getting of all basic prompts first (not Git). I imagine there are more users who use basic authentication rather than those who use it as a workaround for a PAT.

src/shared/Microsoft.Git.CredentialManager.Tests/HostProviderRegistryTests.cs
mjcheetham: We get the last provider back because it's provider3 and the test is checking that if we set the override to provider3 that it will be returned, even if by the 'auto' detection logic provider3 is NOT valid.

mjcheetham:
> It would be good if we had a test to simulate the popular scenario "Basic" scenario we have today. Show that that routes to "Generic" with "WindowsAuth" off.

Look at the test `GenericHostProvider_CreateCredentialAsync_LegacyAuthorityBasic_ReturnsBasicCredentialNoWiaCheck`. This checks that if the `GCM_AUTHORITY` settings is set to `basic` that we do a basic prompt and ignore WIA.

> If there is a way to way to validate we get the warning message that would be great too!

We can add checks for what is written to the trace and stderr streams, but these sorts of assertions tend to be brittle.

docs/configuration.md
mjcheetham: Added a line in the docs explaining that WIA is an umbrella term for NTLM and Kerberos.

docs/configuration.md
mjcheetham: Added.

src/shared/Microsoft.Git.CredentialManager.Tests/HostProviderRegistryTests.cs
mjcheetham: Added stderr warning message.

src/shared/Microsoft.Git.CredentialManager/Authentication/BasicAuthentication.cs
jrbriggs: Makes sense on the messaging. I really like prefixing the lines with who's talking @mjcheetham

src/shared/Microsoft.Git.CredentialManager.Tests/HostProviderRegistryTests.cs
jeschu1: Let me know your thoughts on this one.  Now that I know the meaning really well I certainly won't forget it, but I do think actual names are quicker to digest.

src/shared/Microsoft.Git.CredentialManager/Authentication/BasicAuthentication.cs
mjcheetham: https://github.com/microsoft/Git-Credential-Manager-Core/issues/63

Created an issue to track this idea of prefixing 'who is speaking'.

PR #62, opened 2019-07-10
Use Git's HTTP proxy configuration

### LINKED ISSUE #11: Support HTTP proxies
The .NET Framework GCM supports being used from behind a proxy. We should look at what is required (if anything) to allow these customer scenarios to work with GCM Core.

mjcheetham:
Use the Git HTTP proxy configuration to create new configured `HttpClient` instances with proxy support.

Introduce the `IWebProxyFactory` component that extracts proxy configuration from Git and the environment to create new `IWebProxy` objects.

Fixes #11

### DIFF
--- a/docs/netconfig.md
+++ b/docs/netconfig.md
@@ -1,0 +1,105 @@
+# Network and HTTP configuration
+
+Git Credential Manager Core's network and HTTP(S) behavior can be configured in a few different ways via [environment variables](environment.md) and [configuration options](configuration.md).
+
+## HTTP Proxy
+
+If your computer sits behind a network firewall that requires the use of a proxy server to reach repository remotes or the wider Internet, there are various methods for configuring GCM to use a proxy.
+
+The simplist way to configure a proxy for _all_ HTTP(S) remotes is to [use the standard Git HTTP(S) proxy setting `http.proxy`](https://git-scm.com/docs/git-config#Documentation/git-config.txt-httpproxy).
+
+For example to configure a proxy for all remotes for the current user:
+
+```shell
+git config --global http.proxy http://proxy.example.com
+```
+
+To specify a proxy for a particular remote you can [use the `remote.<name>.proxy` repository-level setting](https://git-scm.com/docs/git-config#Documentation/git-config.txt-remoteltnamegtproxy), for example:
+
+```shell
+git config --local remote.origin.proxy http://proxy.example.com
+```
+
+The advantage to using these standard configuration options is that in addition to GCM being configured to use the proxy, Git itself will be configured at the same time. This is probably the most commonly desired case in environments behind an Internet-blocking firewall.
+
+### Authenticated proxies
+
+Some proxy servers do not accept anonymous connections and require authentication. In order to specify the credentials to be used with a proxy, you can specify the username and password as part of the proxy URL setting.
+
+The format follows [RFC 3986 section 3.2.1](https://tools.ietf.org/html/rfc3986#section-3.2.1) by including the credentials in the 'user information' part of the URI. The password is optional.
+
+```text
+protocol://username[:password]@hostname
+```
+
+For example, to specify the username `john.doe` and the password `letmein123` for the proxy server `proxy.example.com`:
+
+```text
+https://john.doe:letmein123@proxy.example.com
+```
+
+If you have special characters (as defined by [RFC 3986 section 2.2](https://tools.ietf.org/html/rfc3986#section-2.2)) in your username or password such as `:`, `@`, or any other non-URL friendly character you can URL-encode them ([section 2.1](https://tools.ietf.org/html/rfc3986#section-2.2)).
+
+For example, a space character would be encoded with `%20`.
+
+### Other proxy options
+
+GCM Core supports other ways of configuring a proxy for convenience and compatibility.
+
+1. GCM-specific configuration options (_**only** respected by GCM; **deprecated**_):
+   - `credential.httpProxy`
+   - `credential.httpsProxy`
+1. cURL environment variables (_also respected by Git_):
+   - `HTTP_PROXY`
+   - `HTTPS_PROXY`
+   - `ALL_PROXY`
+1. `GCM_HTTP_PROXY` environment variable (_**only** respected by GCM; **deprecated**_)
+
+## TLS Verification
+
+If you are using self-signed TLS (SSL) certificates with a self-hosted host provider such as GitHub Enteprise Server or Azure DevOps Server (previously TFS), you may see the following error message when attempting to connect using Git and/or GCM:
+
+```shell
+$ git clone https://ghe.example.com/john.doe/myrepo
+fatal: The remote certificate is invalid according to the validation procedure.
+```
+
+The **recommended and safest option** is to acquire a TLS certificate signed by a public trusted certificate authority (CA). There are multiple public CAs; here is a non-exhaustive list to consider: [Let's Encrypt](https://letsencrypt.org/), [Comodo](https://www.comodoca.com/), [Digicert](https://www.digicert.com/), [GoDaddy](https://www.godaddy.com/web-security/ssl-certificate), [GlobalSign](https://www.globalsign.com/en/ssl/).
+
+If it is not possible to **obtain a TLS certifiate from a trusted 3rd party** then you should try to add the _specific_ self-signed certificate or one of the CA certificates in the verification chain to your operating system's trusted certificate store ([macOS](https://support.apple.com/en-gb/guide/keychain-access/kyca2431/mac), [Windows](https://blogs.technet.microsoft.com/sbs/2008/05/08/installing-a-self-signed-certificate-as-a-trusted-root-ca-in-windows-vista/)).
+
+If you are _unable_ to either **obtain a trusted certificate**, or trust the self-signed certificate you can disable certificate verification in Git and GCM.
+
+---
+**Security Warning** :warning:
+
+Disabling verification of TLS (SSL) certificates removes protection against a [man-in-the-middle (MITM) attack](https://en.wikipedia.org/wiki/Man-in-the-middle_attack).
+
+Only disable certificate verification if you are sure you need to, are aware of all of the risks, and are unable to trust specific self-signed certificates (as described above).
+
+---
+
+The [environment variable `GIT_SSL_NO_VERIFY`](https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables#_networking) and [Git configuration option `http.sslVerify`](https://git-scm.com/docs/git-config#Documentation/git-config.txt-httpsslVerify) can be used to control TLS (SSL) certifcate verification.
+
+To disable verification for a specific remote (for example <https://example.com>):
+
+```shell
+git config --global http.https://example.com.sslVerify false
+```
+
+To disable verification for the current user for **_all remotes_** (**not recommended**):
+
+```shell
+# Environment variable (Windows)
+SET GIT_SSL_NO_VERIFY=1
+
+# Environment variable (macOS/Linux)
+export GIT_SSL_NO_VERIFY=1
+
+# Git configuration (Windows/macOS/Linux)
+git config --global http.sslVerify false
+```
+
+---
+
+**Note:** You may also experience similar verification errors if you are using a network traffic inspection tool such as [Telerik Fiddler](https://www.telerik.com/fiddler). If you are using such tools please consult their documentation for trusting the proxy root certificates.
--- a/src/shared/Microsoft.Git.CredentialManager.Tests/HttpClientFactoryTests.cs
+++ b/src/shared/Microsoft.Git.CredentialManager.Tests/HttpClientFactoryTests.cs
@@ -1,6 +1,10 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
+using System;
+using System.Net;
 using System.Net.Http;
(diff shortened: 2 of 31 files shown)


src/shared/Microsoft.Git.CredentialManager/Constants.cs
```diff
@@ -17,6 +17,10 @@ public static class EnvironmentVariables
             public const string GcmTraceMsAuth     = "GCM_TRACE_MSAUTH";
             public const string GcmDebug           = "GCM_DEBUG";
             public const string GitTerminalPrompts = "GIT_TERMINAL_PROMPT";
+            public const string CurlAllProxy       = "ALL_PROXY";
+            public const string CurlHttpProxy      = "HTTP_PROXY";
+            public const string CurlHttpsProxy     = "HTTPS_PROXY";
+            public const string GcmHttpProxy       = "GCM_HTTP_PROXY";
```
jeschu1: Should Gcm have a specific Proxy Env variable?  Should it come from git instead?

src/shared/Microsoft.Git.CredentialManager/HttpClientFactory.cs
```diff
@@ -20,17 +22,55 @@ public interface IHttpClientFactory
...
+                {
+                    Proxy = proxy,
+                    UseProxy = true
+                };
+            }
+
+            if (!_settings.IsCertificateVerificationEnabled)
+            {
+#if NETFRAMEWORK
+                ServicePointManager.ServerCertificateValidationCallback = (req, cert, chain, errors) => true;
```
jeschu1: This is interesting, based on the type of .Net we have to do different settings here?  How fun was this one to find :-)?

src/shared/Microsoft.Git.CredentialManager/Settings.cs
```diff
@@ -55,20 +90,109 @@ public Settings(IEnvironmentVariables environmentVariables, IGit git)
...
+                 *        http.proxy
+                 *
+                 *   3. cURL environment variables
+                 *        HTTPS_PROXY
+                 *        HTTP_PROXY
+                 *        ALL_PROXY
+                 *
+                 *   4. Legacy GCM proxy environment variable
+                 *        GCM_HTTP_PROXY
+                 *
```
jeschu1: Ideally how many ways to set the proxy would we want to support/need?

src/shared/Microsoft.Git.CredentialManager/Settings.cs
```diff
@@ -55,20 +90,109 @@ public Settings(IEnvironmentVariables environmentVariables, IGit git)
...
+
+                bool isHttps = StringComparer.OrdinalIgnoreCase.Equals(Uri.UriSchemeHttps, RemoteUri?.Scheme);
+
+                Uri proxyConfig;
+
+                // 1. GCM proxy Git configuration
+                if (isHttps && TryGetUriSetting(null, GitCredCfg.SectionName, GitCredCfg.HttpsProxy, out proxyConfig) ||
+                    TryGetUriSetting(null, GitCredCfg.SectionName, GitCredCfg.HttpProxy,  out proxyConfig))
+                {
+                    return proxyConfig;
```
jeschu1: Is it worth logging that we used the GitCredCfg.HttpsProxy, setting and what the proxy was?  Same question for other 3 code paths.

src/shared/Microsoft.Git.CredentialManager/Settings.cs
```diff
@@ -55,20 +90,109 @@ public Settings(IEnvironmentVariables environmentVariables, IGit git)
...
+                               TryGetUriSetting(KnownEnvars.CurlAllProxy,   null, null, out proxyConfig))
+                {
+                    return proxyConfig;
+                }
+
+                // 4. Legacy GCM proxy environment variable
+                if (TryGetUriSetting(KnownEnvars.GcmHttpProxy, null, null, out proxyConfig))
+                {
+                    // TODO: warning trace?
+                    // the 'GCM_HTTP_PROXY' variable has been deprecated, use '{ keyName }' instead.
```
jeschu1: I like that idea :-)

src/shared/Microsoft.Git.CredentialManager/Trace.cs
```diff
@@ -337,4 +338,28 @@ private static string FormatText(string message, string filePath, int lineNumber
             return text;
         }
     }
+
+    public static class TraceExtensions
+    {
+        public static void WriteProxyInfo(this ITrace trace, IWebProxy proxy)
+        {
+            var uri = proxy.GetProxy(new Uri("https://example.com"));
```
jeschu1: Can you explain what this function is doing?  Sample output in comments would be helpful.

src/shared/Microsoft.Git.CredentialManager/UriExtensions.cs
```diff
@@ -2,11 +2,37 @@
...
+            EnsureArgument.NotNull(uri, nameof(uri));
+            userName = null;
+            password = null;
+
+            if (string.IsNullOrWhiteSpace(uri.UserInfo))
+            {
+                return false;
+            }
+
+            string[] split = uri.UserInfo.Split(new[] {':'}, count: 2);
```
jeschu1: Can we add a comment with an example of the expected format?

src/shared/Microsoft.Git.CredentialManager/HttpClientFactory.cs
```diff
@@ -20,17 +22,55 @@ public interface IHttpClientFactory
...
+            _settings = settings;
+        }
+
         public HttpClient CreateClient()
         {
-            // Initialize a new HttpClient
-            var client = new HttpClient();
+            HttpClientHandler handler;
+
+            if (TryCreateProxy(out IWebProxy proxy))
```
jeschu1: you know what to do here :-)

src/shared/Microsoft.Git.CredentialManager/HttpClientFactory.cs
mjcheetham: A quick Internet search produced some help on how to configure the proxy settings in both the .NET Core and Framework HTTP stacks.

src/shared/Microsoft.Git.CredentialManager/Settings.cs
mjcheetham: Good question.. cc: @jrbriggs

I've gone for maximal compatibility here currently.

src/shared/Microsoft.Git.CredentialManager/UriExtensions.cs
mjcheetham: Added.

src/shared/Microsoft.Git.CredentialManager/Settings.cs
mjcheetham: Added.

src/shared/Microsoft.Git.CredentialManager/Constants.cs
mjcheetham: This is just here for compatibility purposes. We warn about using this deprecated option.

docs/netconfig.md
```diff
@@ -0,0 +1,106 @@
...
+If you have special characters (as defined by [RFC 3986 section 2.2](https://tools.ietf.org/html/rfc3986#section-2.2)) in your username or password such as `:`, `@`, or any other non-URL friendly character you can URL-encode them ([section 2.1](https://tools.ietf.org/html/rfc3986#section-2.2)).
+
+For example, a space character would be encoded with `%20`.
+
+### Other proxy options
+
+GCM Core supports other ways of configuring a proxy for convenience and compatibility.
+
+1. GCM-specific configuration options (_**only** respected by GCM_):
+   - `credential.httpProxy`
```
jeschu1: is it worth mentioning this is being deprecated here too?

src/shared/Microsoft.Git.CredentialManager.Tests/HttpClientFactoryTests.cs
```diff
@@ -22,12 +26,105 @@ public void HttpClientFactory_GetClient_SetsDefaultHeaders()
...
+            {
+                RemoteUri = repoRemoteUri,
+                RepositoryPath = repoPath
+            };
+            var httpFactory = new HttpClientFactory(Mock.Of<ITrace>(), settings, Mock.Of<IStandardStreams>());
+
+            bool result = httpFactory.TryCreateProxy(out IWebProxy proxy);
+
+            Assert.False(result);
+            Assert.Null(proxy);
```
jeschu1: Is it worth verifying any fields on the proxy?

src/shared/Microsoft.Git.CredentialManager.Tests/SettingsTests.cs
```diff
@@ -191,6 +191,266 @@ public void Settings_IsSecretTracingEnabled_EnvarFalsey_ReturnsFalse()
...
+             // Test case 1: cURL environment variables > GCM_HTTP_PROXY
+            envarDict[Constants.EnvironmentVariables.GcmHttpProxy] = value1.ToString();
+            envarDict[Constants.EnvironmentVariables.CurlHttpProxy] = value2.ToString();
+            RunTest(value2);
+
+             // Test case 2: http.proxy > cURL environment variables
+            configDict[$"{Constants.GitConfiguration.Http.SectionName}.{Constants.GitConfiguration.Http.Proxy}"] = value3.ToString();
+            RunTest(value3);
+
+             // Test case 3: credential.httpProxy > http.proxy
```
jeschu1: did we skip 4 cause it's being depricated?  If so I'm fine with that.

src/shared/Microsoft.Git.CredentialManager/CommandContext.cs
```diff
@@ -112,52 +95,7 @@ public CommandContext()
...
-                    {
-                        AutoFlush = true,
-                        NewLine = LineFeed,
-                    };
-                }
-
-                return _stdErr;
-            }
-        }
+        public IStandardStreams Streams { get; }
```
jeschu1: There are many changes for the Streams update.  I really like that you put the changes for this in it's own commit with details.  If you hadn't nicely done that I think would have asked for a separate PR.  But since you did go you :-)!

src/shared/Microsoft.Git.CredentialManager/HttpClientFactory.cs
```diff
@@ -20,17 +23,68 @@ public interface IHttpClientFactory
...
         {
-            // Initialize a new HttpClient
-            var client = new HttpClient();
+            _trace.WriteLine("Creating new HTTP client instance...");
+
+            HttpClientHandler handler;
+
+            if (TryCreateProxy(out IWebProxy proxy))
+            {
+                _trace.WriteLine("HTTP client is using the configured proxy.");
```
jeschu1: To verify do we print out the details of the proxy being using anywhere?  That would be helpful for debugging i think.

src/shared/Microsoft.Git.CredentialManager/Settings.cs
mjcheetham: We log and warn the user that a legacy setting is being used now (including a link to how to configure the supported ways).

docs/netconfig.md
mjcheetham: The line before this says:

> ..for convenience and compatibility.

Perhaps I should make it even more explicit still?

src/shared/Microsoft.Git.CredentialManager.Tests/HttpClientFactoryTests.cs
mjcheetham: There are no other fields on `IWebProxy` sadly (without doing some suspicious casting and/or reflection. 😢

src/shared/Microsoft.Git.CredentialManager/HttpClientFactory.cs
mjcheetham: Yes, we print out the proxy settings used each time a new `HttpClient` instance is created.

mjcheetham: https://github.com/mjcheetham/Git-Credential-Manager-Core/blob/28d9cb68dacfb724414ff876b391e68e2d9792c1/src/shared/Microsoft.Git.CredentialManager/HttpClientFactory.cs#L145

mjcheetham:
> My main thing to double check is that we trace the proxy somewhere.

In the `HttpClientFactory` we trace the proxy information when creating and configuring a new `HttpClient` instance.

https://github.com/mjcheetham/Git-Credential-Manager-Core/blob/28d9cb68dacfb724414ff876b391e68e2d9792c1/src/shared/Microsoft.Git.CredentialManager/HttpClientFactory.cs#L145

> Bonus points if we say which path we used to set it.

Sadly we don't know who called us at this level beyond if the method was a deprecated one or not.

src/shared/Microsoft.Git.CredentialManager.Tests/SettingsTests.cs
mjcheetham: We did not skip any. This test is checking the precedence of the 4 options:

```
                     1            2            3
credential.httpProxy > HTTP_PROXY > http.proxy > GCM_HTTP_PROXY
```

There are three inequalities to check for these four options.

docs/netconfig.md
jeschu1: I think it doesn't hurt to just have (deprecated) in the notes

Issue #57, opened 2019-06-28
Add self-configuration verbs


mjcheetham:
**Feature description**

Rather than having platform/distribution-mechanism specific post-install configuration logic, we should just move this to .NET code, and introduce some new CLI verbs that can be easily invoked from the installers etc.

- [ ] `configure`/`install`:
  - Configure ourselves as the credential helper, ideally in user/global config.
- [ ] `unconfigure`
  - Remove GCM from the configured credential helper. Should we reset to previous or platform default? Just remove our setting?
- [ ] `uninstall`
  - Invoke the uninstall script on Mac and invoke the Inno uninstaller on Windows?

Issue #46, opened 2019-06-05
Enhance Windows installer to configure itself as helper in all Git installs


mjcheetham:
The current Windows installer just drops the GCM Core binaries in %ProgramFiles%. We should also look at setting the `credential.helper` configuration entry, as well as the required `useHttpPath = true` entry for dev.azure.com remotes.

A few options include:
- Replicate GCM Windows and introduce an `install` command that discovers all Git installations (including msys installs?) and copies itself into the directory, and sets the `credential.helper = manager`
- Set the current user's `~/.gitconfig` to:
```ini
[credential]
    helper = # empty value first to reset the helper list from --system
    helper = <path/to/gcmcore>
```
- Put `%ProgramFiles%\Git Credential Manager` on the `%PATH%`? _(problem: how do we support side-by-side GCM Core and GCM Windows if they're both called `git-credential-manager.exe`?)_
- Something else?

These options require answering the question about should GCM Core be a machine-wide or user-scoped install?

PR #64, opened 2019-08-16
Quote bash variables in installation scripts

czipperz:
When running `uninstall.sh` I ran into the issue that bash variables weren't properly quoted so the script produced error messages.  This should fix these problems.

### DIFF
--- a/src/osx/Installer.Mac/build.sh
+++ b/src/osx/Installer.Mac/build.sh
@@ -8,16 +8,16 @@ echo "Building Installer.Mac..."
 
 # Directories
 THISDIR="$( cd "$(dirname "$0")" ; pwd -P )"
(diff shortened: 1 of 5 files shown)


mjcheetham:
Thanks for the contribution, @czipperz! All looks good to me :shipit: 

> @mjcheetham what kind of automated validation do we have for this script?

@derrickstolee, there is not automated validation for the _uninstall.sh_ script unfortunately. There is work planned to change the way installation (and hence changes to uninstall) on Windows and macOS works. I'll make a note that we might want to test the install/uninstall flows as part of that work. See #46, #57.

The other scripts that are for build are run in the automated CI and PR builds (and on local dev machines), so we have coverage of those.

Issue #42, opened 2019-05-13
Add option to use AAD ATs over Azure DevOps PATs


mjcheetham:
We may want to look at directly using the access tokens (ATs) from MSAL/ADAL rather than exchanging them for Azure DevOps personal access tokens (PATs).

There were concerns about the maximum supported size of a HTTP request header with cURL in GCM Windows which meant we opted for (compact) PATs. This may no longer be a limitation.

A few other possible (performance) concerns:

- Since we'd be asking MSAL/ADAL to get the AT each time, we're now shelling out to the helper proc for every Git auth request whereas currently we just P/Invoke to Keychain/WinCredAPI.
- PATs 'just work' and we don't need to know or resolve the tenant for the Azure DevOps organisation, which would now require another web call for every Git auth request. We'd rather not have to introduce an org-tenant cache (people can switch an org's tenant as a common 'growing up' scenario from MSA to AAD).
- If we move to always calling MSAL/ADAL _and_ we're sharing the Visual Studio token cache files we might be running into contention for that file/locking the cache since GCM does not live for a long time (we always must read/write to the file for each call), and many applications call/spawn Git in the background (see VSCode's 'auto-fetch every 10 seconds' feature).
- The Azure DevOps server might take longer to validate the AT is valid vs a PAT, or there may be increased load server-side (if an AAD<->Azure DevOps call is required to do this).

mjcheetham:
Note to the implementor:

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1259

When we come to moving to implementing using AAD Access Tokens only (rather than PATs) we should cache the MSAL instance metadata to allow for network-less calls to `AcquireTokenSilent`.

mjcheetham:
We should look to making this possible now that the changes in #49 are complete.
This behaviour (PAT vs AAD AT) should be controlled via an environment variable and/or Git configuration entry.

mjcheetham:
Abandoning and marking as inactive whilst we reevaluate priorities for this work.

PR #65, opened 2019-08-19
Add local cache of Azure Repos organisation authorities

mjcheetham:
To avoid making a web call each time we want to create a new credential (to get the authentication authority for the Azure DevOps organisation), we cache the value against the org name in an INI file that is stored in `~/.gcm/azure-repos/store.ini`.

This is an important web call to avoid in an upcoming PR that will introduce an AAD AT-only mode where we need the authority for each GCM 'get' request, and not just ones when the PAT was not found (see #42).

Should an authentication fail, we clear the organisation cache entry for that failing remote. This will allow us to auto-clear stale authority values when, for example, an Azure DevOps organisation moves from being MSA- to AAD-backed.

This INI-based store will also be utilised in future PRs for storing user state for the AAD AT-only mode (see #42). A simple file/directory-based format was considered, but there are problems with escaping special characters. A JSON-based format could also be used. The INI based format was chosen to be consistent with Git's own configuration/storage format that is INI-like.

### DIFF
--- a/src/shared/Microsoft.Git.CredentialManager/IniFileValueStore.cs
+++ b/src/shared/Microsoft.Git.CredentialManager/IniFileValueStore.cs
@@ -1,0 +1,145 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+using System;
+using System.IO;
+
+namespace Microsoft.Git.CredentialManager
+{
+    /// <summary>
+    /// Value store that uses an INI file as the persistent storage.
+    /// </summary>
+    public class IniFileValueStore : ITransactionalValueStore<string, string>
+    {
+        private readonly IFileSystem _fileSystem;
+        private readonly IniSerializer _serializer;
+        private readonly string _filePath;
+        private readonly string _parentPath;
+        private readonly object _fileLock = new object();
+        private IniFile _iniFile;
+
+        public IniFileValueStore(IFileSystem fileSystem, IniSerializer serializer, string filePath)
+        {
+            _fileSystem = fileSystem;
+            _serializer = serializer;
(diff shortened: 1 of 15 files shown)


src/shared/Microsoft.Git.CredentialManager/IniFileValueStore.cs
```diff
@@ -0,0 +1,145 @@
...
+                {
+                    _serializer.Serialize(_iniFile, writer);
+                    _fileSystem.WriteAllText(_filePath, writer.ToString());
+                }
+            }
+        }
+
+        public bool TryGetValue(string key, out string value)
+        {
+            lock (_fileLock)
```
jrbriggs: What are we locking here? The name indicates the file, but I believe the data has been loaded at this point.

src/shared/Microsoft.Git.CredentialManager/IniFileValueStore.cs
```diff
@@ -0,0 +1,145 @@
...
+                    throw new ArgumentException($"Invalid key '{key}'.", nameof(key));
+                }
+
+                return _iniFile.TryGetValue(section, scope, property, out value);
+            }
+        }
+
+        public void SetValue(string key, string value)
+        {
+            lock (_fileLock)
```
jrbriggs: Same question on getting crisp with the locking semantics.

jrbriggs:
To be clear, the "waiting on author" is for the 4 questions left in the code itself. The meta question on the choice of ini files is a philosophical gut check, not something to go retool.

mjcheetham:
Abandoning whilst we reevaluate priorities for this work. Branch is archived in the upstream repository in branch `vs/ado-cache`.

PR #69, opened 2019-09-02
New Install 1/6: Optimise the use of Git configuration from the Settings component

mjcheetham:
Keep hold of a single `IGitConfiguration` object when creating an instance of the `Settings` component.

When querying for scoped settings now we enumerate all configuration entries and store all those that have the same section and property name locally up-front. This prevents us from calling `TryGetValue` on the configuration object multiple times.

Add tracing to the libgit2 calls.

### DIFF
--- a/src/shared/Microsoft.Git.CredentialManager.Tests/Interop/LibGit2Tests.cs
+++ b/src/shared/Microsoft.Git.CredentialManager.Tests/Interop/LibGit2Tests.cs
@@ -7,6 +7,7 @@ using System.IO;
 using System.Text;
 using Microsoft.Git.CredentialManager.Interop;
 using Microsoft.Git.CredentialManager.Interop.Posix.Native;
+using Microsoft.Git.CredentialManager.Tests.Objects;
 using Xunit;
 
 namespace Microsoft.Git.CredentialManager.Tests.Interop
@@ -16,7 +17,8 @@ namespace Microsoft.Git.CredentialManager.Tests.Interop
         [Fact]
         public void LibGit2_GetRepositoryPath_NotInsideRepository_ReturnsNull()
         {
-            var git = new LibGit2();
+            var trace = new NullTrace();
+            var git = new LibGit2(trace);
             string randomPath = Path.Combine(Path.GetTempPath(), $"{Guid.NewGuid():N}");
             Directory.CreateDirectory(randomPath);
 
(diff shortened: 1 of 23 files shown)


src/shared/Microsoft.Git.CredentialManager.Tests/Interop/LibGit2Tests.cs
```diff
@@ -16,7 +17,8 @@ public class LibGit2Tests
         [Fact]
         public void LibGit2_GetRepositoryPath_NotInsideRepository_ReturnsNull()
         {
-            var git = new LibGit2();
+            var trace = new NullTrace();
+            var git = new LibGit2(trace);
```
jeschu1: tiny tiny nit: var git = new LibGit2(new NullTrace()); could have been another option

src/shared/Microsoft.Git.CredentialManager.Tests/Interop/LibGit2Tests.cs
```diff
@@ -28,20 +30,97 @@ public void LibGit2_GetRepositoryPath_NotInsideRepository_ReturnsNull()
...
             }
         }
 
+        [Fact]
+        public void LibGit2Configuration_Enumerate_CallbackReturnsTrue_InvokesCallbackForEachEntry()
+        {
+            string repoPath = CreateRepository();
+            Git(repoPath, "config --local foo.name lancelot").AssertSuccess();
+            Git(repoPath, "config --local foo.quest seek-holy-grail").AssertSuccess();
+            Git(repoPath, "config --local foo.favcolor blue").AssertSuccess();
```
jeschu1: Your test names make me laugh :-)

src/shared/Microsoft.Git.CredentialManager.Tests/Interop/LibGit2Tests.cs
```diff
@@ -28,20 +30,97 @@ public void LibGit2_GetRepositoryPath_NotInsideRepository_ReturnsNull()
...
+                    if (name.StartsWith("foo."))
+                    {
+                        actualVisitedEntries.Add((name, value));
+                    }
+
+                    // Continue enumeration
+                    return true;
+                }
+
+                config.Enumerate(cb);
```
jeschu1: i'm guessing you are using "cb" to validate that we only have the 3 entries that start with 'foo' and they match.  May be worth a comment or two.

src/shared/Microsoft.Git.CredentialManager/EnumerableExtensions.cs
```diff
@@ -25,5 +25,28 @@ public static IEnumerable<TSource> ConcatMany<TSource>(this IEnumerable<TSource>
...
             return result;
         }
+
+        /// <summary>
+        /// Creates a <see cref="HashSet{TSource}"/> from an <see cref="IEnumerable{TSource}"/>.
+        /// </summary>
+        /// <param name="sequence">Sequence to construct a <see cref="HashSet{TSource}"/> with.</param>
+        /// <typeparam name="TSource">Type of the sequence elements.</typeparam>
+        /// <returns><see cref="HashSet{TSource}"/> containing unique elements from the sequence.</returns>
+        public static HashSet<TSource> ToHashSet<TSource>(this IEnumerable<TSource> sequence)
```
jrbriggs: Are these used?

src/shared/Microsoft.Git.CredentialManager/Interop/LibGit2.cs
```diff
@@ -129,6 +173,7 @@ private void Dispose(bool disposing)
         {
             unsafe
```
jeschu1: should we check the disposing flag?

src/shared/Microsoft.Git.CredentialManager.Tests/Interop/LibGit2Tests.cs
mjcheetham: > i'm guessing you are using "cb" to validate that we only have the 3 entries that start with 'foo' and they match.

Actually, no. The configuration object will contain more than just the three entries we explicitly add to the local configuration level in the test.

We filter to 'foo' section entries here to ignore any machine or user level entries which are 'noise' on the developer and build machines.

The test still checks that we enumerate entries in the configuration, and that we enumerate _at least_ those entries we expect.

The test below tests that enumeration can be stopped by the callback.

src/shared/Microsoft.Git.CredentialManager/EnumerableExtensions.cs
mjcheetham: They used to be used in an earlier revision.. removing..

src/shared/Microsoft.Git.CredentialManager/Interop/LibGit2.cs
mjcheetham: You're correct. In fact I've noticed a few other `IDisposable` objects not implementing the disposable pattern 100% correctly.

I'm going to create an abstract `DisposableObject` class that will do this and ensure all object in GCM Core that are disposable derive from this.

PR #70, opened 2019-09-02
New Install 2/6: Rename main executable to GCM Core to make SxS easier

mjcheetham:
Rename the main executable to include "-core" to make side-by-side installs of GCM Core and Windows-only GCM easier.

We can just change the configuration between "manager" and "manager-core".

~**Note for reviewers:** please wait until #69 is complete for an easier time reviewing, or just view the latest commit in this PR as this branch is based on changes from #69.~

### DIFF
--- a/.github/ISSUE_TEMPLATE/auth-problem.md
+++ b/.github/ISSUE_TEMPLATE/auth-problem.md
@@ -8,7 +8,7 @@ assignees: ''
 
 **Which version of GCM Core are you using?**
 
-From a terminal, run `git-credential-manager version` and paste the output.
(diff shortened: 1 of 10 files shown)


mjcheetham:
> Approved assuming we don't also need to update
> 
> ```
> src\osx\Installer.Mac\scripts\configure-git.sh:11:git config --system credential.helper /usr/local/share/gcm-core/git-credential-manager
> ```
> 
> which I assume will be addressed by the rest of this series. If not, we should fix that here.

Ah! Good catch. Yes, this is addresses in #73 but I'll update it in this PR too so that each merge commit is 'good' 😄

PR #71, opened 2019-09-02
New Install 3/6: Introduce IEnvironment to replace IEnvironmentVariables

mjcheetham:
Replace the simple IEnvironmentVariables envar wrapper with a component that includes the envars and other methods for manipulating the $PATH.

This will be useful in future commits where we need to inspect and update the $PATH for postinstall and preuninstall.

Also expose the new IEnvironment and existing IGit components on the command context in prep.

**Note for reviewers:** please wait until #70 is complete for an easier time reviewing, or just view the latest commit in this PR as this branch is based on changes from #70.


src/shared/Microsoft.Git.CredentialManager/Interop/Posix/PosixEnvironment.cs
```diff
@@ -0,0 +1,52 @@
...
+                string[] paths = pathValue.Split(':');
+                return paths.Any(x => IsSamePath(x, directoryPath));
+            }
+
+            return false;
+        }
+
+        public virtual void AddDirectoryToPath(string directoryPath, bool system)
+        {
+            throw new NotImplementedException();
```
mjcheetham: We don't bother implementing the $PATH manipulation methods for POSIX (or specifically for macOS) because we don't need them (yet).

The postinstall script run by the package installer creates a GCM Core symlink in /usr/local/bin which is already on the $PATH.

src/shared/Microsoft.Git.CredentialManager.Tests/SettingsTests.cs
```diff
@@ -24,10 +23,10 @@ public void Settings_IsDebuggingEnabled_EnvarUnset_ReturnsFalse()
         [Fact]
         public void Settings_IsDebuggingEnabled_EnvarTruthy_ReturnsTrue()
         {
-            var envars = new EnvironmentVariables(new Dictionary<string, string>
+            var envars = new TestEnvironment
```
jrbriggs: In the interest of DRY, should we extract a helper function to set up the test environment?

src/shared/Microsoft.Git.CredentialManager/IEnvironment.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// <param name="system">True to update the path for all users on the system, false to update only the user's path.</param>
+        void RemoveDirectoryFromPath(string directoryPath, bool system);
+
+        /// <summary>
+        /// Check if two paths are the same for the current platform and file system. Symbolic links are not followed.
+        /// </summary>
+        /// <param name="a">File path.</param>
+        /// <param name="b">File path.</param>
+        /// <returns>True if both file paths are the same, false otherwise.</returns>
+        bool IsSamePath(string a, string b);
```
jrbriggs: This feels a bit out of place to hang off `IEnvironment`. Should this live on `IFileSystem`?

src/shared/Microsoft.Git.CredentialManager.Tests/SettingsTests.cs
mjcheetham: There is also the argument about being explicit in unit tests; being able to see what the preconditions and postconditions are for each test without jumping into several other methods.

Not sure what benefit this would do us here either since each test has a different setup with different variables.

src/shared/Microsoft.Git.CredentialManager/IEnvironment.cs
mjcheetham: Yes. I'll move this to be on `IFileSystem` and the implementation be split between each platform (`WindowsFileSystem`, `Posix/MacOSFileSystem` etc)

PR #72, opened 2019-09-02
New Install 4/6: Extend IGitConfiguration to allow manipulating multivars

mjcheetham:
Extend the IGitConfiguration to allow manipulation of Git configuration 'multivars' (config entries that can have multiple values). This will be used in future commits to set the credential.helper variable to have two values.

**Note for reviewers:** please wait until #71 is complete for an easier time reviewing, or just view the latest commit in this PR as this branch is based on changes from #71.

### DIFF
--- a/src/shared/Microsoft.Git.CredentialManager/IGit.cs
+++ b/src/shared/Microsoft.Git.CredentialManager/IGit.cs
@@ -14,12 +14,56 @@ namespace Microsoft.Git.CredentialManager
         void Enumerate(GitConfigurationEnumerationCallback cb);
 
         /// <summary>
+        /// Get a snapshot of the configuration filtered to the specified level.
+        /// </summary>
+        /// <param name="level">Configuration level filter.</param>
+        /// <returns>Git configuration snapshot.</returns>
+        IGitConfiguration GetFilteredConfiguration(GitConfigurationLevel level);
+
+        /// <summary>
         /// Try and get the value of a configuration entry as a string.
         /// </summary>
         /// <param name="name">Configuration entry name.</param>
         /// <param name="value">Configuration entry value.</param>
         /// <returns>True if the value was found, false otherwise.</returns>
         bool TryGetValue(string name, out string value);
+
+        /// <summary>
+        /// Set the value of a configuration entry.
+        /// </summary>
+        /// <param name="name">Configuration entry name.</param>
+        /// <param name="value">Configuration entry value.</param>
+        void SetValue(string name, string value);
+
+        /// <summary>
+        /// Deletes a configuration entry from the highest level.
+        /// </summary>
+        /// <param name="name">Configuration entry name.</param>
+        void DeleteEntry(string name);
+
+        /// <summary>
+        /// Get all values of a multivar configuration entry.
+        /// </summary>
+        /// <param name="name">Configuration entry name.</param>
+        /// <param name="regexp">Regular expression to filter which variables we're interested in. Use null to indicate all.</param>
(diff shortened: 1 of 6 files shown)


src/shared/Microsoft.Git.CredentialManager/Interop/LibGit2.cs
```diff
@@ -66,6 +67,8 @@ public string GetRepositoryPath(string path)
             _trace.WriteLine($"Discovering repository from path '{path}'...");
 
             var buf = new git_buf();
+
+            _trace.WriteLine($"Discovering repository from path '{path}'...");
```
jrbriggs: Why do we trace this again?

src/shared/Microsoft.Git.CredentialManager/Interop/LibGit2.cs
```diff
@@ -166,6 +213,83 @@ public unsafe bool TryGetValue(string name, out string value)
 
             return false;
         }
+        public unsafe void SetValue(string name, string value)
```
jrbriggs: nit: blank line between methods

src/shared/Microsoft.Git.CredentialManager/IGit.cs
```diff
@@ -13,13 +13,56 @@ public interface IGitConfiguration : IDisposable
...
+        /// <returns>All values of the multivar configuration entry.</returns>
+        IEnumerable<string> GetMultivarValue(string name, string regexp);
+
+        /// <summary>
+        /// Set a multivar configuration entry value.
+        /// </summary>
+        /// <param name="name">Configuration entry name.</param>
+        /// <param name="regexp">Regular expression to indicate which values to replace.</param>
+        /// <param name="value">Configuration entry value.</param>
+        void SetMultivarValue(string name, string regexp, string value);
```
jrbriggs: It's unclear to me from the interface docs if this appends or replaces values. Can we clarify?

src/shared/Microsoft.Git.CredentialManager/Interop/LibGit2.cs
mjcheetham: Bad merge!

src/shared/Microsoft.Git.CredentialManager/IGit.cs
mjcheetham: It can do either, depending on the value of the regular expression.

The [libgit2 docs](https://libgit2.org/libgit2/#HEAD/group/config/git_config_set_multivar) for the underlying method `git_config_set_multivar` is:

> Set a multivar in the local config file.
>
> git_config * cfg : where to look for the variable
> const char * name : the variable's name
> const char * regexp : a regular expression to indicate which values to replace
> const char * value : the new value.

What I've observe happens (in pseudo code) is:

```python
for entry in entries
    if entry.name matches regexp
        entry.value = value
        return

newEntry := { name, value }
entries.add(newEntry)
```

..in a sense "setting/replacing" the value of the 'matched entry' (and if that is an entry that doesn't exist, it is created and the value 'set/replaced').

src/shared/Microsoft.Git.CredentialManager/Interop/LibGit2.cs
mjcheetham: Damn you merge! 😢

src/shared/Microsoft.Git.CredentialManager/IGit.cs
mjcheetham: I've added a `<remarks/>` section indicating this behaviour.

PR #73, opened 2019-09-02
New Install 5/6: Add IConfigurationService and (un)configure commands

### LINKED ISSUE #57: Add self-configuration verbs

mjcheetham:
Add service that can configure and unconfigure components at the user or system level, and two commands that map to configure and unconfigure.

Call these new commands from the Windows and Mac (un)installers.

Update the help command output to include the new commands.

Partially addresses #57, #46.

~**Note for reviewers:** please wait until #72 is complete for an easier time reviewing, or just view the latest commit in this PR as this branch is based on changes from #72.~


src/osx/Installer.Mac/uninstall.sh
```diff
@@ -1,22 +1,23 @@
...
-if [ "$HELPER" = "/usr/local/share/gcm-core/git-credential-manager-core" ]
+# Ensure we're running as root
+if [ $(id -u) != "0" ]
 then
-	echo "Resetting credential helper to 'osxkeychain'..."
-	sudo git config --system credential.helper osxkeychain
-else
-	echo "GCM was not configured as the Git credential helper."
+	sudo "$0" "$@"
+	exit $?
```
jeschu1: we try to sudo and exit?

src/shared/Microsoft.Git.CredentialManager.Tests/Commands/UnconfigureCommandTests.cs
```diff
@@ -0,0 +1,67 @@
...
+using Xunit;
+
+namespace Microsoft.Git.CredentialManager.Tests.Commands
+{
+    public class UnconfigureCommandTests
+    {
+        [Theory]
+        [InlineData("unconfigure", true)]
+        [InlineData("UNCONFIGURE", true)]
+        [InlineData("uNcOnFiGuRe", true)]
```
jeschu1: What you've done for configure/unconfigure seems great, I don't know if we need to support all cases (upper/lower chars), but the work seems minimal and have great tests so what you have here is definitely fine by me.

src/shared/Microsoft.Git.CredentialManager.Tests/ConfigurationServiceTests.cs
```diff
@@ -0,0 +1,96 @@
...
+            service.AddComponent(component3.Object);
+
+            await service.ConfigureAsync(false);
+
+            component1.Verify(x => x.ConfigureAsync(false, context.Environment, It.IsAny<IGitConfiguration>()), Times.Once);
+            component2.Verify(x => x.ConfigureAsync(false, context.Environment, It.IsAny<IGitConfiguration>()), Times.Once);
+            component3.Verify(x => x.ConfigureAsync(false, context.Environment, It.IsAny<IGitConfiguration>()), Times.Once);
+        }
+
+        [Fact]
```
jeschu1: The change between this / the next test seems to be based on weather or not we run Async.  Can you confirm how that relates to the change in test name (System/User)

src/shared/Microsoft.Git.CredentialManager.Tests/StringExtensionsTests.cs
```diff
@@ -203,12 +203,12 @@ public void StringExtensions_TrimUntilIndexOf_String(string input, string trim,
...
-        [InlineData("fooTRIMbar", "tRiM", StringComparison.Ordinal, "fooTRIMbar")]
-        [InlineData("fooTRIMbar", "TRIM", StringComparison.OrdinalIgnoreCase, "bar")]
-        [InlineData("fooTRIMbar", "trim", StringComparison.OrdinalIgnoreCase, "bar")]
-        [InlineData("fooTRIMbar", "tRiM", StringComparison.OrdinalIgnoreCase, "bar")]
+        [InlineData("fooTRIMbarTRIMsoup", "TRIM", StringComparison.Ordinal, "barTRIMsoup")]
+        [InlineData("fooTRIMbarTRIMsoup", "trim", StringComparison.Ordinal, "fooTRIMbarTRIMsoup")]
+        [InlineData("fooTRIMbarTRIMsoup", "tRiM", StringComparison.Ordinal, "fooTRIMbarTRIMsoup")]
+        [InlineData("fooTRIMbarTRIMsoup", "TRIM", StringComparison.OrdinalIgnoreCase, "barTRIMsoup")]
+        [InlineData("fooTRIMbarTRIMsoup", "trim", StringComparison.OrdinalIgnoreCase, "barTRIMsoup")]
+        [InlineData("fooTRIMbarTRIMsoup", "tRiM", StringComparison.OrdinalIgnoreCase, "barTRIMsoup")]
```
jeschu1: Maybe try "fooTrim" and make sure it yields "" for "Trim".

Didnt' see on first glance why did you update these?

src/shared/Microsoft.Git.CredentialManager.Tests/StringExtensionsTests.cs
```diff
@@ -271,5 +271,28 @@ public void StringExtensions_TrimUntilLastIndexOf_String_Null_ThrowsArgumentNull
...
             Assert.Throws<ArgumentNullException>(() => StringExtensions.TrimUntilLastIndexOf(null, "://"));
         }
+
+        [Theory]
+        [InlineData("fooTRIMbar", "TRIM", "foobar")]
+        [InlineData("fooTRIMbar", "trim", "fooTRIMbar")]
+        [InlineData("fooTRIMbar", "tRiM", "fooTRIMbar")]
+        public void StringExtensions_TrimMiddle_String(string input, string trim, string expected)
+        {
+            string actual = StringExtensions.TrimMiddle(input, trim);
```
jeschu1: Does this just remove a particular string?  Does it have to be in the middle?  Wondering if RemoveString or similar would work better than TrimMiddle

src/shared/Microsoft.Git.CredentialManager/Interop/MacOS/MacOSEnvironment.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace Microsoft.Git.CredentialManager.Interop.MacOS
+{
+    public class MacOSEnvironment : PosixEnvironment
+    {
+        public override bool IsSamePath(string a, string b)
+        {
+            a = Path.GetFullPath(a);
+            b = Path.GetFullPath(b);
+
+            // TODO: check if APFS/HFS+ is in case-sensitive mode
```
jeschu1: First the first release can we leave the above as a TODO?

src/osx/Installer.Mac/uninstall.sh
mjcheetham: Yes.. if we are not running as root we invoke ourselves `"@0"` with all args `"$@"` with sudo to force running as root. We then exit the script returning the last error code `$?`.

src/shared/Microsoft.Git.CredentialManager.Tests/ConfigurationServiceTests.cs
mjcheetham: > or not we run Async.

This isn't correct. We're always doing `await service.UnconfigureAsync(bool)` but the boolean we pass is 'unconfigure the user' or 'unconfigure the system', hence the test name change. 😄

src/shared/Microsoft.Git.CredentialManager.Tests/StringExtensionsTests.cs
mjcheetham: These are changes from a previous commit/PR that has been rebased away now.
The reason for the change was the _new_ command `TrimMiddle` cuts out the first 'trim' and leaves the prefix and suffix. This existing 'trim' method will cut out the prefix too, and I added a test to ensure it removes only the first match.

src/shared/Microsoft.Git.CredentialManager/Interop/MacOS/MacOSEnvironment.cs
mjcheetham: Commit is from a previous PR (now merged). This is OK for the preview. Once a user hits a problem we can reevaluate. Most macOS users don't change this.

src/shared/Microsoft.Git.CredentialManager.Tests/StringExtensionsTests.cs
mjcheetham: TrimMiddle because it doesn't trim from the end of the string, like the rest of the trim commands. Instead it 'cuts out' the matched text (i.e, it keeps the text before and after the matched string).

src/shared/Microsoft.Git.CredentialManager.Tests/ConfigurationServiceTests.cs
jeschu1: I see, that is a bit confusing when just given a true/false, but makes sense

jrbriggs: +1 maybe consider an enum for the level to configure for clarity?

src/shared/Microsoft.Git.CredentialManager/ConfigurationService.cs
```diff
@@ -0,0 +1,107 @@
...
+        /// </summary>
+        string Name { get; }
+
+        /// <summary>
+        /// Configure the environment and Git to work with this hosting provider.
+        /// </summary>
+        /// <param name="system">Configure the component for all users on the system.</param>
+        /// <param name="environment">Environment variables.</param>
+        /// <param name="configuration">User-level Git configuration.</param>
+        Task ConfigureAsync(bool system, IEnvironment environment, IGitConfiguration configuration);
```
jrbriggs: How about an enum rather than the bool to make this a bit simpler to grok? We already have `GitConfigurationLevel`.

src/shared/Microsoft.Git.CredentialManager.Tests/ConfigurationServiceTests.cs
mjcheetham: Added and plumbed through an enum!

src/shared/Microsoft.Git.CredentialManager/ConfigurationService.cs
mjcheetham: Added and plumbed through a new `ConfigurationTarget` enumeration!

PR #74, opened 2019-09-02
New Install 6/6: Implement (un)configuration for GCM Core

### LINKED ISSUE #57: Add self-configuration verbs

mjcheetham:
Add configuration components, which will ensure the exectuable is on the PATH (on Windows) and configure itself and Git's credential helper in the user's config. Azure Repos hosts on dev.azure.com will also have `useHttpPath` set to true.

Fixes #57, #46.

~**Note for reviewers:** please wait until #73 is complete for an easier time reviewing, or just view the latest commit in this PR as this branch is based on changes from #73.~


src/shared/Git-Credential-Manager/Program.cs
```diff
@@ -29,5 +32,24 @@ public static void Main(string[] args)
                 Environment.Exit(exitCode);
             }
         }
+
+        private static string GetApplicationPath()
+        {
+            Assembly entryAssembly = Assembly.GetExecutingAssembly();
```
jeschu1: Would any of these work as shorter options for the path?

1. Application.startupPath
2. System.IO.Path.GetDirectoryName( System.Reflection.Assembly.GetExecutingAssembly().Location)

src/shared/Microsoft.AzureRepos.Tests/AzureReposHostProviderTests.cs
```diff
@@ -171,5 +171,88 @@ public async Task AzureReposProvider_GetCredentialAsync_ReturnsCredential()
...
+            string key = $"{Constants.GitConfiguration.Credential.SectionName}.https://dev.azure.com.{Constants.GitConfiguration.Credential.UseHttpPath}";
+            var provider = new AzureReposHostProvider(new TestCommandContext());
+
+            var environment = new TestEnvironment();
+            var config = new TestGitConfiguration(new Dictionary<string, IList<string>>
+            {
+                [key] = new List<string> {"true"}
+            });
+
+            await provider.ConfigureAsync(
```
jeschu1: Is it easy / worth validating global hasn't been changed?

src/shared/Microsoft.AzureRepos.Tests/AzureReposHostProviderTests.cs
```diff
@@ -171,5 +171,88 @@ public async Task AzureReposProvider_GetCredentialAsync_ReturnsCredential()
...
+            Assert.Single(config.Dictionary);
+            Assert.True(config.Dictionary.TryGetValue(key, out IList<string> actualValues));
+            Assert.Single(actualValues);
+            Assert.Equal("true", actualValues[0]);
+        }
+
+        [Fact]
+        public async Task AzureReposHostProvider_ConfigureAsync_UseHttpPathSetFalse_SetsUseHttpPathTrue()
+        {
+            string key = $"{Constants.GitConfiguration.Credential.SectionName}.https://dev.azure.com.{Constants.GitConfiguration.Credential.UseHttpPath}";
```
jeschu1: nit: this string is used in several places, maybe a public const somewhere instead

src/shared/Microsoft.AzureRepos/AzureReposHostProvider.cs
```diff
@@ -103,5 +104,49 @@ protected override void ReleaseManagedResources()
...
+            IEnvironment environment, EnvironmentVariableTarget environmentTarget,
+            IGitConfiguration configuration, GitConfigurationLevel configurationLevel)
+        {
+            string useHttpPathKey = $"{KnownGitCfg.Credential.SectionName}.https://dev.azure.com.{KnownGitCfg.Credential.UseHttpPath}";
+
+            using (IGitConfiguration targetConfig = configuration.GetFilteredConfiguration(configurationLevel))
+            {
+                if (targetConfig.TryGetValue(useHttpPathKey, out string currentValue) && currentValue.IsTruthy())
+                {
+                    Context.Trace.WriteLine("Azure Repos host provider configuration is already set correctly.");
```
jeschu1: Maybe: Azure Repos host privder ${useHttpPathKey} is already set to true

src/shared/Microsoft.Git.CredentialManager/Application.cs
```diff
@@ -105,5 +126,104 @@ protected bool WriteException(Exception ex)
...
+                }
+            }
+
+            string helperKey = $"{Constants.GitConfiguration.Credential.SectionName}.{Constants.GitConfiguration.Credential.Helper}";
+            string gitConfigAppName = GetGitConfigAppName();
+
+            using (IGitConfiguration targetConfig = configuration.GetFilteredConfiguration(configurationLevel))
+            {
+                string[] currentValues = targetConfig.GetMultivarValue(helperKey, Constants.RegexPatterns.Any).ToArray();
+                if (currentValues.Length != 2 || !string.IsNullOrWhiteSpace(currentValues[0]) ||
```
jeschu1: Why are looking for exactly 2 vars?  Why do we only check the first for NullOrWhiteSpace?

src/shared/Git-Credential-Manager/Program.cs
jrbriggs: Neither of those include the filename, which I believe is the intent of the method.

src/shared/Microsoft.Git.CredentialManager/Constants.cs
```diff
@@ -13,6 +13,39 @@ public static class Constants
         public const string ProviderIdAuto  = "auto";
         public const string AuthorityIdAuto = "auto";
 
+        public static class RegexPatterns
```
jrbriggs: I haven't searched extensively but I only found usage of `Any`, `Empty`, and `None`. Shall we remove the unused patterns?

src/shared/Git-Credential-Manager/Program.cs
mjcheetham: > Neither of those include the filename, which I believe is the intent of the method.

@jrbriggs is correct, but also..

> Application.startupPath

This is [a WinForms type](https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.application.startuppath?view=netframework-4.8); we're a console application.

src/shared/Microsoft.Git.CredentialManager/Application.cs
mjcheetham: We require the Git configuration to look like this:

```ini
[credential]
    # technically we don't care how many helper entries appear before
    helper = # empty string here clears any helpers included previously
    helper = manager-core # GCM Core!
    # there must be no more helper entries after
```

I guess really we should be looking for _at least_ 2 vars and then only check the last two for empty & GCM Core. I'll look at changing it.

src/shared/Microsoft.Git.CredentialManager/Application.cs
```diff
@@ -111,5 +126,94 @@ protected bool WriteException(Exception ex)
...
+            string gitConfigAppName = GetGitConfigAppName();
+
+            string[] currentValues = configuration.GetMultivarValue(helperKey, Constants.RegexPatterns.Any).ToArray();
+            if (currentValues.Length != 2 || !string.IsNullOrWhiteSpace(currentValues[0]) ||
+                currentValues.Last() != gitConfigAppName)
+            {
+                Context.Trace.WriteLine("Updating Git credential helper configuration...");
+
+                // Clear any existing entries in the global configuration.
+                configuration.DeleteMultivarEntry(helperKey, Constants.RegexPatterns.Any);
```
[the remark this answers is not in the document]
mjcheetham: Old comment.

src/shared/Microsoft.AzureRepos.Tests/AzureReposHostProviderTests.cs
mjcheetham: What is "global" ?

src/shared/Microsoft.Git.CredentialManager/Constants.cs
mjcheetham: Sure.

src/shared/Microsoft.Git.CredentialManager/Application.cs
```diff
@@ -105,5 +136,114 @@ protected bool WriteException(Exception ex)
...
+                 * [credential]
+                 *     ...                         # any number of helper entries
+                 *     helper =                    # an empty value to reset/clear any previous entries
+                 *     helper = {gitConfigAppName} # the expected executable value in the last position & directly following the empty value
+                 *
+                 */
+
+                string[] currentValues = targetConfig.GetMultivarValue(helperKey, Constants.RegexPatterns.Any).ToArray();
+                if (currentValues.Length < 2 ||
+                    !string.IsNullOrWhiteSpace(currentValues[currentValues.Length - 1]) ||    // second to last entry is empty
```
jeschu1: Question on the comments here

Wouldn't the second to last entry be "currentValues.Length -2 ", and the last entry be "currentValues.Length-1"

mjcheetham: Ah yes! Correct! Will fix and ensure a unit test catches this.

mjcheetham: Fixed.

PR #79, opened 2019-09-19
Fix real-sign Windows build to actually sign

mjcheetham:
Ensure the Signing MicroBuild plugin has been installed when doing a real-signed Windows build.

PR #80, opened 2019-09-20
[Port] Fix real-sign Windows build to actually sign

mjcheetham:
Ensure the Signing MicroBuild plugin has been installed when doing a real-signed Windows build.

Port of fix already merged directly to the `release` branch (#79).


shango420:
const createOrder = (data, actions) => {
          return fetch('/api/paypal/order', {
              method: 'POST'
          }).then(res => {
              return res.json();
          }).then(json => {
              return json.orderID;
          });
      };

PR #85, opened 2019-12-04
Notarize and staple the installer pkg on macOS

mjcheetham:
Enable installation and executable on macOS Catalina (10.15) now that
Gate Keeper requires notarization. We staple the notarization ticket to
the package too to allow for offline installs.

### DIFF
--- a/.azure-pipelines/templates/osx/pack.signed/step5-dist.yml
+++ b/.azure-pipelines/templates/osx/pack.signed/step5-dist.yml
@@ -1,0 +1,36 @@
+steps:
+  - task: DownloadPipelineArtifact@1
+    displayName: Download signed installer
+    inputs:
+      buildType: 'current'
+      artifactName: 'tmp.macinstaller_signed'
+      downloadPath: '$(Build.StagingDirectory)\pkg'
+
+  - task: DownloadPipelineArtifact@1
+    displayName: Download signed payload
+    inputs:
+      buildType: 'current'
+      artifactName: 'tmp.macpayload_signed'
+      downloadPath: '$(Build.StagingDirectory)\payload'
+
+  - task: DownloadPipelineArtifact@1
+    displayName: Download symbols
+    inputs:
+      buildType: 'current'
+      artifactName: 'tmp.macsymbols'
+      downloadPath: '$(Build.StagingDirectory)\symbols'
+
+  - script: src/osx/SignFiles.Mac/notarize-pkg.sh -id "$(AppleId)" -p "$(AppleIdPassword)" -pkg '$(Build.StagingDirectory)\pkg\*.pkg'
+    displayName: Notarize and staple installer package
+
+  - script: |
+      cp "$(Build.StagingDirectory)/pkg/*.pkg" "$(Build.StagingDirectory)/publish/"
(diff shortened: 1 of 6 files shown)


.azure-pipelines/templates/osx/pack.signed/step5-dist.yml
```diff
@@ -0,0 +1,36 @@
...
+      downloadPath: '$(Build.StagingDirectory)\payload'
+
+  - task: DownloadPipelineArtifact@1
+    displayName: Download symbols
+    inputs:
+      buildType: 'current'
+      artifactName: 'tmp.macsymbols'
+      downloadPath: '$(Build.StagingDirectory)\symbols'
+
+  - script: src/osx/SignFiles.Mac/notarize-pkg.sh -id "$(AppleId)" -p "$(AppleIdPassword)" -pkg '$(Build.StagingDirectory)\pkg\*.pkg'
```
jrbriggs: Won't we need to also notarize the platform-native auth helper?

mjcheetham: Notarization happens recursively inside of PKG package files and DMG disk images.
You can notarize the following things:
1. Bundles (.app, .kext, etc)
2. Installer package files
3. Disk images
4. Mach-o binaries (like the MSAuth helper UNIX executable)

You can staple the following things:
1. Bundles
2. Installer package files
3. Disk images

Flat binaries cannot be stapled, but are still notarized either by manual submission, or submission as part of a 'container'.

wilbaker: Will these secrets (e.g. `"$(AppleIdPassword)"`) be echoed to the build output?

I was looking at the documentation for handling secrets here:

https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch

And it recommended using environment variables, but the details as to why are a bit light.

(Full disclosure: I am still very new to yaml definitions and this might be fine as it is now)

jrbriggs: Ah, thanks for clarifying!

mjcheetham: I think the problem the docs talk about is that although Azure Pipelines output will try to make sure that it will mask any secret variables or environment variables, any tools or script hosts _may_ echo out their command lines which might not be masked.

I can make these environment variables if needed, but I should also point out this is how VFS for Git is currently [running its notarization script](https://dev.azure.com/mseng/AzureDevOps/_apps/hub/ms.vss-ciworkflow.build-ci-hub?_a=edit-build-definition&id=6629).

Using the 'bash' task:
image

In the output (even with debug mode enabled) all you see is:

```
##[debug]arguments=--appleid kewillf@microsoft.com --password *** --kext /Users/vsts/agent/2.158.0/work/1/a/vfs4g_mac_kext_0.6.19276.1_signed/PrjFSKext.kext
...
Formatted command: . '/Users/vsts/agent/2.158.0/work/1/s/Scripts/Mac/notarize.sh' --appleid kewillf@microsoft.com --password *** --kext /Users/vsts/agent/2.158.0/work/1/a/vfs4g_mac_kext_0.6.19276.1_signed/PrjFSKext.kext
```

PR #91, opened 2020-01-31
Add GCM_INTERACTIVE/credential.interactive setting

### LINKED ISSUE #90: Setting to disable interactive/modal auth
GCM for Windows has an important setting for using Git on headless environments:

[GCM_INTERACTIVE](https://github.com/microsoft/Git-Credential-Manager-for-Windows/blob/master/Docs/Environment.md#gcm_interactive) and [credential.interactive](https://github.com/microsoft/Git-Credential-Manager-for-Windows/blob/master/Docs/Configuration.md#interactive)

These are important to bring over so we don't break headless scenarios, as well as being able to forcibly disable interactive auth in certain scenarios.

mjcheetham:
Add a setting to disable all user interaction. By setting `GCM_INTERACTIVE` or `credential.interactive` to a 'falsey' value, GCM Core will now fail and return an error if user interaction is required.

This is useful in headless and unattended environments, such as build servers, where is is preferable to fail than it is to hang waiting for input from a non-existent user.

The setting also existed in the previous GCM for Windows, but its behaviour has been slightly modified to treat 'always' values as 'auto'. See more in the documentation and the code.

The default value is `true` / permit interaction.

Closed #90.

### DIFF
--- a/src/shared/Microsoft.Git.CredentialManager/Authentication/AuthenticationBase.cs
+++ b/src/shared/Microsoft.Git.CredentialManager/Authentication/AuthenticationBase.cs
@@ -62,13 +62,28 @@ namespace Microsoft.Git.CredentialManager.Authentication
             return resultDict;
         }
 
-        protected void EnsureTerminalPromptsEnabled()
+        protected void ThrowIfUserInteractionDisabled()
+        {
+            if (!Context.Settings.IsInteractionAllowed)
+            {
+                string envName = Constants.EnvironmentVariables.GcmInteractive;
+                string cfgName = string.Format("{0}.{1}",
+                    Constants.GitConfiguration.Credential.SectionName,
+                    Constants.GitConfiguration.Credential.Interactive);
+
+                Context.Trace.WriteLine($"{envName} / {cfgName} is false/never; user interactivity has been disabled.");
+
+                throw new InvalidOperationException("Cannot prompt because user interactivity has been disabled.");
+            }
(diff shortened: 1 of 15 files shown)


src/shared/Microsoft.Git.CredentialManager/Authentication/AuthenticationBase.cs
```diff
@@ -62,13 +62,28 @@ protected AuthenticationBase(ICommandContext context)
             return resultDict;
         }
 
+        protected void EnsureUserInteractionAllowed()
```
derrickstolee: This was not the pattern I expected, but now that I think about it more it makes sense. You're only going to show a user-prompt if you _need_ it to continue, but this setting prevents that from happening. The only rational thing to do is to exit with failure.

mjcheetham: Yes. GCM Core follows a `die` model when a bug or impossible situation (need to speak to the user, but cannot) is reached. Tracing contains the most detailed information.

I was thinking about renaming these methods to `ThrowIfUserInteractionDisabled()` to make their behaviour clearer? Thoughts?

derrickstolee: `ThrowIf` seems to be a clearer synonym of `Ensure`, so I'm happy with that rename.

PR #92, opened 2020-02-05
Add ISystemPrompts component and impl for Windows

mjcheetham:
Introduce the `ISystemPrompts` component which provides native/system UI prompts, starting with basic credential prompts. Include a basic implementation for Windows.

We can introduce a Mac implementation later either using a built-in credential prompt, or bundling a our own simple Cocoa app.

This feature will allow basic authentication when the TTY is not available (such as when run from within a GUI app like Visual Studio).

Partly addresses #93.


src/shared/Microsoft.Git.CredentialManager/Interop/Windows/WindowsSystemPrompts.cs
```diff
@@ -0,0 +1,150 @@
...
+        {
+            EnsureArgument.NotNullOrWhiteSpace(resource, nameof(resource));
+
+            string message = $"Enter credentials for '{resource}'";
+
+            var credUiInfo = new CredUi.CredentialUiInfo
+            {
+                BannerArt = IntPtr.Zero,
+                CaptionText = "Git Credential Manager", // TODO: make this a parameter?
+                Parent = IntPtr.Zero, // TODO: get the parent window handle
```
jrbriggs: We'll want to parent this correctly so that Windows can always position the dialog above the parent HWND.

PR #95, opened 2020-02-12
Correctly check for GUI on Win/Mac/POSIX

mjcheetham:
Replace the `PlatformUtils.IsDesktopSession` util method with platform specific components that check use native APIs to determine the state of the current session.

`Environment.UserInteractive` is hard-coded to return true for POSIX and Windows platforms on .NET Core 2.x and 3.x.
dotnet/runtime#770

The .NET 5 implementation (not yet released) fixes this for the Windows platform only. We take a copy of that implementation for Windows.

On macOS we use the `SessionGetInfo` from the Security.framework.

On POSIX in general we also check for the X11 $DISPLAY environment variable.

### DIFF
--- a/src/shared/GitHub/GitHubAuthentication.cs
+++ b/src/shared/GitHub/GitHubAuthentication.cs
@@ -63,7 +63,9 @@ namespace GitHub
 
             // If the GitHub auth stack doesn't support flows such as RFC 8628 and we do not have
             // an interactive desktop session, we cannot offer OAuth authentication.
-            if ((modes & AuthenticationModes.OAuth) != 0 && !Context.IsDesktopSession && !GitHubConstants.IsOAuthDeviceAuthSupported)
+            if ((modes & AuthenticationModes.OAuth) != 0
+                && !Context.SessionManager.IsDesktopSession
+                && !GitHubConstants.IsOAuthDeviceAuthSupported)
             {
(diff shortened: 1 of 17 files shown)


src/shared/GitHub/GitHubAuthentication.cs
```diff
@@ -65,7 +65,9 @@ public async Task<AuthenticationPromptResult> GetAuthenticationAsync(Uri targetU
 
             // If the GitHub auth stack doesn't support flows such as RFC 8628 and we do not have
             // an interactive desktop session, we cannot offer OAuth authentication.
-            if ((modes & AuthenticationModes.OAuth) != 0 && !Context.IsDesktopSession && !GitHubConstants.IsOAuthDeviceAuthSupported)
+            if ((modes & AuthenticationModes.OAuth) != 0
+                && !Context.SessionManager.IsDesktopSession
+                && !GitHubConstants.IsOAuthDeviceAuthSupported)
```
dscho, line 70: This is purely a formatting change ;-)

mjcheetham: Not quite!

```diff
- !Context.IsDesktopSession
+ !Context.SessionManager.IsDesktopSession

dscho: Ah! I had missed that... Thank you!

PR #99, opened 2020-04-06
Add an OAuth2 client implementation

mjcheetham:
Add an OAuth2 client implementation that supports the authorization code grant, and the device authorization grant (device code) flows.

This client will be used by both the BitBucket and GitHub host providers in subsequent PRs.

Here are the RFCs for OAuth2 ([6749](https://tools.ietf.org/html/rfc6749)) and for the Device Auth Flow ([8628](https://tools.ietf.org/html/rfc8628)) that I followed when implementing.


src/shared/Microsoft.Git.CredentialManager/Authentication/OAuth/OAuth2Client.cs
```diff
@@ -0,0 +1,440 @@
...
+
+    public class OAuth2Client : IOAuth2Client
+    {
+        private readonly HttpClient _httpClient;
+        private readonly OAuth2ServerEndpoints _endpoints;
+        private readonly Uri _redirectUri;
+        private readonly string _clientId;
+        private readonly string _clientSecret;
+
+        public OAuth2Client(HttpClient httpClient, OAuth2ServerEndpoints endpoints, string clientId, Uri redirectUri = null, string clientSecret = null)
```
jrbriggs, line 66: Should we document which parameters with defaults are required for a particular flow? For example, I can construct an `OAuth2Client` with no `redirectUri`, but then I cannot call `GetAuthorizationCodeAsync` on that client.

src/shared/Microsoft.Git.CredentialManager/Authentication/OAuth/OAuth2Client.cs
```diff
@@ -0,0 +1,440 @@
...
+            }
+
+            var requestUriBuilder = new UriBuilder(_endpoints.AuthorizationEndpoint)
+            {
+                Query = queryParams.ToQueryString()
+            };
+
+            Uri requestUri = requestUriBuilder.Uri;
+
+            // TODO: handle error response
```
jrbriggs, line 107: Error handling feels like something we should address before merging, even if it's just tracing the response code.

src/shared/Microsoft.Git.CredentialManager/Authentication/OAuth/OAuth2Client.cs
```diff
@@ -0,0 +1,440 @@
...
+                    }
+                }
+                catch (TimeoutException)
+                {
+                    // Back-off exponentially (2 * x = x + x)
+                    retryInterval += retryInterval;
+                }
+
+                // Wait the polling interval before retrying
+                await Task.Delay(retryInterval);
```
jrbriggs, line 270: This has no stop case if the user cannot finish the device flow. Should it?

src/shared/Microsoft.Git.CredentialManager/Authentication/OAuth/OAuth2Client.cs
```diff
@@ -0,0 +1,440 @@
...
+
+            return request;
+        }
+
+        private static string GenerateRandomString()
+        {
+            return Guid.NewGuid().ToString("N");
+        }
+
+        private Exception CreateExceptionFromResponse(string json)
```
jrbriggs, line 324: Naming nit: this doesn't just create the exception, it also throws it. Let's rename or have it return the exception for the caller to throw. Latter might be preferable to avoid the extra stack frame?

src/shared/Microsoft.Git.CredentialManager/Authentication/OAuth/SystemWebBrowser.cs
```diff
@@ -0,0 +1,76 @@
...
+        private readonly string _responseHtml;
+
+        public SystemWebBrowser(string responseHtml = null)
+        {
+            _responseHtml = responseHtml ?? DefaultResponseHtml;
+        }
+
+        public async Task<Uri> OpenAndWaitForRedirect(Uri startUri, Uri redirectUri)
+        {
+            // TODO: ensure redirect is localhost
```
jrbriggs, line 24: Shall we clean up these TODOs as well?

src/shared/Microsoft.Git.CredentialManager/UriExtensions.cs
```diff
@@ -2,12 +2,29 @@
...
+            return ParseQueryString(uri.Query);
+        }
+        
+        private static IDictionary<string, string> ParseQueryString(string query)
+        {
+            return query
+                .TrimStart('?')
+                .Split('&')
+                .ToDictionary(
+                    x => x.Split('=')[0],
```
jrbriggs, line 24: Shouldn't this be UrlDecoded as well?

src/shared/Microsoft.Git.CredentialManager/UriExtensions.cs
```diff
@@ -2,12 +2,29 @@
...
+                    x => HttpUtility.HtmlDecode(x.Split('=')[1]));
```
jrbriggs, line 25: Why HtmlDecode and not UrlDecode if we're working with a query string?

src/shared/Microsoft.Git.CredentialManager/Authentication/OAuth/OAuth2Client.cs
mjcheetham: Ah! My bad.. this _should_ be `return`ing the `Exception` here..

src/shared/Microsoft.Git.CredentialManager/UriExtensions.cs
mjcheetham: Whoops!

src/shared/Microsoft.Git.CredentialManager/Authentication/OAuth/OAuth2Client.cs
mjcheetham: Added cancellation token support.

src/shared/Microsoft.Git.CredentialManager/Authentication/OAuth/SystemWebBrowser.cs
mjcheetham: Done!

mjcheetham:
> This is... a lot of code. And because I'm unfamiliar with the code I will admit that I am not able to give a confident approval.

Yes, sorry about that! I guess I could have split the commit into implementing the various parts of the OAuth2 spec separately, but not sure that'd help with the review of the protocol (unless you want to sit down and read the RFC 😃)

src/shared/Microsoft.Git.CredentialManager/Authentication/OAuth/OAuth2Client.cs
mjcheetham: This is an error in my logic! Redirect URI does not have to be specified for this flow to work.. the server should use the default configured redirect URI.

PR #101, opened 2020-04-17
Support OAuth based authentication on GitHub; include a UI refresh

mjcheetham:
Add an OAuth2 web browser-based authentication flow to replace the username & password flow for GitHub.com and augment the u/p flow for GHE instances. Also use a refreshed WPF-based UI on Windows for GitHub authentication. The UI supports selecting OAuth/browser authentication and also displays the URL of a GitHub Enterprise instance if authentication to one.

Currently both the 'basic' and OAuth options are presented for GitHub.com because the OAuth application has not yet been whitelisted for all organisations. In addition, GHE does not currently offer OAuth as an option as the application has not been deployed to GHE instances (we don't know what version to key off).

To override the supported authentication modes, set the `GCM_GITHUB_AUTHMODES` environment variable to a combination of `oauth` and `basic` (separate multiple values by a single space).

**Note:** The Device Authorization Grant flow (RFC 8628) is not currently supported by GitHub.com or GHE so this has been hidden behind a single compile-time constant. If this flow is ever supported, one only need to update the device authorization URL and remove that constant.

**Reviewer tip:** Reviewing this PR commit-by commit may be easier.

Prompt|Current UI|New UI
-|-|-
Credentials<br/>(github.com)|image|image
Credentials<br/>(Enterprise)|(same as above)|image
Two-factor|image|image

The new UI can also detect and only present authentication modes that are applicable:
Basic only|OAuth only
-|-
image|image

On non-Windows platforms (or when there is no interactive desktop session) a TTY-based menu is shown when there are multiple authentication options available:

```
Select an authentication method for 'https://my-ghe-instance.com/':
  1. Web browser (default)
  2. Username/password
option (enter for default): _
```

If there is only one option available the menu is not presented, and one of the following messages is presented:

```
Enter GitHub credentials for 'https://my-ghe-instance.com/'...
Username: mjcheetham
Password: 
```

```
```


src/shared/GitHub/GitHubResources.resx
```diff
@@ -0,0 +1,141 @@
...
+    <resheader name="version">
+        <value>1.3</value>
+    </resheader>
+    <resheader name="reader">
+        <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=2.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+    </resheader>
+    <resheader name="writer">
+        <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=2.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+    </resheader>
+    <data name="AuthenticationResponseSuccessHtml" xml:space="preserve"><![CDATA[<!DOCTYPE html><html><head>
```
mjcheetham, line 21: This HTML is presented after a successful authentication. It looks like this:

**This is an old screenshot.. I fixed the typo!**

image

src/shared/GitHub/GitHubResources.resx
```diff
@@ -0,0 +1,141 @@
...
+        <span class="plus">+</span>
+        <span class="gcm"/>
+    </div>
+    <div class="box">
+        <h1>Authentication Succeeded</h1>
+        <p>You may now close this tab and return to the application.</p>
+    </div>
+</body></html>]]></data>
+
+    <data name="AuthenticationResponseFailureHtmlFormat" xml:space="preserve"><![CDATA[<!DOCTYPE html><html><head>
```
mjcheetham, line 76: This HTML is presented after a failed authentication. It looks like this:

image

derrickstolee:
I'm going to document my issues when trying to get an installer for testing:

1. I couldn't figure out how to build an installer from Visual Studio. I eventually landed on the following command-line mechanism:

```sh
$ dotnet build Git-Credential-Manager-Core.sln -c WindowsDebug
```

2. It seems that the GitHub build is leaving some cruft in the `out` directory:
image

3. After fixing my PATH by restarting Git Bash, I deleted my github credentials and fetched a private repo. When going to the web view, I got this authorization page:
image

docs/configuration.md
```diff
@@ -136,8 +136,33 @@ Configure GCM Core to use the a proxy for network operations.
...
 ```
 
 **Also see: [GCM_HTTP_PROXY](environment.md#GCM_HTTP_PROXY-deprecated)**
+
+---
+
+### credential.gitHubAuthModes
+
+Override the available authentication modes presented during GitHub authentication.
+If this option is not set the available authentication modes will be automatically detected.
```
derrickstolee, line 152: "If this option is not set, then..."

src/shared/GitHub.Tests/GitHubHostProviderTests.cs
```diff
@@ -129,6 +141,210 @@ public void GitHubHostProvider_GetCredentialKey_GistHost_ReturnsCorrectKey()
...
+            ghApiMock.Setup(x => x.GetMetaInfoAsync(targetUri)).ReturnsAsync(metaInfo);
+
+            var provider = new GitHubHostProvider(context, ghApiMock.Object, ghAuthMock.Object);
+
+            AuthenticationModes actualModes = await provider.GetSupportedAuthenticationModesAsync(targetUri);
+
+            Assert.Equal(expectedModes, actualModes);
+        }
+
+
```
derrickstolee, line 272: nit: extra empty line

src/shared/GitHub/GitHubAuthentication.cs
```diff
@@ -27,50 +59,107 @@ public class GitHubAuthentication : AuthenticationBase, IGitHubAuthentication
...
+                            new TerminalMenuItem(2, "Username/password")
+                        };
 
-                userName = Context.Terminal.Prompt("Username");
-                password = Context.Terminal.PromptSecret("Password");
-            }
+                        int option = menu.Show();
+
+                        if (option == 1) goto case AuthenticationModes.OAuth;
+                        if (option == 2) goto case AuthenticationModes.Basic;
```
derrickstolee, line 133: This deviation from normal style is probably worth it here.

src/windows/GitHub.Authentication.Helper.Windows/Properties/Resources.Designer.cs
```diff
@@ -9,14 +9,17 @@
 //------------------------------------------------------------------------------
 
 namespace GitHub.Authentication.Helper.Properties {
+    using System;
+    
+
```
derrickstolee: nit: extra empty line

Git-Credential-Manager.sln
```diff
@@ -199,13 +199,9 @@ Global
 		{8DBBAB0A-970D-4BE3-958C-8CDC92F76549}.WindowsRelease|Any CPU.ActiveCfg = Release|Any CPU
 		{8DBBAB0A-970D-4BE3-958C-8CDC92F76549}.WindowsRelease|Any CPU.Build.0 = Release|Any CPU
 		{4C2DBC8A-B3F2-4C64-870A-BA79DA4BD403}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
-		{4C2DBC8A-B3F2-4C64-870A-BA79DA4BD403}.Debug|Any CPU.Build.0 = Debug|Any CPU
```
derrickstolee: Perhaps this deletion is why I didn't get the Windows installer on the default build. The `Release|Any CPU` line is similar.

src/shared/GitHub/GitHubAuthentication.cs
mjcheetham: Yeah. I'm usually quite strict about 'one liner' control flow blocks, but this form seemed to improve clarity IMO.

src/shared/Microsoft.Git.CredentialManager/TerminalMenu.cs
```diff
@@ -0,0 +1,120 @@
...
+            {
+                throw new ArgumentException("A default menu item has already been added.");
+            }
+
+            _items.Add(item);
+        }
+
+        public int Show()
+        {
+            TerminalMenuItem[] itemsArr = _items.ToArray();
```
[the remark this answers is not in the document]
mjcheetham: This is an artefact of past implementation designs; previously the `Id` of an item was just the index of the item in the array. Therefore I needed the _index_ of the default menu item. Now the `Id` is explicitly set in the `TerminalMenuItem` instead. Will change to avoid `ToArray`.

src/shared/Microsoft.Git.CredentialManager/TerminalMenu.cs
```diff
@@ -0,0 +1,120 @@
...
+                    if (hasDefault)
+                    {
+                        return itemsArr[defaultIndex].Id;
+                    }
+
+                    _terminal.WriteLine("No default option is configured.\n");
+                    continue;
+                }
+
+                if (!int.TryParse(optionStr, out int option))
```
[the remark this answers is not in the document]
mjcheetham: The API currently only allows `TerminalMenuItem`s be have integer IDs. If we wanted to be smarter in the future, we could just change it to be `string` and have smarter matching. I'll update the error message in this case to be `$"Invalid option '{optionStr}'. Expected a number.\n"`.

src/windows/GitHub.Authentication.Helper.Windows/Properties/Resources.Designer.cs
mjcheetham: This is a generated file.

Git-Credential-Manager.sln
mjcheetham: See my answer to your build question.

mjcheetham:
> I'm going to document my issues when trying to get an installer for testing:
> 
> 1. I couldn't figure out how to build an installer from Visual Studio. I eventually landed on the following command-line mechanism:
> 
> ```shell
> $ dotnet build Git-Credential-Manager-Core.sln -c WindowsDebug
> ```

The solution contains 6 configurations; a set of 3 pairs:
1. Debug
1. Release
1. WindowsDebug
1. WindowsRelease
1. MacDebug
1. MacRelease

The Debug and Release configurations build only the `src/shared` components (i.e, no installers, no UI helpers).
To build the platform specific bits you need to select the correct configuration in Visual Studio. For example `WindowsDebug`:

image

I'd rather we didn't have to have all these options, but Visual Studio (Mac and Windows) insists on trying to load and do design-time builds on projects it doesn't understand (like `WindowsDesktop` SDK projects on Mac), and fails with many errors. Cross-compilation is not supported for these projects so we need to exclude them from the build.

---

> 2. It seems that the GitHub build is leaving some cruft in the `out` directory:
>    image

These `wpftmp` directories are created by the WPF build process. They sit sibling to the main project output and not inside.. why? This is just what the standard .NET Core MSBuild targets/props do for WPF projects.

---

> 3. After fixing my PATH by restarting Git Bash, I deleted my github credentials and fetched a private repo. When going to the web view, I got this authorization page:
>    image
> 
> How can we make this authorization not be @mjcheetham?

This is currently presented because only select 'GitHub-owned' OAuth applications have been whitelisted to not need such a prompt. This is the reason the OAuth bits are currently disabled in the code as of this PR. We need to reach out to GitHub auth teams to whitelist us. They are already in discussion with Visual Studio teams about a similar problem.

derrickstolee:
> > How can we make this authorization not be @mjcheetham?
> 
> This is currently presented because only select 'GitHub-owned' OAuth applications have been whitelisted to not need such a prompt. This is the reason the OAuth bits are currently disabled in the code as of this PR. We need to reach out to GitHub auth teams to whitelist us. They are already in discussion with Visual Studio teams about a similar problem.

I'm fine with the authorization prompt, but I'd rather it be @microsoft (or equivalent) instead of @mjcheetham.

mjcheetham:
> > > How can we make this authorization not be @mjcheetham?
> 
> > 
> 
> > This is currently presented because only select 'GitHub-owned' OAuth applications have been whitelisted to not need such a prompt. This is the reason the OAuth bits are currently disabled in the code as of this PR. We need to reach out to GitHub auth teams to whitelist us. They are already in discussion with Visual Studio teams about a similar problem.
> 
> 
> 
> I'm fine with the authorization prompt, but I'd rather it be @microsoft (or equivalent) instead of @mjcheetham.

Sure! There is a big “Transfer” button I can press to move the app registration to another person or Organisation. However I’m not sure who that would then be “owned” by.. who is behind the curtain/mask? (Who is the admin of the @microsoft org)

derrickstolee:
> > > > How can we make this authorization not be @mjcheetham?
> > 
> > 
> > > 
> > 
> > 
> > > This is currently presented because only select 'GitHub-owned' OAuth applications have been whitelisted to not need such a prompt. This is the reason the OAuth bits are currently disabled in the code as of this PR. We need to reach out to GitHub auth teams to whitelist us. They are already in discussion with Visual Studio teams about a similar problem.
> > 
> > 
> > I'm fine with the authorization prompt, but I'd rather it be @microsoft (or equivalent) instead of @mjcheetham.
> 
> Sure! There is a big “Transfer” button I can press to move the app registration to another person or Organisation. However I’m not sure who that would then be “owned” by.. who is behind the curtain/mask? (Who is the admin of the @microsoft org)

I think having something rubber-stamped by the bureaucracy of @microsoft is better than a "random" user account ;) As long as there is a process to do this, I'm happy to wait until the right time to go through that process.

mjcheetham:
> I think having something rubber-stamped by the bureaucracy of @microsoft is better than a "random" user account ;) As long as there is a process to do this, I'm happy to wait until the right time to go through that process.

Oh I agree! My concern was more on not knowing the process yet and then end up losing it down the rabbit hole.

src/shared/GitHub/GitHubOAuth2Client.cs
```diff
@@ -0,0 +1,33 @@
+using System;
+using System.Net.Http;
+using Microsoft.Git.CredentialManager.Authentication.OAuth;
+
+namespace GitHub
+{
+    public class GitHubOAuth2Client : OAuth2Client
+    {
+        private static readonly string ClientId = "0120e057bd645470c1ed";
+        private static readonly string ClientSecret = "18867509d956965542b521a529a79bb883344c90";
```
jeffhostetler, line 10: secrets ??    not sure what the scope of these secrets are, but should we have a "TODO" to revisit and/or get them from a key vault of some kind?  again, not sure how these are used, so just asking.

jeffhostetler:
WRT the 6 projects in the solution, IIRC it was that kind of weird behaviors that led me to have 2 single platform .sln files with per-platform project files.  It seemed like some VS or VSmac tool was always trying to add things to generic "Debug" or "Release" and muck things up.

src/shared/GitHub/GitHubOAuth2Client.cs
mjcheetham: These are only really "secrets" if this OAuth application was used from a web server (confidential client) where it would have privileged access. For native apps (public client) this cannot be 'secret' since it will eventually need to exist in the application binary.

If you take a look inside something like the GitHub VS Extension, GitHub Desktop, or CLI, you'll find the ID and Secret baked into the assembly.

jeffhostetler:
Dumb question: in one of the dialogs there is a "redirecting to localhost:51990".  what is that and/or who owns that?  just wondering if that can be abused.

derrickstolee:
> Dumb question: in one of the dialogs there is a "redirecting to localhost:51990". what is that and/or who owns that? just wondering if that can be abused.

The credential manager has that port open. I imagine that if something else has that port that the manager opens a different port and the HTML is modified accordingly. @mjcheetham to confirm.

mjcheetham:
> > Dumb question: in one of the dialogs there is a "redirecting to localhost:51990". what is that and/or who owns that? just wondering if that can be abused.
> 
> The credential manager has that port open. I imagine that if something else has that port that the manager opens a different port and the HTML is modified accordingly. @mjcheetham to confirm.

Correct. GCM finds a free port, binds to it (waiting for the redirect), and then sends this `localhost:<free-port>` as the redirect URI to the server. GitHub (and others) allow loose matching of the port of redirect URIs when it's localhost.
The app is registered at GitHub as having a redirect URI of `http://localhost` which is valid for `http://localhost:*`.

jeffhostetler:
So (just to complete my mental picture), GCM listens on a localhost socket, spawns a headless browser and points it at the socket and sends the browser the login dialog html/css.  when the user presses login/ok/whatever, the browser sends the "form" content (or challenge) over SSL to the remote server and the remove server sends a 302 with a yes or no payload/body that redirects back to the existing GCM instance.  the GCM then sends a 200 response to the browser instance (with an OK or a try again page).  In the case of an OK it also returns the creds to the foreground Git instance.

my concern was that a bad guy could kill the localhost server after the browser had started and hijack the port and capture the 302 response from the remote server.  but it sounds like the creds aren't present in that response and wouldn't help them.

mjcheetham:
> So (just to complete my mental picture), GCM listens on a localhost socket, spawns a headless browser and points it at the socket and sends the browser the login dialog html/css. when the user presses login/ok/whatever, the browser sends the "form" content (or challenge) over SSL to the remote server and the remove server sends a 302 with a yes or no payload/body that redirects back to the existing GCM instance. the GCM then sends a 200 response to the browser instance (with an OK or a try again page). In the case of an OK it also returns the creds to the foreground Git instance.

Pretty much, yes. For some context here is the abstract flow copied from RFC 6749:
```
+--------+                               +---------------+
|        |--(A)- Authorization Request ->|   Resource    |
|        |                               |     Owner     |
|        |<-(B)-- Authorization Grant ---|               |
|        |                               +---------------+
|        |
|        |                               +---------------+
|        |--(C)-- Authorization Grant -->| Authorization |
| Client |                               |     Server    |
|        |<-(D)----- Access Token -------|               |
|        |                               +---------------+
|        |
|        |                               +---------------+
|        |--(E)----- Access Token ------>|    Resource   |
|        |                               |     Server    |
|        |<-(F)--- Protected Resource ---|               |
+--------+                               +---------------+
```

And since that's pretty abstract indeed, here's what's actually going on:
```
+--------+               +-------+
|        |--(B)- Open -->| User  |----------------- +
|  GCM   |               | Agent |<----------+      |
|        |               +-------+          (D)    (C)
| +------+                 |  ^            Authz  Authz
| |      |<-(E)- Redirect -+  |            Grant Request
| |Socket|      w/ authcode   |              |      |
| |      |--(F)-- HTML OK ----+              |      V
| +------+        Response               +---------------+
|        |                               | Authorization |
|        |                               |    Server     |
|        |--(G)-- Authorization Grant -->|               |
|        |                               |               |
|        |<-(H)----- Access Token -------|               |
|        |                               +---------------+
+--------+
  ^   |
 (A) (I)
  |   V
+--------+
|        |                               +---------------+
|  Git   |--(J)----- Access Token ------>|     Git       |
|        |                               |    Remote     |
|        |<-(K)--- Git objects/packs ----|               |
+--------+                               +---------------+
```

> my concern was that a bad guy could kill the localhost server after the browser had started and hijack the port and capture the 302 response from the remote server. but it sounds like the creds aren't present in that response and wouldn't help them.

The redirect response contains an "authorization code" (the grant) which can be redeemed for an access token. If a bad guy captured that auth code, yes they could redeem the access token. There is an extension to the OAuth2 protocol named Proof Key for Code Exchange (PKCE, pronounced "pixy") (RFC7636) which mitigates this by including some extra data in a challenge/response to ensure the entity requesting the authorization grant is also the one redeeming the code for the token.

```
                                          +-------------------+
                                          |   Authz Server    |
+--------+                                | +---------------+ |
|        |--(A)- Authorization Request ---->|               | |
|        |       + t(code_verifier), t_m  | | Authorization | |
|        |                                | |    Endpoint   | |
|        |<-(B)---- Authorization Code -----|               | |
|        |                                | +---------------+ |
| Client |                                |                   |
|        |                                | +---------------+ |
|        |--(C)-- Access Token Request ---->|               | |
|        |          + code_verifier       | |    Token      | |
|        |                                | |   Endpoint    | |
|        |<-(D)------ Access Token ---------|               | |
+--------+                                | +---------------+ |
                                          +-------------------+
```

I have _not_ implemented PKCE. GitHub does _not_ support PKCE.

jeffhostetler:
wow!  thanks for the explanation.  there's a lot going on there.

mjcheetham:
@jeffhostetler I've implemented the auth code hijacking protection PKCE in #102 for peace of mind.
If and when the server-side supports PKCE, the client will become more secure automagically.

gwestersf:
Looks great. Do you need any support from us?

PR #102, opened 2020-04-22
Implement RFC 7636 PKCE in OAuth client and fix redirect URI bug

mjcheetham:
Implement the Proof Key for Code Exchange (RFC 7636) specification in the OAuth2 client. This helps prevent hijacking of the authorisation code by malicious applications running on the user's machine.

Also fix a bug that was identified in the use of redirect URIs between the authorisation and token endpoints. If the redirect URI is "localhost" without an explicit port number, the port number we find and open _must_ be included in the redirect URI sent to the token endpoint (i.e, the redirect URI containing the generated port number used in the token endpoint call).


src/shared/Microsoft.Git.CredentialManager/Authentication/OAuth/OAuth2CryptographicGenerator.cs
```diff
@@ -0,0 +1,107 @@
...
+            // for maximum entropy. At the same time we want to avoid using the padding
+            // character '=' because this character is percent-encoded when used in URLs.
+            // To avoid padding we need the number of input bytes to be divisible by 3.
+            //
+            // In order to achieve 128 base64url characters AND avoid padding we should
+            // generate exactly 96 random bytes. Why 96 bytes? 96 is divisible by 3 and:
+            //
+            //   96 bytes -> 768 bits -> 128 base64url characters (6 bits per character)
+            //
+            var buf = new byte[96];
```
bgavrilMS, line 73: Yes, this is also how we do it https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/94e6e2e3baf840b9e06c8afbe2197d8ec0c0871f/src/client/Microsoft.Identity.Client/Platforms/netcore/NetCoreCryptographyManager.cs#L21



But we use 32 bytes... and perform the [padding logic](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/94e6e2e3baf840b9e06c8afbe2197d8ec0c0871f/src/client/Microsoft.Identity.Client/Utils/Base64UrlHelpers.cs#L69) - base64url encoding (which is classical base64 + padding) is very common in OAuth ...

src/shared/Microsoft.Git.CredentialManager/Authentication/OAuth/OAuth2SystemWebBrowser.cs
```diff
@@ -42,55 +42,35 @@ public OAuth2SystemWebBrowser(OAuth2WebBrowserOptions options)
             _options = options;
         }
 
-        public async Task<Uri> GetAuthenticationCodeAsync(Uri authorizationUri, Uri redirectUri, CancellationToken ct)
+        public Uri UpdateRedirectUri(Uri uri)
```
bgavrilMS, line 45: This is so much like MSAL ....

src/shared/Microsoft.Git.CredentialManager/Authentication/OAuth/OAuth2CryptographicGenerator.cs
mjcheetham: Is there a reason for the 32 bytes? (42.6667 base64 chars with a pad that'll always be trimmed)?

src/shared/Microsoft.Git.CredentialManager/Authentication/OAuth/OAuth2SystemWebBrowser.cs
mjcheetham: Looks like this is the only way to propogate the localhost with a 'lazy port' redirect URI from the authorisation endpoint call to the token call.

src/shared/Microsoft.Git.CredentialManager/Authentication/OAuth/OAuth2CryptographicGenerator.cs
bgavrilMS: I don't remember the reason, there was probably a reference implementation somewhere that was used. It did pass security reviews though.

bgavrilMS: But obviously your argument of using 96 bytes for more entropy is valid, this is a better implementation.

bgavrilMS: I'll log a bug on MSAL.

mjcheetham: The nice thing about 96 bytes is this works out to a base64url string that doens't need padding!
We still trim any padding characters to be doubly sure however 😁

PR #103, opened 2020-04-24
Allow GitHub OAuth params to be overridden at runtime

mjcheetham:
Allow the OAuth client ID, secret, and redirect URI to be overridden at runtime using environment variables or config.
This allows for easier testing for developers. These configuration options are not officially supported and are therefore not documented. They are for development only.

### DIFF
--- a/src/shared/GitHub/GitHubConstants.cs
+++ b/src/shared/GitHub/GitHubConstants.cs
@@ -11,6 +11,13 @@ namespace GitHub
 
         public const string AuthHelperName = "GitHub.UI";
 
+        public const string OAuthClientId = "0120e057bd645470c1ed";
+        public const string OAuthClientSecret = "18867509d956965542b521a529a79bb883344c90";
+        public static readonly Uri OAuthRedirectUri = new Uri("http://localhost/");
+        public static readonly Uri OAuthAuthorizationEndpointRelativeUri = new Uri("/login/oauth/authorize", UriKind.Relative);
+        public static readonly Uri OAuthTokenEndpointRelativeUri = new Uri("/login/oauth/access_token", UriKind.Relative);
+        public static readonly Uri OAuthDeviceEndpointRelativeUri = new Uri("/login/oauth/authorize/device", UriKind.Relative);
+
         /// <summary>
         /// The GitHub required HTTP accepts header value
         /// </summary>
@@ -51,6 +58,9 @@ namespace GitHub
         public static class EnvironmentVariables
         {
             public const string AuthenticationModes = "GCM_GITHUB_AUTHMODES";
+            public const string DevOAuthClientId = "GCM_DEV_GITHUB_CLIENTID";
+            public const string DevOAuthClientSecret = "GCM_DEV_GITHUB_CLIENTSECRET";
+            public const string DevOAuthRedirectUri = "GCM_DEV_GITHUB_REDIRECTURI";
         }
 
         public static class GitConfiguration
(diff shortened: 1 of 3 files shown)


src/shared/GitHub/GitHubConstants.cs
```diff
@@ -51,13 +58,19 @@ public static class OAuthScopes
         public static class EnvironmentVariables
         {
             public const string AuthenticationModes = "GCM_GITHUB_AUTHMODES";
+            public const string OAuthClientId = "GCM_GITHUB_CLIENTID";
```
jeffhostetler, line 61: The cover letter says that these are for testing only and not officially supported.
I'm wondering if we want to encode that in the env var name.  Git uses GIT_TEST_*
for test-only vars and GIT_* for supported vars, for example.

src/shared/GitHub/GitHubOAuth2Client.cs
```diff
@@ -1,33 +1,72 @@
...
+                Constants.GitConfiguration.Credential.SectionName, GitHubConstants.GitConfiguration.Credential.OAuthClientId,
+                out string clientId))
+            {
+                return clientId;
+            }
+
+            return GitHubConstants.OAuthClientId;
+        }
+
+        private static Uri GetRedirectUri(ISettings settings)
```
jeffhostetler, line 46: This would let me route the redirect to something besides localhost.  I can't express
whether this bothers me or whether I'm just being paranoid.  I tried to walk thru the
MITM diagrams from the other day, but I'm in over my head I fear.  It may not be a
problem, but I don't know that.

I'm wondering what you would use this env var for in testing -- if as I understand it,
the responses need to come back into the builtin GCM server to advance the state
machine.  It seems like this would route the redirect somewhere else and you wouldn't
get responses back in the GCM.  But again, I don't completely understand the flow.

Alternatively, would it be safer to have this env var be under a #if DEBUG ?

mjcheetham: Redirect URIs in OAuth2 are tied to the client ID. That is to say each registered application must define at least one redirect URI, and during the various endpoint calls the client must send a  redirect URI that is associated with that app.

For example GCM just uses `http://localhost/` as the only redirect URI (loopback hosts are special cased in that the port is lazily matched; otherwise port must `===` too).

Other applications such as GitHub CLI use `http://127.0.0.1/callback` as the redirect URI (it includes the path `/callback`).

Also note that `127.0.0.1 != localhost` textually, so these are considered different redirect URIs.

> This would let me route the redirect to something besides localhost.

In theory yes, in practice no since the `OAuth2SystemWebBrowser` component that we use in GCM (the component that launches the browser, and starts the HTTP listener for the redirect) mandates that the redirect URI `Uri::IsLoopback` [here](https://github.com/microsoft/Git-Credential-Manager-Core/blob/7efe85294f661cb42fb336873355749c3d3c5d37/src/shared/Microsoft.Git.CredentialManager/Authentication/OAuth/OAuth2SystemWebBrowser.cs#L47-L50) and [here](https://github.com/microsoft/Git-Credential-Manager-Core/blob/7efe85294f661cb42fb336873355749c3d3c5d37/src/shared/Microsoft.Git.CredentialManager/Authentication/OAuth/OAuth2SystemWebBrowser.cs#L64-L67).

So we won't end up gifting `http://evil-server.com` with a authorisation code because:
1. We won't even start the auth flow via the browser unless we have a local redirect URI,
2. The server would need to agree to send the code to the redirect URI.

The 'problem' that might occur however is redirecting to _another_ localhost application listening on with a different path prefix. This is 'fixed' with PKCE as mentioned and implemented by GCM previously, but that requires server support.

src/shared/GitHub/GitHubConstants.cs
mjcheetham: That’s a good idea. I like GCM to try and stay consistent with git practices.

Issue #37, opened 2019-04-29
Add a BitBucket host provider


mjcheetham:
GCM Windows supports BitBucket as a Git host provider.. GCM Core should provide some support for this provider.

On Windows there is custom Windows-only UI that could/should be ported. On macOS we can use a TTY-based prompt (like we use for GitHub currently) for now.

mminns:
Fantastic you've merged this.

I was playing with this myself. While it worked I've never had the time to polish it.

https://github.com/itofinity/Git-Credential-Manager-Core/tree/issue/37-Bitbucket-Support


I have a GUI component written in Avalonia that works cross platform as well
https://github.com/itofinity/Atlassian-Authentication-Helper

this is a screen shot from macOS
![Annotation 2020-05-15 133542](https://user-images.githubusercontent.com/815107/82051034-12fa6900-96b1-11ea-9c37-6af5433acc03.png)

mminns:
FWIW running this on Windows for Bitbucket the OAuth success page isn't being displayed.

![Annotation 2020-05-15 210235](https://user-images.githubusercontent.com/815107/82091791-e87bd080-96ef-11ea-8c9e-c64311784189.png)

mjcheetham:
Hi @mminns,

> Fantastic you've merged this.
> 
> I was playing with this myself. While it worked I've never had the time to polish it.
> 
> https://github.com/itofinity/Git-Credential-Manager-Core/tree/issue/37-Bitbucket-Support

It's awesome to see interest in getting Bitbucket into GCM Core 😃
Looks like we've duplicated a lot of work here, which is a shame 😢

> I have a GUI component written in Avalonia that works cross platform as well
> https://github.com/itofinity/Atlassian-Authentication-Helper
> 
> this is a screen shot from macOS
> ![Annotation 2020-05-15 133542](https://user-images.githubusercontent.com/815107/82051034-12fa6900-96b1-11ea-9c37-6af5433acc03.png)

This sounds like an interesting path forward for x-plat UI. I've never really used or explored much with Avalonia before. I assume Avalonia is pretty stable and somewhat compatible with existing WPF-flavoured XAML?

As you can see from #116 I'm in the process of getting a WPF/Windows only UI in for Bitbucket, based on the GCM for Windows version but with updated iconography and colours. I wonder if we can port that to Avalonia?

> FWIW running this on Windows for Bitbucket the OAuth success page isn't being displayed.

[...]

PR #104, opened 2020-04-27
Add Bitbucket support (text prompts only)

mjcheetham:
Add a host provider to support username/password and OAuth authentication to Atlassian Bitbucket.
The auth flow is based on the logic from GCM Windows, and follows the following chart:

```
            Yes   +---------+   No
     +------------+ Has RT? +-------------+
     |            +---------+             |
     v                                    v
+----+----+                        +------+------+
|  OAuth  |   Fail                 |     Has     | No
| Refresh +---------+              | credential? +----+
+---+-----+         |              +--+----------+    |
    |               |                 |               |
OK  |               |             Yes |               |
    v               v                 v               v
+---+---+       +---+----+   Yes   +--+---+    +------+-+
| Store +<------+ OAuth  +<--------+ 2FA? +<---+  U/P   |
|  RT   |       | Prompt |         +--+---+    | Prompt |
+---+---+       +--------+            |        +--------+
    |                                 | No
    v                                 v
+---+----+                         +--+---------+
| Return |                         |   Return   |
|   AT   |                         | credential |
+--------+                         +------------+

```

- We first check for the presence of an OAuth refresh token, and if it exists we use it to create a fresh access token.
- If we fail to use the refresh token then we restart the OAuth flow.
- If we do not have a refresh token then we try and use any existing credential and check if 2FA is required.
- If 2FA is required we start an OAuth flow.

There are no GUI prompts in this PR; text-based prompts are used instead. A port of the Windows Bitbucket UI will follow this PR.

Goes towards implementing #37


src/shared/Atlassian.Bitbucket/BitbucketConstants.cs
```diff
@@ -0,0 +1,42 @@
...
+
+namespace Atlassian.Bitbucket
+{
+    public static class BitbucketConstants
+    {
+        public const string BitbucketBaseUrlHost = "bitbucket.org";
+        public static readonly Uri BitbucketApiUri = new Uri("https://api.bitbucket.org");
+        public const string BitbucketAuthHelperName = "Atlassian.Bitbucket.UI";
+
+        public const string OAuth2ClientId = "b5AKdPfpgFdEGpKzPE";
```
mjcheetham, line 13: This is a new application "Git Credential Manager" still 'owned' by myself. We can swap this out for the app ID used by the old GCM, which was the same as Sourcetree's! I'd like to try and reach out and whitelist our application (and make it more Microsoft 'official').

derrickstolee: and this is registered through BitBucket?

mjcheetham: Yes. This is registered through Bitbucket. I've started the process of transferring the GitHub GCM app to a verified organisation (following MS policy), so I'd need to do the same for Bitbucket, but that's new ground AFAIK. Perhaps the Azure Pipelines folks know more here?
The alternative is to continue to 'steal' Sourcetree's ID?

mjcheetham: I'll continue to use Sourcetree's values until we can do this.

mminns:
I wrote the original GCMW implementation, I'd be happy to help test this.

mjcheetham:
> Are there any existing build artefacts are should I just build from master?

Hi @mminns! 
You'll probably want to build from a branch in my fork (there is a bug where I forgot to include the actual Bitbucket host provider DLL needed! 🤦) until this PR is completed: #117 

https://github.com/mjcheetham/Git-Credential-Manager-Core/tree/fix-bb

## Windows
1. Build the Installer.Windows project in the WindowsDebug configuration inside VS
2. Look in `out/windows/Installer.Windows/bin/Debug` and you'll find the exe installer

## Mac
1. Run `dotnet build -c MacDebug` from the root of the repo.
2. Look in `out/osx/Installer.Mac/pkg/Debug` and you'll find the pkg installer

This build only has a command-line auth prompt experience and not the WPF-based UI. This is coming in this PR: #116.

PR #120, opened 2020-05-18
helpers: allow auth helpers to be overridden by config

### LINKED ISSUE #119: Introduce a mechanism for plugging in external auth UI helpers
At the moment GCM Core can shell out to external 'helper' applications that display authentication GUIs to capture input and notify the user of an OAuth flow. The name of the helpers are hard-coded, and the binaries must exist next to the main GCM Core binary on disk. This makes it hard to replace or add helpers not shipped in-box.

**Proposal**

1. Introduce a configurable 'search path' setting (environment variable?) that external helpers can plug in to let GCM Core discover them.
2. Add a setting to override the name of the UI helper for each host provider (e.g., `credential.githubAuthHelper = gh-cl-helper`)
3. 3rd party helpers can add themselves on the 'helper search path' and set the `credential.*AuthHelper` option

**Issues**

1. Malicious apps could insert themselves in the search path and capture/send-home input secrets. How can we verify or trust which application we're launching? Is this onus on the user?
2. Multiple helpers could start fighting over the order of the search path setting, which might be confusing/annoying for users. Do we need to take more control over this? `git-credential-manager-core install-plugin <x>`? Too complicated?

mjcheetham:
Allow authentication UI helpers to be overridden by environment variables and Git configuration options.

Closes #119

### DIFF
--- a/src/shared/GitHub/GitHubAuthentication.cs
+++ b/src/shared/GitHub/GitHubAuthentication.cs
@@ -3,9 +3,7 @@
 using System;
 using System.Threading.Tasks;
 using System.Collections.Generic;
-using System.IO;
 using System.Net.Http;
-using System.Reflection;
 using System.Text;
 using System.Threading;
 using Microsoft.Git.CredentialManager;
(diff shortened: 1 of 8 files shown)


src/shared/GitHub/GitHubAuthentication.cs
```diff
@@ -231,29 +229,11 @@ public async Task<OAuth2TokenResult> GetOAuthTokenAsync(Uri targetUri, IEnumerab
...
-            }
-
-            string executableDirectory = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
-            path = Path.Combine(executableDirectory, helperName);
-            if (!Context.FileSystem.FileExists(path))
-            {
-                Context.Trace.WriteLine($"Did not find helper '{helperName}' in '{executableDirectory}'");
-
-                // We currently only have a helper on Windows. If we failed to find the helper we should warn the user.
-                if (PlatformUtils.IsWindows())
```
dscho, line 248: I guess that losing this part isn't dramatic, eh?

mjcheetham: Not really, as it points to an installation issue which is only resolved by reinstalling. There is a fallback for scenarios when you have terminal access.

dscho: Okay!

PR #122, opened 2020-05-22
release: add required token to release workflow

mjcheetham:
Add the required GITHUB_TOKEN token to the release workflow.


mjcheetham:
As you can see from the run [here](https://github.com/microsoft/Git-Credential-Manager-Core/runs/699462770?check_suite_focus=true), we're missing the `github-token` input.

PR #126, opened 2020-05-22
Don't attempt to deserialize nulls from bitbucket response

kzu:
In my case, the field `has_2fa_enabled` was coming back null and
that caused deserialization to fail with:

```
Error converting value {null} to type 'System.Boolean'. Path 'has_2fa_enabled', line 1, position 79.
```

Another alternative is to make the property `bool?` instead.


kzu:
It's from Bitbucket.org, also, it didn't matter if I had 2fa enabled or not, it consistently failed :/

PR #127, opened 2020-05-23
homebrew: fix automatic Homebrew update

mjcheetham:
The Octokit API does not find releases by tag when they are prerelease (like all the beta releases of GCM Core are).

Instead enumerate the list of releases and manually filter to the matching tag name instead.


mjcheetham:
> I had a look at https://developer.github.com/v3/repos/releases/#get-a-single-release and could not find out a better way, either.

Yep. It's a shame but there seems to be a server-side filtering of pre-release release tags.
Once we move from the 'beta' moniker for GCM Core we can revert back to the simpler 'get release with tag' call, and drop the prerelease from the pipelines that create the GitHub releases.

PR #131, opened 2020-06-09
windows: compile to 32-bit x86 only

### LINKED ISSUE #112: Build 32-bit version of GCM Core for Windows
We currently only build a 64-bit version of GCM Core for Windows. We should also build a 32-bit version.
LibGit2 is our only native dependency.

mjcheetham:
Since we want to ship with Git for Windows which offers both an x86 and x86_64 option, we should build an x86 (32-bit) application rather than an x86_64 (64-bit) one. Since 32-bit applications can run on 64-bit versions of Windows (including Windows on ARM) this is the more compatible target.

We could also publish both a 32- and 64-bit build of GCM Core on Windows but this seems redundant as GCM Core really does not need 64-bits of addressable memory(!)

Fixes #112


Stanzilla:
I'd vote for providing both

Issue #47, opened 2019-06-05
Create new client ID for GCM when speaking to Azure Repos


mjcheetham:
We are currently using the same client ID as Visual Studio for ADAL (this is what GCM Windows also used). We should move to our own client ID, and have it join the Microsoft FoCI.

Question(s): should we join FoCI and become a 1st party app or not? Is consent not something we want to show?

vtbassmatt:
@mjcheetham this is only for Azure Repos, right? If so, I'll specify that in the title and we'll keep it open but marked "inactive" until we hear from the Repos team.

mjcheetham:
Correct. Azure Repos only. Updated the title.

kyle-rader:
Watching

mjcheetham:
In order to correctly support WAM (Windows broker) we require a new client ID/first-party application. Removing the `inactive` tag from this issue.

kyle-rader:
How does this impact WAM? We use WAM today with the VS id without issue generally.

mjcheetham:
We believe there is an issue using MSAs with the existing VS client ID as it was never enabled in the backend, and changing it after the fact is not possible. VS are moving to a new client ID also.

cc: @bgavrilMS

bgavrilMS:
@kyle-rader - VS uses a different client_id to talk to WAM. Most apps that use **MSA-passthrough** are not correctly configured to talk to WAM. Changing existing config is deemed too risky in some cases. New client id is better path forward.

vtbassmatt:
Necro'ing this oldest issue in the GCM repo :)

I'm on the engineering systems team for a large product group at Microsoft. We're trying to improve our Mac users' experience, and copious "random auth prompts in a browser" has emerged as both a friction point and potential security issue†. I understand from #2043 that we need to add some additional redirect URIs to the app registration, which likely isn't going to be possible on the existing VS registration. Is it possible to get some renewed interest in a separate ADO/GCM registration? Thanks!

†<small>As in, are we inadvertently training our users to click yes and do the 2FA whenever some browser window pops open?</small>

bgavrilMS:
@vtbassmatt - can you please reach out to me directly ? bogavril

vtbassmatt:
Will do!

mjcheetham:
I'm happy to announce we're now finally ready to switch to our own client ID!

https://github.com/git-ecosystem/git-credential-manager/pull/2416

vtbassmatt:
Woo!

PR #148, opened 2020-07-17
Fix Azure Repos sign-in with Microsoft Accounts

### LINKED ISSUE #143: Cannot sign in with MSA against MSA AzDO org
When I try to push to https://dev.azure.com/git-for-windows/git-sdk-64, I am asked to sign in with my account (which is an MSA one). The dialog then complains "You can't sign in here with a personal account. Use your work or school account instead." (But obviously, when I _do_ use my work account, the push fails due to lack of permissions.)

mjcheetham:
Use the /organizations authority for MSA accounts with Azure DevOps/Repos. This is because we're using MSA pass-through, an internal Microsoft mechanism to support both MSA and 'work' (AAD) accounts with the same auth stacks.

You should be able to use /common, but this doens't work. At the same time we're using ADAL Obj-C on macOS rather than MSAL.NET like we do on Windows, and ADAL speaks to the "v1" AAD endpoints, which don't know the /organizations tenant 😢 

For macOS we need to fudge the authority _back_ to /common for MSA accounts.

The "correct" fix here is to move from MSA pass-through, us our own client ID (#47) and drop the ADAL Obj-C component and add UI on .NET Core for MSAL.NET... all these things would take a _long_ time to do.

Fixes #143

### DIFF
--- a/src/osx/Microsoft.Authentication.Helper.Mac/Source/main.m
+++ b/src/osx/Microsoft.Authentication.Helper.Mac/Source/main.m
@@ -100,6 +100,23 @@ int main(int argc, const char * argv[]) {
         NSString* redirectUri = [configs objectForKey:@"redirectUri"];
         NSString* interactive = [configs objectForKey:@"interactive"];
 
+        // Because ADAL only supports the v1 endpoints we need to transform any request
+        // for the /organizations or /consumers authority to the /common one or else
+        // we get errors back from the server.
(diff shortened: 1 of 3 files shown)


mjcheetham:
The PR build was in fact green, but GitHub Checks API and Azure Pipelines was not having a good day... 😢 
https://dev.azure.com/mseng/AzureDevOps/_build/results?buildId=12622663&view=results

PR #152, opened 2020-07-21
Automatically update Homebrew Cask on release

mjcheetham:
On release, automatically update the Homebrew Cask in the microsoft/git Tap.

### DIFF
--- a/.github/workflows/release-homebrew.yaml
+++ b/.github/workflows/release-homebrew.yaml
@@ -1,34 +1,18 @@
 name: "release-homebrew"
 on:
   release:
-    types: [published]
+    types: [released]
 
 jobs:
   release:
     runs-on: ubuntu-latest
     steps:
-    - name: Get macOS package version
-      uses: actions/github-script@0.3.0
-      id: version
-      with:
-        github-token: ${{secrets.GITHUB_TOKEN}}
-        result-encoding: string
-        script: |
-          const { data } = await github.repos.getReleaseByTag({
-            owner: context.repo.owner,
-            repo: context.repo.repo,
-            tag: process.env.GITHUB_REF
-          });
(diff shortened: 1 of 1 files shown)


.github/workflows/release-homebrew.yaml
```diff
@@ -1,34 +1,18 @@
...
-      uses: mjcheetham/update-homebrew@v1
+      uses: mjcheetham/update-homebrew@v1.1
       with:
         token: ${{ secrets.HOMEBREW_TOKEN }}
         tap: microsoft/git
         name: git-credential-manager-core
         type: cask
-        version: ${{ steps.version.outputs.result }}
+        releaseAsset: gcmcore-osx-(.*)\.pkg
+        alwaysUsePullRequest: true
```
mjcheetham, line 18: There's a question if we want to go ahead and push a commit to the `master` branch in the Tap, **_OR_** create a PR. Creating a PR gives us one last 'manual check' option.

derrickstolee: Let's start with a PR and see how that goes. Thanks!

If we find that this is stable and we don't want the extra manual step, then we can change this for a later release.

.github/workflows/release-homebrew.yaml
```diff
@@ -1,34 +1,18 @@
 name: "release-homebrew"
 on:
   release:
-    types: [published]
+    types: [released]
```
derrickstolee, line 4: Does this trigger on `pre-release` to `release` instead of publishing a draft to `pre-release`?

mjcheetham: So right now, this change will mean the Homebrew PR only gets created when doing a `release`/not-a-draft/not-a-pre-release.

PR #157, opened 2020-08-06
User/kyrader/ci

kyle-rader:
Starting the creation of a Github Actions CI workflow, specifically including Linux.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -1,26 +1,48 @@
 name: Continous Integration
+
 on:
+  push:
+    branches: [ linux ]
   pull_request:
-    branches:
-      - master
-      - linux
-  workflow_dispatch:
+    branches: [ linux ]
 
 jobs:
   build_and_test_ubuntu:
-    runs-on: ubuntu-latest
     name: "Build & Test: Ubuntu"
+    runs-on: ubuntu-latest
     steps:
-      - name: Hello Linux
-        run: |
-          echo "Hello World Github Actions"
-          echo "Running on: $(uname -a)"
+    - name: Hello Linux
+      run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,24 @@
+name: Continous Integration
+on:
+  pull_request:
+    branches:
+      - master
```
derrickstolee, line 5: include `linux` here.

derrickstolee: I wonder if we need to merge this in order to enable it.

kyle-rader: That's what I'm guessing. And I fear it needs to merge into master. I recalled we talked about merging `linux` in to `master`. I think this would simplify our CI setup and help us move from one functional point to another. But if this wouldn't work with he current CI / shipping process we can relegate to the `linux` branch. Given there's not a GH actions - is there an Azure pipeline check that runs Windows CI?

derrickstolee: There is an Azure Pipeline, but it's probably only checked on PRs to `master`.

derrickstolee, line 5:
```suggestion
      - master
      - linux
```

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,24 @@
+name: Continous Integration
+on:
+  pull_request:
```
derrickstolee, line 3:
```suggestion
on:
  push
  pull_request:

```

derrickstolee:
cc: @chrisd8088. Here is just an example of a `.yml` file that should be similar to what you need in microsoft/scalar.

Issue #160, opened 2020-08-20
Add username to target name of GitHub credential


gaborcsardi:
Currently it is cumbersome to handle multiple GitHub credentials with the github provider. One solution to this would be to include the username in the target name, just like it is included for the generic provider.

Relevant discussion: https://github.com/microsoft/Git-Credential-Manager-for-Windows/issues/749

PR to fix the same issue in Git Credential Manager for Windows: https://github.com/microsoft/Git-Credential-Manager-for-Windows/pull/891

Thanks much for considering this!

mjcheetham:
Hi @gaborcsardi! Thanks for opening this issue and your interest in GCM Core.

We are looking at changing the way we store and recall stored credentials in general in GCM Core, specifically such that multi-user scenarios like this are supported better. Unfortunately to get the best experience a change in how Git interacts with credential helpers (like GCM Core) is likely to be required.

We are currently looking at submitting such a change to Git, as well as an overhaul in GCM Core to better support scenarios as best we can without breaking backwards compatibility when used on down-level versions of Git (without our proposed patches), and to take advantage of a Git that may support our patches to 100% support all scenarios.

Note: the change we're looking at introducing to Git is to have Git echo _all_ `key=value` pairs that are returned over stdout, not just the ones known to Git internally. This will allow us to preserve some state between the independent calls Git makes to the helpers (`get` followed by a `store/erase`) in an authentication flow.

I will update this issue with more information once we have it 😄

gaborcsardi:
@mjcheetham Thanks a lot for the update! Makes sense.

FWIW I did a quick review on how the current (commonly used) helpers support multiple credentials, it is here: https://github.com/r-lib/gitcreds/blob/master/vignettes/gitcreds-advanced.Rmd

mjcheetham:
Awesome write up! That's a useful summary of the differing behaviours of helpers in the wild.
As I mentioned we're hoping to match more closely with the wincred or osxkeychain models, but handling multi-user accounts a little better.

mjcheetham:
Hi @gaborcsardi! I have an update for you here.

We've got a pre-release out now which should allow multiple users for the same host provider, such as GitHub.

https://github.com/microsoft/Git-Credential-Manager-Core/releases/tag/v2.0.246-beta

[...]

Issue #156, opened 2020-07-30
`fatal:invalid URI: failed to parse host name.` when access an repository with port in url


yyjdelete:
**Which version of GCM Core are you using?**

From a terminal, run `git-credential-manager-core version` and paste the output.
>Git Credential Manager version 2.0.194-beta+819e6bc120 (Windows, .NET Framework 4.0.30319.42000)
>git version 2.28.0.windows.1

**Which Git host provider are you trying to connect to?**

* [ ] Azure DevOps
* [ ] Azure DevOps Server (TFS/on-prem)
* [x] GitHub
* [ ] GitHub Enterprise
* [ ] Bitbucket
* [x] Other - please describe

**Can you access the remote repository directly in the browser using the remote URL?**
It need to be an repo needed auth(or test push instead?)

From a terminal, run `git remote -v` to see your remote URL.
>https://github.com:443/yyjdelete/zzzz
or any other private repository with non-default port (http or https), the above one with 443 is just used as an quick test.

* [ ] Yes
* [ ] No, I get a permission error
* [x] No, for a different reason - please describe

When I try to access an repository with port in url which need auth , I see an new fatal shown after update to git-for-windows 2.28 and switch to GCM-Core(with `credential.helper=manager-core`)

---

**Expected behavior**

No `fatal` message should be shown.

**Actual behavior**

>fatal:invalid URI: failed to parse host name.

**Logs**

Set the environment variables `GCM_TRACE=1` and `GIT_TRACE=1` and re-run your Git command. Review and redact any private information and attach the log.
```
git clone https://github.com:443/yyjdelete/zzzz 3
16:58:07.131533 exec-cmd.c:237          trace: resolved executable dir: D:/Tools/Git/mingw64/bin
16:58:07.133519 git.c:444               trace: built-in: git clone https://github.com:443/yyjdelete/zzzz 3
Cloning into '3'...
16:58:07.152508 run-command.c:663       trace: run_command: git remote-https origin https://github.com:443/yyjdelete/zzzz
16:58:07.176495 exec-cmd.c:237          trace: resolved executable dir: D:/Tools/Git/mingw64/libexec/git-core
16:58:07.177511 git.c:704               trace: exec: git-remote-https origin https://github.com:443/yyjdelete/zzzz
16:58:07.177511 run-command.c:663       trace: run_command: git-remote-https origin https://github.com:443/yyjdelete/zzzz
[...]

Issue #142, opened 2020-07-01
Can't get basic auth credentials without supplying a username


mkincaid:
**Which version of GCM Core are you using?**

Git Credential Manager version 2.0.164-beta+1264f234c7 (Windows, .NET Framework 4.0.30319.42000)

(Can also reproduce on macOS)

**Which Git host provider are you trying to connect to?**

HTTP(S) basic authentication

**Expected behavior**

```
>git-credential-manager-core store
protocol=https
host=host.name
username=foo
password=bar

>git-credential-manager-core get
protocol=https
host=host.name

protocol=https
host=host.name
username=foo
password=bar
```

**Actual behavior**

```
>git-credential-manager-core get
protocol=https
host=host.name

Enter credentials for 'https://host.name/'
```
This case of needing to fill the username is very common. For example, the series of interactions with Git described above happens if the remote is defined as `https://host.name/path.git` (as opposed to `https://foo@host.name/path.git`). That means that, in practice, GCM Core appears to the user not to be able save credentials for https remotes without `username@` in the URI (actually, they do get saved just fine, but under a different key than the one that’s used to try to retrieve them the next time).

**Possible resolutions**

The quandary arises from the implicit design assumption that any `get` request carries enough information to generate a unique, stable credential key that can be directly retrieved from the store. In the case of basic authentication without knowledge of a username, that doesn't hold.

In [GCM for Mac and Linux](https://github.com/microsoft/Git-Credential-Manager-for-Mac-and-Linux/blob/7f954d807e45be91798420ba54475b007ee5862f/src/main/java/com/microsoft/alm/gitcredentialmanager/OperationArguments.java#L67) the targetURI is always constructed, i.e. both retrieved and saved, as ``protocol://host`` (never including the username). This implementation works, but removes some flexibility (e.g. the option to save credentials for different usernames on the same host and distinguish among them by specifying username@ in the remote URL).

[...]

PR #162, opened 2020-09-09
Revamp credential storage system to allow for smarter searching

mjcheetham:
This PR should hopefully address several bugs which originate from the (flawed) design of the credential storage abstraction ( https://github.com/microsoft/Git-Credential-Manager-Core/issues/142, https://github.com/microsoft/Git-Credential-Manager-Core/issues/160) as well as a simple port number parsing bug (https://github.com/microsoft/Git-Credential-Manager-Core/issues/156)

The original design/assumption was that for each request we get from Git (via a `get` call) we could create a unique 'credential key' that encoded the username and URL for the hosting provider. Whilst this worked for GitHub, Azure Repos, and Bitbucket when using _one user_ per provider (or AzDevOps organisation), it failed to work for the generic provider when the username was not specified as part of the remote URL, nor did it work for multi-user scenarios with the other providers.

Following the design of the `osxkeychain` and `wincred` Git credential helpers, we now store the username in a separate field from the URL (called a 'service' to use the macOS Keychain terminology). This means when doing a credential lookup in response to a `get` call, we can search for a credential with the matching URL (service), and optionally a matching account if a username was passed, or _any_ user if none was specified.

One problem however arises from this multi-user support which is specific to the Azure Repos provider..

**Background:** Azure Repos supports two kinds of URL: `org.visualstudio.com` and `dev.azure.com/org`. In the former we have all the information we need to perform authentication in the authority (protocol + hostname + orgname), but with the latter we require at least the first component of the path to get the organisation name.

Git only supplies the authority information (protocol + hostname) by default, meaning we have to force the `useHttpPath` Git configuration option to `true` meaning we always get the path component. This also meant that for Azure Repos users using the dev.azure.com-style URL we could not differentiate between a real desire to store credentials against the full remote URL including path or not. This was already an existing limitation.

However, to help 'workaround' this URL+Git design incompatibility, Azure Repos also made the decision to recommend users put the AzDevOps organisation name as the _user info_ in the remote URL for dev.azure.com URLs. For example: `https://org@dev.azure.com/org/project/_git/repo`. Previously we ignored the username (since we only supported one user), but now we have a problem with storing credentials against the _real_ username vs the 'fake' org username value.

Without a change to Git to either: a) always provide the full remote URL on each `get` request as a separate input line, or b) provide a way to maintain state between the `get` and `store/erase` calls, we cannot implement multi-user _nor_ full-path support for `dev.azure.com` users. For now, we maintain the existing behaviour of one _one user_ and _no full path_ storage support for credentials against the `dev.azure.com`-style URLs.

We _do_ however implement full mutli-user and full-path-storage support for the older `*.visualstudio.com`-style URLs.
For those who wish to use multiple users per Azure DevOps organisation, or store a credential against the full remote URL, they must use the `vs.com`-style URLs which are still supported.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -168,3 +168,20 @@ git config --global credential.gitHubAuthModes "oauth basic"
 ```
 
 **Also see: [GCM_GITHUB_AUTHMODES](environment.md#GCM_GITHUB_AUTHMODES)**
+
+---
+
+### credential.namespace
+
+Use a custom namespace prefix for credentials read and written in the OS credential store.
+Credentials will be stored in the format `{namespace}:{service}`.
+
+Defaults to the value `git`.
+
+#### Example
+
+```shell
+git config --global credential.namespace "my-namespace"
+```
+
+**Also see: [GCM_NAMESPACE](environment.md#GCM_NAMESPACE)**
--- a/src/shared/Microsoft.AzureRepos/AzureReposHostProvider.cs
+++ b/src/shared/Microsoft.AzureRepos/AzureReposHostProvider.cs
@@ -10,8 +10,9 @@ using KnownGitCfg = Microsoft.Git.CredentialManager.Constants.GitConfiguration;
 
 namespace Microsoft.AzureRepos
 {
-    public class AzureReposHostProvider : HostProvider, IConfigurableComponent
+    public class AzureReposHostProvider : DisposableObject, IHostProvider, IConfigurableComponent
     {
+        private readonly ICommandContext _context;
         private readonly IAzureDevOpsRestApi _azDevOps;
         private readonly IMicrosoftAuthentication _msAuth;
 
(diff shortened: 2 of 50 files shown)


src/shared/Microsoft.Git.CredentialManager/Interop/Windows/WindowsCredentialManager.cs
```diff
@@ -9,119 +11,289 @@ namespace Microsoft.Git.CredentialManager.Interop.Windows
...
+            }
 
-                default:
-                    Win32Error.ThrowIfError(result);
-                    return false;
+            if (!string.IsNullOrWhiteSpace(account))
+            {
+                string escapedAccount = account.Replace('@', '_');
+                sb.AppendFormat("{0}@", escapedAccount);
             }
```
derrickstolee, line 284: This hunk isn't covered by tests. Seems important.

src/shared/Microsoft.Git.CredentialManager/Interop/Windows/WindowsCredentialManager.cs
```diff
@@ -9,119 +11,289 @@ namespace Microsoft.Git.CredentialManager.Interop.Windows
...
+                if (!string.IsNullOrWhiteSpace(serviceUri.AbsolutePath) &&
+                    !StringComparer.OrdinalIgnoreCase.Equals(serviceUri.AbsolutePath, targetUri.AbsolutePath))
                 {
-                    Marshal.FreeHGlobal(w32Credential.CredentialBlob);
+                    return false;
                 }
+
+                // URLs match
+                return true;
             }
```
derrickstolee, line 259: This part is also not covered.

src/shared/Microsoft.AzureRepos/AzureReposHostProvider.cs
```diff
@@ -90,28 +151,97 @@ public override async Task<ICredential> GenerateCredentialAsync(InputArguments i
...
+        /// <remarks>
+        /// For dev.azure.com-style URLs we use the path arg to get the Azure DevOps organization name.
+        /// We ensure the presence of the path arg by setting credential.useHttpPath = true at install time.
+        ///
+        /// The result of this workaround is that we are now unable to determine if the user wanted to store
+        /// credentials with the full path or not for dev.azure.com-style URLs.
+        ///
+        /// Rather than always assume we're storing credentials against the full path, and therefore resulting
+        /// in an personal access token being created per remote URL/repository, we never store against
+        /// the full path and always store with the organization URL "dev.azure.com/org".
```
derrickstolee, line 181: So we still require `credential.useHttpPath`, but we are dropping _some_ data from the URL?

docs/configuration.md
```diff
@@ -168,3 +168,20 @@ git config --global credential.gitHubAuthModes "oauth basic"
 ```
 
 **Also see: [GCM_GITHUB_AUTHMODES](environment.md#GCM_GITHUB_AUTHMODES)**
+
+---
+
+### credential.namespace
```
derrickstolee, line 174: Did this need to be in this PR, or could it be a different one?

derrickstolee (commented):
Finished my first pass. Just a few random thoughts here and there.

*Question:* What is the user experience for someone using a `dev.azure.com` URL with current GCM Core and then they upgrade? Do they now need to re-auth for every organization? Or will the existing `org@dev.azure.com` credentials still work? It is harder to be confident in such compatibility scenarios, as the tests only check the current version of the code.

docs/configuration.md
mjcheetham: The `namespace` support is needed to maintain back-compat with older GCM's credentials (they're all prefixed with `git:`). We've pushed the logic for building the names into the credential stores, rather than the host providers (who used to just always prepend `git:`).

We could hardcode `git:`, only to then pull it out in a subsequent PR as a setting of course?

src/shared/Microsoft.AzureRepos/AzureReposHostProvider.cs
mjcheetham: GCM Core has _always_ required `credential.useHttpPath` for `dev.azure.com` URLs, since we don't implement the same hackery of walking the process hierarchy and extracting the parent process' command lines like GCM for Windows does. Plus we cannot do this method at all on non-Windows systems (requires we run as `root`).

In this regard, GCM Core does not support the scenario of using the full path with `dev.azure.com` URLs whereas GCM4Windows does. I'd rather break compat here than introduce that code which when first introduced caused systems to crash due to memory leaks, infinite tree walking, and issues with PID reuse.

In 99% of cases, users of GCM4W will _not_ have `useHttpPath` set and will be storing against the org URL anyway.

The preferred fix here is to change Git itself so we can avoid forcing `useHttpPath` to true for `dev.azure.com`.

mjcheetham:
> Finished my first pass. Just a few random thoughts here and there.
> 
> _Question:_ What is the user experience for someone using a `dev.azure.com` URL with current GCM Core and then they upgrade? Do they now need to re-auth for every organization? Or will the existing `org@dev.azure.com` credentials still work? It is harder to be confident in such compatibility scenarios, as the tests only check the current version of the code.

Current GCM Core does the same thing.. for `dev.azure.com/org` URLs we only ever store against the org URL.
In fact, current GCM also does the same thing for `org.vs.com` URLs when it didn't need to, so this change is actually only increasing the number of scenarios, not cutting them.

derrickstolee (approved):
I didn't find anything new in a second read.

src/shared/Microsoft.Git.CredentialManager/ICredentialStore.cs
```diff
@@ -9,24 +9,27 @@ namespace Microsoft.Git.CredentialManager
...
-        /// <param name="credential">Credential to store.</param>
-        void AddOrUpdate(string key, ICredential credential);
+        /// <param name="service">Name of the service this credential is for. Use null to match all values.</param>
+        /// <param name="account">Account associated with this credential. Use null to match all values.</param>
+        /// <param name="secret">Secret value to store.</param>
+        void AddOrUpdate(string service, string account, string secret);
 
         /// <summary>
-        /// Delete credential from the store with the specified key.
+        /// Delete credential from the store that was previously obtained from <see cref="Query"/> or <see cref="Get"/>.
```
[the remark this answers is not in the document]
mjcheetham: Nope! Removed and updated the comment.

src/shared/Microsoft.Git.CredentialManager/Interop/Windows/WindowsCredentialManager.cs
mjcheetham: Added a new test of the username escaping in the target name.

src/shared/Microsoft.Git.CredentialManager/Interop/Windows/WindowsCredentialManager.cs
mjcheetham: This part is now exercised by the tests that creates two credentials with the same service name (so the target name will not directly equal the service name, since the target now also contains the username).

PR #164, opened 2020-09-10
Add a Linux .deb installer CI workflow

kyle-rader:
# Build a `.deb` package

## Overview
Adding another GHA workflow to build installers. Happy to rename this to specifically build the Linux installer. The other CI I made uses a matrix run to the base build and test on all platforms.

If there are platform specific tasks, like platform specific tests and building an installer we could make a platform specific workflow for each of those.

### One `build.sh`
This PR is a adding a build step but is a net reduction in bash from merging into one `build.sh` for the Payload.Linux project. I found it very confusing to have the build not happen in `build.sh`. The duplicated arg parsing in the 3 bash scripts seems like an un-needed 3x amount of potential parsing bugs plus each script was redefining several vars `build.sh` already defined.  

## Future work
I'm figuring out internally how to onboard this project to the ESRP signing process so we can publish to [packages.microsoft.com](https://packages.microsoft.com). I expect I'll have to make more edits for that process specifically to either download the signing client and run it or send the bits off to be signed. I don't yet know if we can do this in GHA.


src/linux/Payload.Linux/build.sh
```diff
@@ -32,17 +38,106 @@ case "$i" in
...
+mkdir -p "$PAYLOAD" "$SYMBOLOUT" "$DEBPKG"
+
+# Publish core application executables
+echo "Publishing core application..."
+dotnet publish "$GCM_SRC" \
+	--configuration="$CONFIGURATION" \
+	--framework="$FRAMEWORK" \
+	--runtime="$RUNTIME" \
+    --self-contained=true \
+    "/p:PublishSingleFile=True" \
```
mjcheetham, line 87: I've thought about making GCM Core a .NET Core "single-file" on all platforms recently actually.
Perhaps we should look at setting this in the `shared/Git-Credential-Manager` project file directly?

One issue however with the single-file model is that of signing. As I understand it, the single-file format is basically a self extracting archive, that extracts to some location and then forwards execution to this extracted location.

The binaries inside the archive (ELF on Linux, Mach-O on macOS, and PE on Windows) will not be themselves signed. Is this an issue? I know Gatekeeper on Mac can be unhappy to run executables, and also there's Apple Notarization. For Windows I know theres some groups inside of MSFT that require all signed binaries on their build machines too.

For Linux this is less an issue of course, since there's no equivalent to Authenticode or Apple Signing for ELF and the disjointed Linux world, AFAIK.

So perhaps what I'm saying is I've just talked myself out of changing this for all projects 😆

.github/workflows/build-installers.yml
```diff
@@ -0,0 +1,38 @@
+name: Build-Installers
+
+on:
+  push:
+    branches: [ master ]
```
derrickstolee, line 5: We will likely want these on `release` branch, too? cc: @mjcheetham

src/linux/Payload.Linux/build.sh
```diff
@@ -44,8 +44,8 @@ ROOT="$( cd "$THISDIR"/../../.. ; pwd -P )"
 SRC="$ROOT/src"
 OUT="$ROOT/out"
 GCM_SRC="$SRC/shared/Git-Credential-Manager"
-PAYLOAD_SRC="$SRC/linux/Payload.Linux"
-PAYLOAD_OUT="$OUT/linux/Payload.Linux"
+# PAYLOAD_SRC="$SRC/linux/Payload.Linux"
```
derrickstolee, line 47: Perhaps remove the line instead of commenting it out?

src/linux/Payload.Linux/build.sh
```diff
@@ -32,17 +38,106 @@ case "$i" in
...
+tar -czvf "$TAROUT" * || exit 1
+popd
+
+# Build .deb
+INSTALL_TO="$DEBPKG/usr/bin/"
+mkdir -p "$DEBPKG/DEBIAN" "$INSTALL_TO"
+
+# make the debian control file
+cat >"$DEBPKG/DEBIAN/control" <<EOF
+Package: gcmcore
```
derrickstolee, line 125: Question: do we want the package to be `gcmcore` or `gcm-core` or `git-credential-manager-core`?

src/linux/Payload.Linux/build.sh
```diff
@@ -32,17 +38,106 @@ case "$i" in
...
+Version: $VERSION
+Section: vcs
+Priority: optional
+Architecture: $ARCH
+Depends:
+Maintainer: GCM-Core <gcmcore@microsoft.com>
```
derrickstolee, line 131: Is this a real email address?

src/linux/Payload.Linux/build.sh
```diff
@@ -32,17 +38,106 @@ case "$i" in
...
+# Copy single binary to target installation location
+cp "$PAYLOAD/git-credential-manager-core" "$INSTALL_TO"
 
-# Layout and pack
-"$PAYLOAD_SRC/layout.sh" --configuration="$CONFIGURATION" --output="$PAYLOAD" || exit 1
-"$PAYLOAD_SRC/pack.sh" --payload="$PAYLOAD" --output="$TAROUT" || exit 1
+dpkg-deb --build "$DEBPKG" "$DEBOUT"
 
-echo "Build of Payload.Linux complete."
+echo "Pack complete."
```
derrickstolee, line 143: Is it possible to publish the `.deb` as a build artifact? I'd love to download the `.deb` from a PR and install it locally for testing.

derrickstolee (commented):
Very happy with the progress. Mostly a few nits and questions. The biggest improvement I could see would be a way to push the `.deb` as a build artifact so it is easier to consume from CI/PR builds. That could be an important step for when we do create a publishing workflow.

src/linux/Payload.Linux/build.sh
kyle-rader: CloudBuild is moving to requiring all binaries be signed IIRC so I would expect that to eventually be an issue if using gcm core to auth in cloud build (windows). This is an excellent question I'll see if I find the answer to.

src/linux/Payload.Linux/build.sh
kyle-rader: Not yet. But I'm happy to create it and add us as owners.

src/linux/Payload.Linux/build.sh
kyle-rader: @mjcheetham I'm guessing we should follow the precedent for the windows and mac distro naming? Which I think is `git-credential-manager-core` ?

src/linux/Payload.Linux/build.sh
kyle-rader: Yep!

kyle-rader:
Both the `tar.gz` and `.deb` are now uploaded as `Installers.zip` for every workflow run. 
image

derrickstolee:
> Both the `tar.gz` and `.deb` are now uploaded as `Installers.zip` for every workflow run.

This is excellent. The build I downloaded is [available at this link](https://github.com/microsoft/Git-Credential-Manager-Core/suites/1178957211/artifacts/17237693). I found out that GitHub creds are required for this file, so my simple `curl` process from my Linux box was insufficient to download the file. I'll work on that, but it is not a requirement for this PR.

derrickstolee:
I downloaded it on my Windows machine and uploaded it to my Linux machine with `scp`. 

```sh
Not Foundstolee@stolee-linux-metal:/_scratch$ unzip gcmcore.zip 
Archive:  gcmcore.zip
  inflating: gcmcore-linux_amd64.release.2.0.227.40047.deb  
  inflating: gcmcore-linux_amd64.release.2.0.227.40047.tar.gz  

stolee@stolee-linux-metal:/_scratch$ sudo dpkg -i gcmcore-linux_amd64.release.2.0.227.40047.deb 
Selecting previously unselected package gcmcore.
(Reading database ... 207831 files and directories currently installed.)
Preparing to unpack gcmcore-linux_amd64.release.2.0.227.40047.deb ...
Unpacking gcmcore (2.0.227.40047) ...
Setting up gcmcore (2.0.227.40047) ...

stolee@stolee-linux-metal:/_scratch$ git-credential-manager-core version
Git Credential Manager version 2.0.216-beta+4c43d7b1a6 (Linux, .NET Core 3.1.7)
```

derrickstolee (approved):

src/linux/Payload.Linux/build.sh
mjcheetham: Are alias supported? I realise `git-credential-manager-core` is a mouthful, but it is the official name (and I'm a stickler for formalities 😛)..

src/linux/Payload.Linux/build.sh
mjcheetham: `gcmsupport@microsoft.com` _is_ a real email. We can use this. It forwards to the team.

.github/workflows/build-installers.yml
mjcheetham: Yes please 👍

mjcheetham (approved):
LGTM! Thanks @kyle-rader 

One question for everyone.. do we want to/can we have the Debian package run `git-credential-manager-core configure` after install or not? And if so, with or without `--system`?

For those not familiar with the `configure` command.. this will set the following configuration in either the user's (`--global` in Git parlance) `~/.gitconfig, or the `/etc/gitconfig` (`--system`) file:

```ini
[credential]
    helper =
    helper = manager-core
    useHttpPath = true
```

src/linux/Payload.Linux/build.sh
mjcheetham: Oh and `https://aka.ms/gcmcore` is also a short-link that we own (it redirects to the repository page on GitHub). We also own <https://github.com/GitCredentialManager>, but this organisation cannot host any code repositories but only contact or readme information, per Microsoft OSS policy.

PR #171, opened 2020-09-18
Remove username/password authentication for GitHub.com

mjcheetham:
Now that we have the GCM OAuth application in the allow-list for GitHub apps, we no longer need to offer username/password authentication for GitHub.com!

### DIFF
--- a/src/shared/GitHub.Tests/GitHubHostProviderTests.cs
+++ b/src/shared/GitHub.Tests/GitHubHostProviderTests.cs
@@ -175,7 +175,7 @@ namespace GitHub.Tests
         public async Task GitHubHostProvider_GetSupportedAuthenticationModes_OverrideInvalid_ReturnsDetectedValue()
         {
             var targetUri = new Uri("https://github.com");
-            var expectedModes = GitHubConstants.DotDomAuthenticationModes;
+            var expectedModes = GitHubConstants.DotComAuthenticationModes;
 
             var context = new TestCommandContext
             {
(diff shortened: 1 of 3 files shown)


derrickstolee (approved):

AlexanderLanin:
Sorry, I'm new here. Shouldn't this change also remove this?
https://github.com/microsoft/Git-Credential-Manager-Core/blob/c3a543a1352dba953e027f7a92ee4a94b2293a22/src/shared/GitHub/GitHubHostProvider.cs#L98

And therefore this?
https://github.com/microsoft/Git-Credential-Manager-Core/blob/c3a543a1352dba953e027f7a92ee4a94b2293a22/src/shared/GitHub/GitHubHostProvider.cs#L129


Maybe it can be reused for GitHub Enterprise with ADFS somehow. But so far I don't see how. That would be quite nice since for me GCM is not (yet) allowed via OAuth. And now password login is gone for creating PATs as well.

mjcheetham:
Hi @AlexanderLanin,

> But so far I don't see how. That would be quite nice since for me GCM is not (yet) allowed via OAuth. And now password login is gone for creating PATs as well.

You should be able to set the [`GCM_GITHUB_AUTHMODES` environment variable](https://github.com/microsoft/Git-Credential-Manager-Core/blob/master/docs/environment.md#gcm_github_authmodes) or the [Git config setting `credential.gitHubAuthModes
`](https://github.com/microsoft/Git-Credential-Manager-Core/blob/master/docs/configuration.md#credentialgithubauthmodes) to `"basic oauth"` to enable both options again. 

Also, since username/password support has been removed from the GitHub server, you'd need to manually generate a PAT and enter that as the password in the prompt.

Let me know if there are any issues with this workaround and create a new issue. Thanks!

PR #183, opened 2020-09-25
gitconfig: avoid deadlock on large amounts of config data

dscho:
This fixes the problem I encountered today after upgrading from GCM Core v2.0.194.40577 to v2.0.246.34937: due to the sheer size of my repository config (`git config -l` produces over 25kB of data), the credential manager would just hang, waiting for `git config` to be done, while the latter waited for the former to read its output.

### DIFF
--- a/src/shared/Microsoft.Git.CredentialManager/GitConfiguration.cs
+++ b/src/shared/Microsoft.Git.CredentialManager/GitConfiguration.cs
@@ -101,6 +101,9 @@ namespace Microsoft.Git.CredentialManager
             using (Process git = _git.CreateProcess($"config --null {level} --list"))
             {
                 git.Start();
+                // To avoid deadlocks, always read the output stream first and then wait
+                // TODO: don't read in all the data at once; stream it
+                string data = git.StandardOutput.ReadToEnd();
                 git.WaitForExit();
 
                 switch (git.ExitCode)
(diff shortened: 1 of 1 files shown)


src/shared/Microsoft.Git.CredentialManager/GitConfiguration.cs
```diff
@@ -101,6 +101,8 @@ public void Enumerate(GitConfigurationEnumerationCallback cb)
             using (Process git = _git.CreateProcess($"config --null {level} --list"))
             {
                 git.Start();
+                // To avoid deadlocks, always read the output stream first and then wait.
```
derrickstolee, line 104: Could you also keep the `TODO` from before? We still want to track work to stream the data instead of creating one giant `string`.

derrickstolee (approved):

src/shared/Microsoft.Git.CredentialManager/GitConfiguration.cs
dscho: Sure!

kyle-rader (approved):

src/shared/Microsoft.Git.CredentialManager/GitConfiguration.cs
dscho: Done!

mjcheetham (approved):
:shipit:

mjcheetham:
> `git config -l` produces over 25kB of data

P.s., yes this exposed a bug in GCM Core, but perhaps you should consider trimming your config.. just a bit? 🤣 
`git config --war-and-peace`

dscho:
> perhaps you should consider trimming your config.. just a bit? 🤣
> `git config --war-and-peace`

I _know_ I have a lot of branches and remotes... But it's not my fault that git/git has so many forks that I need to keep track of! 😁

```sh
$ git remote -v | wc
     58     174    3095

$ git config --get-regexp 'branch\..*\.remote' | wc
     83     166    3682
```

PR #197, opened 2020-10-28
Split Windows installer in to user/system flavours

mjcheetham:
Split the Windows installer into two versions: a single-user installer, and a system-wide installer.

The old installer installed in Program Files (x86) and set the system Git configuration, which is now what the system-flavoured installer does.

The new single-user installer places files in %LOCALAPPDATA% and updates the ~/.gitconfig (global) user Git configuration instead.

The benefit of the single-user installer is that it does not require an administrator to install it, and doesn't affect other users of the system.

There will now be two installers for Windows: `gcmcore-win....` and `gcmcoreuser-win....`. This is the same as VS Code.

### DIFF
--- a/src/windows/Installer.Windows/Installer.Windows.csproj
+++ b/src/windows/Installer.Windows/Installer.Windows.csproj
@@ -18,7 +18,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Tools.InnoSetup" Version="5.6.1" />
+    <PackageReference Include="Tools.InnoSetup" Version="6.0.5" />
     <PackageReference Include="MicroBuild.Core" Version="0.2.0">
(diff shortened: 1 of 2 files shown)


derrickstolee (approved):

cyqsimon:
README.md needs to be updated with an explanation for this. I was initially very confused as to why there are two Windows binaries and which one I should install.

mjcheetham:
> README.md needs to be updated with an explanation for this. I was initially very confused as to why there are two Windows binaries and which one I should install.

Hi @cyqsimon, thanks for reminding me here. I've updated the readme in https://github.com/microsoft/Git-Credential-Manager-Core/commit/b9084ddfbd6343e1969dcf1ecee584d882620fa0.

PR #199, opened 2020-10-28
Update the configure command to set the full path to GCM Core

mjcheetham:
Update the configure and unconfigure actions to now set the full file path to GCM Core, rather than rely on `manager-core` resolving to `git-credential-manager-core` on the PATH by Git.

This was a problem because Git for Windows now bundles GCM Core, and even if a standalone install of GCM Core was present on the system with a higher version, since the bundled copy is found on the PATH before anything else, Git was always picking the old one.

The change to using full paths helps fix this issue, and also another issue on macOS where sometimes the /usr/local/bin directory is not on the PATH (such as for the root user, or during a postinstall script for a flat-package [.pkg] file).


derrickstolee (approved):
I think this is the right approach. Just a few tests to get passing and then we can move forward.

PR #200, opened 2020-10-28
Just two tiny typos

Molkree:


derrickstolee (approved):
Thank you for noticing and fixing!

PR #201, opened 2020-10-28
Setup Microsoft ESRP Signing for .deb installer

derrickstolee:
This adds a new workflow build-installers.yml, where hopefully we can either migrate all installer builds in the future (or rename and split them across the platforms.)

This workflow is heavily dependent on a lot of internal Microsoft infrastructure. What we have so far results in a GPG signed .deb for GCM-Core. This is currently uploaded as an artifact on the workflow itself but it not yet published. The 3rd and last build job yet to be added is using setting up yet another internal tool with the right security to use our internal publishing service.

This does the signing, but does not actually publish the package to the Microsoft package feed.

Replaces #170.

### DIFF
--- a/.github/run_esrp_signing.py
+++ b/.github/run_esrp_signing.py
@@ -1,0 +1,112 @@
+import json
+import os
+import glob
+import pprint
+import subprocess
+import sys
+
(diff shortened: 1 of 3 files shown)


mjcheetham (approved):
This looks good for now, but there will most certainly be some cleanup required.

PR #203, opened 2020-10-29
Make the system installer always append GCM Core entries, not outright set/replace existing values (and fix a typo)

mjcheetham:
Rather than set the helper to be _only_ GCM Core in the system configuration when called with `(un)configure --system`, we do what we are already doing in the user-case.
We only append an empty ("") reset entry, and then the GCM full path entry. This means on uninstall of the standalone GCM (be that the system or user install), we restore the previous entry, always.

Catch more existing config cases such as when another 'empty' entry is _after_ GCM. We will now see this case and reconfigure ourselves to be exclusively 'in front'.

Also we are not that meta.. "Git Credential Manager ~Manager~ Core". Whoops.

### DIFF
--- a/src/shared/Microsoft.Git.CredentialManager/Application.cs
+++ b/src/shared/Microsoft.Git.CredentialManager/Application.cs
@@ -160,13 +160,15 @@ namespace Microsoft.Git.CredentialManager
             //     ...                # any number of helper entries (possibly none)
             //     helper =           # an empty value to reset/clear any previous entries (if applicable)
             //     helper = {appPath} # the expected executable value & directly following the empty value
-            //     ...                # any number of helper entries (possibly none)
+            //     ...                # any number of helper entries (possibly none, but not the empty value '')
             //
             string[] currentValues = config.GetAll(helperKey).ToArray();
 
-            // Try to locate an existing app entry with a blank reset/clear entry immediately preceding
+            // Try to locate an existing app entry with a blank reset/clear entry immediately preceding,
+            // and no other blank empty/clear entries following (which effectively disable us).
             int appIndex = Array.FindIndex(currentValues, x => Context.FileSystem.IsSamePath(x, appPath));
-            if (appIndex > 0 && string.IsNullOrWhiteSpace(currentValues[appIndex - 1]))
+            int lastEmptyIndex = Array.FindLastIndex(currentValues, string.IsNullOrWhiteSpace);
+            if (appIndex > 0 && string.IsNullOrWhiteSpace(currentValues[appIndex - 1]) && lastEmptyIndex < appIndex)
             {
                 Context.Trace.WriteLine("Credential helper configuration is already set correctly.");
             }
--- a/src/shared/Microsoft.Git.CredentialManager/Application.cs
+++ b/src/shared/Microsoft.Git.CredentialManager/Application.cs
@@ -162,7 +162,7 @@ namespace Microsoft.Git.CredentialManager
             //     helper = {appPath} # the expected executable value & directly following the empty value
             //     ...                # any number of helper entries (possibly none)
             //
-            string[] currentValues = config.GetRegex(helperKey, Constants.RegexPatterns.Any).ToArray();
+            string[] currentValues = config.GetAll(helperKey).ToArray();
 
             // Try to locate an existing app entry with a blank reset/clear entry immediately preceding
             int appIndex = Array.FindIndex(currentValues, x => Context.FileSystem.IsSamePath(x, appPath));
@@ -179,8 +179,8 @@ namespace Microsoft.Git.CredentialManager
 
                 // Add an empty value for `credential.helper`, which has the effect of clearing any helper value
                 // from any lower-level Git configuration, then add a second value which is the actual executable path.
-                config.ReplaceAll(helperKey, Constants.RegexPatterns.None, string.Empty);
-                config.ReplaceAll(helperKey, Constants.RegexPatterns.None, appPath);
+                config.Add(helperKey, string.Empty);
+                config.Add(helperKey, appPath);
             }
 
             return Task.CompletedTask;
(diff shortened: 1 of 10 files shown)


derrickstolee (approved):

dscho (approved):

dscho:
Maybe `s/type/typo/` in this PR's title? :-)

dscho (approved):

src/shared/Microsoft.Git.CredentialManager/Application.cs
```diff
@@ -147,63 +146,41 @@ Task IConfigurableComponent.ConfigureAsync(ConfigurationTarget target)
...
+            //     ...                # any number of helper entries (possibly none)
+            //
+            string[] currentValues = config.GetRegex(helperKey, Constants.RegexPatterns.Any).ToArray();
 
-                    break;
+            // Try to locate an existing app entry with a blank reset/clear entry immediately preceding
+            int appIndex = Array.FindIndex(currentValues, x => Context.FileSystem.IsSamePath(x, appPath));
+            if (appIndex > 0 && string.IsNullOrWhiteSpace(currentValues[appIndex - 1]))
+            {
+                Context.Trace.WriteLine("Credential helper configuration is already set correctly.");
```
dscho, line 171: Should we also look for empty entries _after_ this (which essentially disable GCM Core)?

src/shared/Microsoft.Git.CredentialManager/Application.cs
```diff
@@ -214,56 +191,41 @@ Task IConfigurableComponent.UnconfigureAsync(ConfigurationTarget target)
...
+
+            int appIndex = Array.FindIndex(currentValues, x => Context.FileSystem.IsSamePath(x, appPath));
+            if (appIndex > -1)
+            {
+                // Check for the presence of a blank entry immediately preceding an app entry in the last position
+                if (appIndex > 0 && appIndex == currentValues.Length - 1 &&
+                    string.IsNullOrWhiteSpace(currentValues[appIndex - 1]))
+                {
+                    // Clear the blank entry
+                    config.UnsetAll(helperKey, Constants.RegexPatterns.Empty);
```
dscho, line 224: What if there are _multiple_ empty entries?

src/shared/Microsoft.Git.CredentialManager/Application.cs
```diff
@@ -147,63 +146,41 @@ Task IConfigurableComponent.ConfigureAsync(ConfigurationTarget target)
...
+                Context.Trace.WriteLine("Updating Git credential helper configuration...");
 
-                default:
-                    throw new ArgumentOutOfRangeException(nameof(target), target, "Unknown configuration target.");
+                // Clear any existing app entries in the configuration
+                config.UnsetAll(helperKey, Regex.Escape(appPath));
+
+                // Add an empty value for `credential.helper`, which has the effect of clearing any helper value
+                // from any lower-level Git configuration, then add a second value which is the actual executable path.
+                config.ReplaceAll(helperKey, Constants.RegexPatterns.None, string.Empty);
```
dscho, line 182: Are we really sure this should be `ReplaceAll`? I thought maybe `Add` or `Append`...

mjcheetham: The manpages say:
```
--add
           Adds a new line to the option without altering any existing values. This is the same as providing
           ^$ as the value_regex in --replace-all.
```

src/shared/Microsoft.Git.CredentialManager/Application.cs
mjcheetham: Not entirely sure what we can do here to identify a specific line in a config file, that matches the same value, using only `git config` commands. Any ideas?

src/shared/Microsoft.Git.CredentialManager/Application.cs
mjcheetham: True.. will also need to check no other empty entries follow `appIndex`.

src/shared/Microsoft.Git.CredentialManager/Application.cs
mjcheetham: Will replace with calls directly to `-add` however for good measure.

src/shared/Microsoft.Git.CredentialManager/Application.cs
mjcheetham: Done. If there are empty entries after the GCM entry, we will attempt to add GCM back 'in front' (with another "" value to make sure GCM is the only helper) again.

src/shared/Microsoft.Git.CredentialManager/Application.cs
mjcheetham: Done. Replaced these usages of `--replace-all` with a regex matching nothing, with `--add`.
Also updated our use of `--get-regexp` with and "any" regex value to `--get-all`.

mjcheetham:
Sorry @dscho, requesting _yet another_ review..

dscho (approved):

src/shared/Microsoft.Git.CredentialManager/GitConfiguration.cs
```diff
@@ -267,18 +267,23 @@ public IEnumerable<string> GetAll(string name)
                 switch (git.ExitCode)
                 {
                     case 0: // OK
+                        string[] entries = data.Split('\0');
+
+                        // Because each line terminates with the \0 character, splitting leaves us with one
+                        // bogus blank entry at the end of the array which we should ignore
```
derrickstolee, line 273: I was thinking that what you really wanted was:

```cs
string[] entries = data.Split('\0', StringSplitOptions.RemoveEmptyEntries);
```

but you actually _do_ care about some empty values in the middle. This is an appropriate workaround.

derrickstolee (approved):

src/shared/Microsoft.Git.CredentialManager/GitConfiguration.cs
mjcheetham: Yep. Entry entries are important. Terminating with `\0` would be useful if writing my own parser over the data, but not when using the `string.Split` method, sadly.

PR #205, opened 2020-11-01
Fix missing space in username input

mastercoms:
This caused the preceding argument to fail due to lack of separation between args


mjcheetham (approved):
Thanks for finding and fixing this @mastercoms!

PR #208, opened 2020-11-02
Cut another release with fix to macOS publishing

mjcheetham:
See #207


mjcheetham:
Not sure what happened with the "build linux installers" workflow here..

```
With the provided path, there will be 3 file(s) uploaded
Total file count: 3 ---- Processed file #2 (66.6%)
Total size of all the files uploaded is 54809299 bytes
Error: read ECONNRESET
```

Looks like GitHub stopped us from uploading the artefacts. Retrying now..

derrickstolee (approved):

PR #212, opened 2020-11-06
Allow user to override which interactive authentication flow used for MS auth

### LINKED ISSUE #210: Provide options to override MS authentication flows
At the moment we force users to use a particular Microsoft authentication flow (via MSAL) based on the current environment.

- If there's an interactive session..
   - ..on Windows we use the embedded browser flow.
   - ..on non-Windows..
      - ..with a native UI helper we use that helper.
      - ..without a native UI helper we use the system browser.
- If there's no interactive session we use device code flow.

The user should be free to select the flow they prefer, rather than forcing them to use the one we deem "best". The system browser or device-code flows can be used enable scenarios like FIDO or Windows Hello, which may be desirable.

Visual Studio already offers an option to change the authentication flow.

image

This can be implemented pretty easily by introduction of a `GCM_MSAUTH_FLOW`/`credential.msAuthFlow` or similar setting that is respected by the `MicrosoftAuthentication` component.

**Note:** Use of the system browser requires a localhost redirect be set in the AAD configuration. We would need to ensure this was set correctly.

mjcheetham:
Allow the user to override which Microsoft authentication flow to use when interaction is required.

The options available are:
 - embedded web view
 - system web view
 - device code

Previously we always picked one of these flows (or used a native external helper) to use. In some instances it may be desirable for the user to force a particular flow, for instance to take advantage of browser sign-in state or OS integrations by using 'system'.

If no option is set, or `auto` is selected the existing behaviour is preserved (we pick for the user).

Fixes #210

### DIFF
--- a/src/shared/Microsoft.Git.CredentialManager/Authentication/AuthenticationBase.cs
+++ b/src/shared/Microsoft.Git.CredentialManager/Authentication/AuthenticationBase.cs
@@ -106,6 +106,13 @@ namespace Microsoft.Git.CredentialManager.Authentication
                 helperName = PlatformUtils.IsWindows() ? $"{defaultValue}.exe" : defaultValue;
             }
 
+            // If the user set the helper override to the empty string then they are signalling not to use a helper
+            if (string.IsNullOrEmpty(helperName))
+            {
+                path = null;
+                return false;
+            }
+
             if (Path.IsPathRooted(helperName))
             {
                 path = helperName;
--- a/src/shared/Microsoft.Git.CredentialManager/Authentication/MicrosoftAuthentication.cs
+++ b/src/shared/Microsoft.Git.CredentialManager/Authentication/MicrosoftAuthentication.cs
@@ -4,6 +4,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Net.Http;
+using System.Runtime.InteropServices;
 using System.Threading.Tasks;
 using Microsoft.Identity.Client;
 using Microsoft.Identity.Client.Extensions.Msal;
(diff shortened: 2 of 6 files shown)


src/shared/Microsoft.Git.CredentialManager/Authentication/AuthenticationBase.cs
```diff
@@ -116,6 +123,7 @@ protected bool TryFindHelperExecutablePath(string envar, string configName, stri
                 path = Path.Combine(executableDirectory!, helperName);
             }
 
+
```
derrickstolee, line 126: nit: extraneous whitespace

src/shared/Microsoft.Git.CredentialManager/Authentication/MicrosoftAuthentication.cs
```diff
@@ -111,41 +121,86 @@ private async Task<JsonWebToken> GetAccessTokenInProcAsync(string authority, str
...
-                        .ExecuteAsync();
+                    case MicrosoftAuthenticationFlowType.Auto:
+                        if (CanUseEmbeddedWebView())
+                            goto case MicrosoftAuthenticationFlowType.EmbeddedWebView;
+
+                        if (CanUseSystemWebView(app, redirectUri))
+                            goto case MicrosoftAuthenticationFlowType.SystemWebView;
+
+                        // Fall back to device code flow
+                        goto case MicrosoftAuthenticationFlowType.DeviceCode;
```
derrickstolee, line 141: The existence of `goto`s makes me want to restructure this code. However, this `switch` is so carefully constructed, that the alternatives actually look worse. Kudos on finding a good use for `goto`.

derrickstolee (approved):

src/shared/Microsoft.Git.CredentialManager/Authentication/MicrosoftAuthentication.cs
mjcheetham: Yeah, this sort of "have several options, and an 'auto' choice that may pick one of those options" pattern is best served with a `switch` with `goto case`, IMO.
It's the only time I've found `goto` is acceptable; plus `goto case` feels safer that a random label.

This same pattern is already present in the `GitHubAuthentication` class for selecting the type of auth flow (user/pass or OAuth) or letting it auto-detect.

PR #221, opened 2020-11-17
Create apt-get release pipeline

kyle-rader:
Creating a new workflow to be triggered off full release to download the signed .deb asset and deploy it.

The workflow does not currently check if the .deb is actually signed, this is something the repo api _will do in the future_ similar to how it currently does for rpm packages.

Builds now failing because of https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/ `set-env` usage in the GitVersioning tools :(

### DIFF
--- a/.github/fetch_release.py
+++ b/.github/fetch_release.py
@@ -1,0 +1,38 @@
+from os import environ as env
+import requests
+import time
+
+if 'RELEASE' not in env or not env['RELEASE']:
+    print("Require env var RELEASE is missing!")
+    exit(1)
+
+release = env['RELEASE']
+
+release_arg = "latest" if release == "latest" else f"tags/{release}"
+release_url = f"https://api.github.com/repos/microsoft/git-credential-manager-core/releases/{release_arg}"
+
+try:
+    r = requests.get(release_url)
+
+    # if at first you don't succeed...
+    if r.status_code != 200:
+        time.sleep(2)
+        r = requests.get(release_url)
+
+    if r.status_code != 200:
+        raise Error(f"Failed to fetch release info from {release_url}")
+
+    asset = [a for a in r.json()['assets'] if a['name'].endswith(".deb")][0]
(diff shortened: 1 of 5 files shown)


.github/workflows/build-signed-deb.yml
```diff
@@ -50,7 +50,7 @@ jobs:
       with:
         name: LinuxInstallers
 
-    - uses: Azure/login@v1.1
+    - uses: azure/login@v1
```
kyle-rader, line 53: This is for https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

mjcheetham:
I've created a PR into the main branch that updates our Nerdbank.GitVersioning to one that uses environment files in GitHub Actions: https://github.com/microsoft/Git-Credential-Manager-Core/pull/224

Once this is merged, we can unblock this branch (might need a rebase).

.github/fetch_release.py
```diff
@@ -0,0 +1,49 @@
...
+if 'RELEASE' in env and env['RELEASE']:
+    release = env['RELEASE'].strip()
+else:
+    release = "latest"
+
+release_arg = "latest" if release == "latest" else f"tags/{release}"
+
+print(f"release_arg set to {release_arg}")
+
+release_url = f"https://api.github.com/repos/microsoft/git-credential-manager-core/releases/{release_arg}"
```
mjcheetham, line 15: I know access to the GitHub API can be throttled when used unauthenticated. Using the `GITHUB_TOKEN` might be useful, but not required for this PR.

mjcheetham (approved):
Looks good to me! Just a comment about using the API without auth can lead to throttling by IP address - not something we need to address immediately.

derrickstolee (approved):

kyle-rader:
Rebased.

Issue #219, opened 2020-11-14
Small addition to GitHub enterprise detection


AlexanderLanin:
Currently GitHub ist auto detected by domain github.com.
While you could work forever on a clever heuristic, it does seem safe to make a minimal adjustment and detect github.* as GitHub.
E.g. github.myorganization.com

As far as I can tell this is https://github.com/microsoft/Git-Credential-Manager-Core/blob/c3a543a1352dba953e027f7a92ee4a94b2293a22/src/shared/GitHub/GitHubHostProvider.cs#L63 and https://github.com/microsoft/Git-Credential-Manager-Core/blob/c3a543a1352dba953e027f7a92ee4a94b2293a22/src/shared/GitHub.Tests/GitHubHostProviderTests.cs#L14

AlexanderLanin:
This is particularly annoying not just because of lack of OAuth in the generic login. But because the generic login is asking for a password which is disabled on GitHub.
Of course this is trivially fixable by config, but it affects too many users in my case. Someone always has trouble setting up a correct config.

PR #227, opened 2020-11-20
Refactor in order to simplify adding more tests

AlexanderLanin:
Besides looking better in any case ;-)
This is to support adding additional test cases for #219 

Review should be easier commit by commit. There are no changes besides moving from Facts to Theories.

Note: this is my first ever written code in C#, so feel free to be as strict as you wish.

### DIFF
--- a/src/shared/GitHub.Tests/GitHubHostProviderTests.cs
+++ b/src/shared/GitHub.Tests/GitHubHostProviderTests.cs
@@ -52,35 +52,26 @@ namespace GitHub.Tests
         }
 
 
-        [Fact]
-        public void GitHubHostProvider_GetCredentialServiceUrl_GitHubHost_ReturnsCorrectKey()
+        [Theory]
+        [InlineData("https://github.com", "https://github.com")]
+        [InlineData("https://gist.github.com", "https://github.com")]
+        public void GitHubHostProvider_GetCredentialServiceUrl(string uriString, string expectedService)
         {
-            const string expectedService = "https://github.com";
+            Uri uri = new Uri(uriString);
+
             var input = new InputArguments(new Dictionary<string, string>
             {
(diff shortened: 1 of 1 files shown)


src/shared/GitHub.Tests/GitHubHostProviderTests.cs
```diff
@@ -97,73 +97,28 @@ public async Task GitHubHostProvider_GetSupportedAuthenticationModes(string uriS
         }
 
 
-
-        [Fact]
-        public async Task GitHubHostProvider_GetSupportedAuthenticationModes_NotDotCom_OldInstanceNoPassword_ReturnsNone()
+        [Theory]
+        [InlineData("https://example.com", null, "0.1", false, AuthenticationModes.None)]
+        [InlineData("https://example.com", null, "0.1", true, AuthenticationModes.Basic)]
+        [InlineData("https://example.com", null, "100.0", false, GitHubConstants.DotComAuthenticationModes)]
```
mjcheetham, line 103: Can you replace the DotCom constant here with the explicit OAuth enumeration value please?
This is a test for the GitHub Enterprise mode discovery/logic and we don't want to make a link here to the DotCom modes constant.

```suggestion
        [InlineData("https://example.com", null, "100.0", false, AuthenticationModes.OAuth)]
```

mjcheetham (changes requested):
Apart from the one comment, this looks good thanks! 👍

src/shared/GitHub.Tests/GitHubHostProviderTests.cs
AlexanderLanin: Sure. That's what was there before. This was intentionally without changing any logic.

mjcheetham (approved):
Thanks! Looks great to me.

Issue #222, opened 2020-11-17
GitHub should ask for "PAT" and not "Password"


AlexanderLanin:
Since OAuth is sometimes disabled on GitHub Enterprise the old form still needs to be maintained. 
It's currently asking for username/password which is basically just wrong. Username/password login is disabled on GitHub.
It should say username/PAT (Potentially with a hint how to generate a PAT and a link to `{domain}/settings/tokens/new`).

See https://github.com/microsoft/Git-Credential-Manager-Core/pull/171#issuecomment-728974236 for how the old dialog can currently be reached.

AlexanderLanin:
Upon further investigation PAT doesn't work at all at the moment, does it?
The code [always](https://github.com/microsoft/Git-Credential-Manager-Core/blob/ba1fe717ce49f0eaa53e6cc2f7e40988119a9622/src/shared/GitHub/GitHubHostProvider.cs#L99) tries to create a PAT which fails considering the user has already entered a PAT.

Am I somehow wrong?

mjcheetham:
> The code [always](https://github.com/microsoft/Git-Credential-Manager-Core/blob/ba1fe717ce49f0eaa53e6cc2f7e40988119a9622/src/shared/GitHub/GitHubHostProvider.cs#L99) tries to create a PAT which fails considering the user has already entered a PAT.

It shouldn't fail here. The call to `GeneratePersonalAccessTokenAsync` with a PAT as the password, will pass the PAT to the `authorizations` API endpoint, which _should_ return a 403 response with the error "This API can only be accessed with username and password Basic Auth".

We [handle this obscure case](https://github.com/microsoft/Git-Credential-Manager-Core/blob/ba1fe717ce49f0eaa53e6cc2f7e40988119a9622/src/shared/GitHub/GitHubRestApi.cs#L137-L153) and return the 'password' (the PAT) as-is, knowing it is a PAT directly. At least, we should handle this case. It's possible the GitHub server no longer returns this case since username/password access was removed from the API.

mjcheetham:
The ability to enter a PAT explicitly for GHES (and GitHub.com) will be available on the next feature release of GCM Core.
PR #236 has been merged in.

PR #231, opened 2020-11-22
GitHub: add PAT support

AlexanderLanin:
As shown in #222 current GCM Core does not only not properly show PAT. It even doesn't work with PAT at all. This PR is supposed to fix both issues.
Minor points of #222 are not yet addressed like showing "What is a PAT?" and "Click here to generate a PAT".

### DIFF
--- a/src/shared/GitHub.Tests/GitHubHostProviderTests.cs
+++ b/src/shared/GitHub.Tests/GitHubHostProviderTests.cs
@@ -435,7 +435,7 @@ namespace GitHub.Tests
 
             var ghAuthMock = new Mock<IGitHubAuthentication>(MockBehavior.Strict);
             ghAuthMock.Setup(x => x.GetAuthenticationAsync(expectedTargetUri, null, It.IsAny<AuthenticationModes>()))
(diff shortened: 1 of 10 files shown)


mjcheetham:
Hi @AlexanderLanin, thanks for the PR to add explicit PAT support to the GitHub provider and UI!

We have been working on a branch with such a feature, but have taken a slightly different approach with plumbing the new mode through the different layers, and how this is presented in the UI. We'll be ready to share this soon.

AlexanderLanin:
So there is no need to continue on the PR? I was about to verify it works with GHES and PATs.
The only functional change I performed was https://github.com/microsoft/Git-Credential-Manager-Core/pull/231/files#diff-03dbffb4273f14f126ebe54ed4c6e0c1852aade112b934d560793e3cfa9ba845R78

mjcheetham:
Hi @AlexanderLanin,

Thanks again for the contribution. As you may have seen, with PR #236 the ability to add a PAT explicitly will be available in the next release. As a result I will be closing this PR. If you have any other issues or improvement suggestions please do contribute by issue or PR. Sorry that this particular PR didn't make it in given the existing work we had lined up at the same time. We do usually accept contributions where appropriate 😃 ❤️

Issue #228, opened 2020-11-20
Should OAuth client id/secret be public?


AlexanderLanin:
To my understanding of OAuth both client id and client secret **must** be secrets. However they are publicly accessible as part of the code base.

See [here](https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets) for GitHub solves secrets.

"To create secrets for a user account repository, you must be the repository owner. To create secrets for an organization repository, you must have admin access."
So I cannot try that.

mjcheetham:
Hi @AlexanderLanin,

I direct and quote you to the [GitHub Bounty page on this issue](https://bounty.github.com/ineligible.html#oauth_client_id_and_secrets_are_publicly_available_in_desktop_and_modile_apps):
> **OAuth client ID and secrets are publicly available in desktop and modile apps**
>
> It is expected that the GitHub-owned clients, such as GitHub for mobile and GitHub CLI, include both the OAuth client ID and OAuth secret. GitHub for mobile uses Universal/Deep links (github://) which helps reduce the risk of any issue presented here by binding the OAuth callback directly to the GitHub mobile application. We intentionally do not use PKCE in client apps and have accepted that risk.

For Public Client applications the client ID and client secret must be present on the user's device to perform the various OAuth2 flows. For Confidential Client applications, yes, the client secret must be secret. GCM Core's GitHub provider is a Public Client, and can only perform actions in the context of an authenticated user. "It can only do what the authenticated user can do themselves."

mjcheetham:
A fun aside, if you use a network inspection tool like Fiddler and watch something like Google Chrome, you'll also see the client ID and client secret for the Chrome application (and it's also in their source code too).

AlexanderLanin:
Thanks @mjcheetham, without your additional explanation the quote itself would not have been enough to understand the reason why it's ok.

PR #233, opened 2020-11-25
Add suppressions for credscan false positives

mjcheetham:
Add the required source suppressions for false positive identification of credentials, raised by the internal CredScan tool.

Most of the identified matches are fake credentials for unit tests. One match was the GitHub OAuth2 application client secret, which per issue #228 valid and an accepted & required 'secret' to be public.

Note to reviewers: yes, sadly the tool does indeed require the `[SuppressMessage...]` C# attribute be _in a comment_ to work... 🤦‍♂️


dscho (approved):
Looks good to me.

jeffhostetler (approved):

PR #234, opened 2020-11-25
Match Git config in a mixed-sensitively way

### LINKED ISSUE #223: Make gitHubAuthModes domain specific
gitHubAuthModes is currently a global setting and not specific to single domains. While this usually doesn't matter much, the use case is having an enterprise server which doesn't support OAuth. So I'm setting this to basic. However this will also disable OAuth on github.com. No way to be more specific.
I cannot just set to "basic oauth" as oauth fails in a very non obvious way with a 404 error message.

```
[credential "github.mycompany.com"]
	provider = GitHub
[credential]
	gitHubAuthModes = basic
```

This is a separate feature, but for the full set see #222 and #219.

mjcheetham:
Git configuration entry keys are neither case sensitive, nor in-sensitive. They are mixed sensitive.

The section and property names are NOT sensitive, but the 'scope' or inner component of the full name IS sensitive.

For example, `credential.example.com.provider` is the same as `CREDENTIAL.example.com.PROVIDER`, but `credential.EXAMPLE.COM.provider` is NOT.

We introduce a new StringComparer type that performs insensitive comparison on the section and property components, and sensitive comparison on the scope component.

Fixes #223


src/shared/Microsoft.Git.CredentialManager/GitConfigurationKeyComparer.cs
```diff
@@ -0,0 +1,80 @@
...
+        {
+            section = null;
+            scope = null;
+            property = null;
+
+            if (string.IsNullOrWhiteSpace(str))
+            {
+                return;
+            }
+
```
jeffhostetler, line 73: Should we assert/throw/return If `str` doesn't have any dots in it?
Just wondering if the scopeLength computation goes negative.

jeffhostetler (approved):

src/shared/Microsoft.Git.CredentialManager/GitConfigurationKeyComparer.cs
mjcheetham: If the input string to either of `GetHashCode` or `Equals` (the only public methods) contains no dots, then we just behave like `StringComparer.Ordinal`. I'll add some tests to confirm this behaviour.

toot-gerly (commented):
_removed_

PR #236, opened 2020-11-25
Add explicit PAT authentication mode for the GitHub Provider

### LINKED ISSUE #222: GitHub should ask for "PAT" and not "Password"

mjcheetham:
Add an authentication mode to the GitHub provider whereby a Personal Access Token can be returned directly, as generated by a user manually. This mode is always supported on both GitHub.com and GHES, as it should always be possible to create a PAT manually for whatever access is required.

This functionality sort-of already exists, but requires a few things:
1. The users needs to enable basic auth for GHES, or force override the authentication modes to allow basic auth (`GCM_GITHUB_AUTHMODES=basic`)
2. Know to enter their username, and the PAT as the password in a basic auth prompt

We update the WPF-based Windows GUI for GitHub authentication prompts to support the third authentication mode "PAT". There are also make some changes to how the dialog window is sized to scale with the size of the content (mins and maxes included), which may now vary in the extreme from all three auth options, to just one.

image
_GitHub.com with OAuth and PAT options_

image
_GHES with basic and PAT options_

image
_GHES with the PAT option only_

image
_A scenario where all options are presented. Currently not possible due to GHES not supporting GCM Core via OAuth, and GitHub.com no longer supporting basic auth - showing to prove the case should this happen such as by user override, or in future development._

Fixes #222

### DIFF
--- a/src/shared/GitHub/GitHubAuthentication.cs
+++ b/src/shared/GitHub/GitHubAuthentication.cs
@@ -28,15 +28,15 @@ namespace GitHub
             AuthenticationMode = mode;
         }
 
-        public AuthenticationPromptResult(ICredential basicCredential)
-            : this(AuthenticationModes.Basic)
+        public AuthenticationPromptResult(AuthenticationModes mode, ICredential credential)
+            : this(mode)
         {
-            BasicCredential = basicCredential;
+            Credential = credential;
         }
 
         public AuthenticationModes AuthenticationMode { get; }
 
-        public ICredential BasicCredential { get; set; }
+        public ICredential Credential { get; set; }
     }
 
     [Flags]
@@ -45,6 +45,9 @@ namespace GitHub
         None  = 0,
         Basic = 1,
         OAuth = 1 << 1,
+        Pat   = 1 << 2,
+
+        All   = Basic | OAuth | Pat
     }
 
     public class GitHubAuthentication : AuthenticationBase, IGitHubAuthentication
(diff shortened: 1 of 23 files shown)


src/shared/GitHub/GitHubAuthentication.cs
```diff
@@ -63,16 +66,24 @@ public async Task<AuthenticationPromptResult> GetAuthenticationAsync(Uri targetU
...
+                {
+                    promptArgs.Append(" --all");
+                }
+                else
+                {
+                    if ((modes & AuthenticationModes.Basic) != 0) promptArgs.Append(" --basic");
+                    if ((modes & AuthenticationModes.OAuth) != 0) promptArgs.Append(" --oauth");
+                    if ((modes & AuthenticationModes.Pat)   != 0) promptArgs.Append(" --pat");
+                }
+                if (!GitHubHostProvider.IsGitHubDotCom(targetUri)) promptArgs.AppendFormat(" --enterprise-url \"{0}\"", targetUri);
```
jeffhostetler, line 85: Do the targetUri and the userName fields need injection armoring??
I can't tell from the context here where the resulting string is going to go, but just wondering if it would be possible to attack things with a carefully crafted string...

jeffhostetler: For example, if the userName field contained a double quote and then some special characters.

src/shared/Microsoft.Git.CredentialManager.Tests/EnsureArgumentTests.cs
```diff
@@ -0,0 +1,33 @@
...
+        }
+
+        [Theory]
+        [InlineData(-3, 0, 10, true, true)]
+        [InlineData(13, 0, 10, true, true)]
+        [InlineData(10, 0, 10, true, false)]
+        [InlineData(0, 0, 10, false, true)]
+        [InlineData(-10, -10, 0, false, true)]
+        [InlineData(0, -10, 0, true, false)]
+        public void EnsureArgument_InRange_ThrowsException(int x, int lower, int upper, bool lowerInc, bool upperInc)
```
jeffhostetler, line 26: Nit: should the name of this test be EnsureArgument_OutOfRange_ThrowsException() ??

jeffhostetler (approved):

src/shared/Microsoft.Git.CredentialManager.Tests/EnsureArgumentTests.cs
mjcheetham: The naming pattern for the unit tests is `Type_Method_Input_Result`, where:

```
type   = EnsureArgument
method = InRange
input  = (null, it's all the different theory data)
result = throws an exception
```

src/shared/GitHub/GitHubAuthentication.cs
mjcheetham: So there _is_ an opportunity here to inject things to the args, for example if the username was literally:

```
alice" --pat "
```

..then you'd be starting the helper with the extra `--pat` argument as the expanded argument list would be:

```
.. --username "alice" --pat "" ..
```

However, there's no opportunity to execute another executable. This is because:

1. the executable being launched is discovered directly from the GCM Core installation directory,
2. the full executable path is specified as a completely separate argument to the .NET `ProcessStartInfo` class from the rest of the executable arguments,
3. .NET does the right thing, and these args are being passed eventually to CreateProcess (and not via a shell which would be vulnerable here to executing things it shouldn't).

So yes, it's possible inject something to the helper args, which isn't great as a creative username might mess up the prompt, but it doesn't present a threat.
The helpers are literally just there to show fancy input text boxes and buttons - they don't perform any logic themselves.

I'll look at adding better quoting and escaping of `"` however.

mjcheetham: Added escaping.

jeffhostetler:
Your new quoting scheme looks good.  Thanks.

PR #241, opened 2020-11-30
Detect github.my-company-server.com as GitHub

### LINKED ISSUE #219: Small addition to GitHub enterprise detection

AlexanderLanin:
Detect github[.subdomain].domain.tld as GitHub in order to better support GitHub Enterprise Server.
Prior to this patch a manual configuration of https://github.com/microsoft/Git-Credential-Manager-Core/blob/master/docs/configuration.md#credentialprovider is required.
Closes #219 

Unfortunately it's unclear whether this will catch any significant portion of such setups as the urls are not known. But it does seem reasonable and it does catch my use case :-)

### DIFF
--- a/src/shared/GitHub/GitHubHostProvider.cs
+++ b/src/shared/GitHub/GitHubHostProvider.cs
@@ -73,12 +73,19 @@ namespace GitHub
             string[] domains = hostName.Split(new char[] { '.' });
 
             // github[.subdomain].domain.tld
-            if (domains.Length >= 3 && domains[0] == "github")
+            if (domains.Length >= 3 &&
+                StringComparer.OrdinalIgnoreCase.Equals(domains[0], "github"))
+            {
                 return true;
+            }
 
             // gist.github[.subdomain].domain.tld
-            if (domains.Length >= 4 && domains[0] == "gist" && domains[1] == "github")
+            if (domains.Length >= 4 &&
+                StringComparer.OrdinalIgnoreCase.Equals(domains[0], "gist") &&
+                StringComparer.OrdinalIgnoreCase.Equals(domains[1], "github"))
+            {
                 return true;
+            }
 
             return false;
         }
@@ -286,7 +293,7 @@ namespace GitHub
             // Credentials for these repositories are the same as the one stored with "github.com".
             // Same for gist.github[.subdomain].domain.tld. The general form was already checked via IsSupported.
             int firstDot = uri.DnsSafeHost.IndexOf(".");
-            if (uri.DnsSafeHost.Substring(0, firstDot).Equals("gist", StringComparison.OrdinalIgnoreCase)) {
+            if (firstDot > -1 && uri.DnsSafeHost.Substring(0, firstDot).Equals("gist", StringComparison.OrdinalIgnoreCase)) {
                 return new Uri("https://" + uri.DnsSafeHost.Substring(firstDot+1));
             }
 
--- a/src/shared/GitHub/GitHubHostProvider.cs
+++ b/src/shared/GitHub/GitHubHostProvider.cs
@@ -52,16 +52,35 @@ namespace GitHub
                 return false;
             }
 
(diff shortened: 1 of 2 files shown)


src/shared/GitHub/GitHubHostProvider.cs
```diff
@@ -52,16 +52,35 @@ public override bool IsSupported(InputArguments input)
...
+            if (StringComparer.OrdinalIgnoreCase.Equals(hostName, GitHubConstants.GitHubBaseUrlHost) ||
+                StringComparer.OrdinalIgnoreCase.Equals(hostName, GitHubConstants.GistBaseUrlHost))
+            {
+                return true;
+            }
+
+            string[] domains = hostName.Split(new char[] { '.' });
+
+            // github[.subdomain].domain.tld
+            if (domains.Length >= 3 && domains[0] == "github")
```
mjcheetham, line 76:
```suggestion
            if (domains.Length >= 3 && StringComparer.OrdinalIgnoreCase.Equals(domains[0], "github"))
```

src/shared/GitHub/GitHubHostProvider.cs
```diff
@@ -52,16 +52,35 @@ public override bool IsSupported(InputArguments input)
...
+                return true;
+
+            // gist.github[.subdomain].domain.tld
+            if (domains.Length >= 4 && domains[0] == "gist" && domains[1] == "github")
```
mjcheetham, line 80:
```suggestion
            if (domains.Length >= 4 &&
                StringComparer.OrdinalIgnoreCase.Equals(domains[0], "gist") &&
                StringComparer.OrdinalIgnoreCase.Equals(domains[1], "github"))
```

src/shared/GitHub/GitHubHostProvider.cs
```diff
@@ -264,10 +283,11 @@ private static Uri NormalizeUri(Uri uri)
...
 
             // Special case for gist.github.com which are git backed repositories under the hood.
-            // Credentials for these repositories are the same as the one stored with "github.com"
-            if (uri.DnsSafeHost.Equals(GitHubConstants.GistBaseUrlHost, StringComparison.OrdinalIgnoreCase))
-            {
-                return new Uri("https://" + GitHubConstants.GitHubBaseUrlHost);
+            // Credentials for these repositories are the same as the one stored with "github.com".
+            // Same for gist.github[.subdomain].domain.tld. The general form was already checked via IsSupported.
+            int firstDot = uri.DnsSafeHost.IndexOf(".");
+            if (uri.DnsSafeHost.Substring(0, firstDot).Equals("gist", StringComparison.OrdinalIgnoreCase)) {
```
mjcheetham, line 289:
```suggestion
            if (firstDot > -1 && uri.DnsSafeHost.Substring(0, firstDot).Equals("gist", StringComparison.OrdinalIgnoreCase)) {
```

mjcheetham (changes requested):
Sorry for the delay here @AlexanderLanin. This seems like a reasonable addition which shouldn't trip up many people, if any (those who have a "github.example.tld" remote that isn't GHES - and then they can always force the `generic` provider).

Just a few changes regarding string comparison and one extra check in substring string handling (just to be extra safe).

mjcheetham (approved):
Looks great, thanks!

PR #243, opened 2020-12-01
Remove use of box drawing characters in warning output

mjcheetham:
Remove the use of box drawing characters in the warning output messages for TLS verification being disabled. Some systems (looking at you Windows) struggle with such characters. Replace them with simpler characters.

Whilst we're at it, also remove box drawing characters from code comments. Even though these aren't a problem, let's just do it for consistency.


derrickstolee (approved):

derrickstolee:
I haven't played with these characters since the Commodore 64 😝

jeffhostetler:
or Turbo Pascal...

jeffhostetler (approved):

PR #257, opened 2021-01-04
Ensure the /usr/local/bin directory exists on install

### LINKED ISSUE #256: Install failed when /usr/local/bin directory does not exist in macOS
'ln' command is failed when "/usr/local/bin" directory does not exist in macOS (Apple Silicon):
```
package_script_service[1860]: ./postinstall: ln: /usr/local/bin/git-credential-manager-core: No such file or directory
```

The script was succeed after I created the directory.

mjcheetham:
Ensure the /usr/local/bin directory exists on macOS installations before we create the symlink for GCM Core. The directory does not exist on freshly installed Macs.

Fixes #256

### DIFF
--- a/nuget.config
+++ b/nuget.config
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <clear />
+    <add key="nuget" value="https://api.nuget.org/v3/index.json" />
+  </packageSources>
+</configuration>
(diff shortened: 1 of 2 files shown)


jeffhostetler (approved):

dscho (approved):
Looks obviously correct to me.

My only (slight) concern is that Homebrew might expect a certain set of permissions of that directory, but it probably adjusts them when being installed, anyway.

mjcheetham:
> My only (slight) concern is that Homebrew might expect a certain set of permissions of that directory, but it probably adjusts them when being installed, anyway.

This was my thought too.. Also, I've just added a nuget.config file to explicitly set nuget.org as the package source to get around these build issues.

PR #272, opened 2021-02-02
Fix bug to ensure correct provider disposal

mjcheetham:
We are not correctly disposing of the host providers registered with the host provider registry. Fix this by actually calling `Dispose()` on the correct objects(!)

### DIFF
--- a/src/shared/Microsoft.Git.CredentialManager/HostProviderRegistry.cs
+++ b/src/shared/Microsoft.Git.CredentialManager/HostProviderRegistry.cs
@@ -194,7 +194,7 @@ namespace Microsoft.Git.CredentialManager
             _httpClient?.Dispose();
 
             // Dispose of all registered providers to give them a chance to clean up and release any resources
-            foreach (IHostProvider provider in _hostProviders.Values)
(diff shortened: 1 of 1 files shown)


derrickstolee (commented):
I see... foreach loops don't work correctly on `ValueCollection`s, so you need to convert it to an enumerator?

[The documentation example](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.values?view=net-5.0) says this loop should work.

mjcheetham:
@derrickstolee ..
> I see... foreach loops don't work correctly on `ValueCollection`s, so you need to convert it to an enumerator?
> 
> [The documentation example](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.values?view=net-5.0) says this loop should work.

The problem is that `_hostProviders` is of type `IDictionary<HostProviderPriority, ICollection<IHostProvider>>`, which means `_hostProviders.Values` is actually of type `ICollection<ICollection<IHostProvider>>`.

I need to unpack those outer collections.

derrickstolee (approved):
...of course. `.SelectMany()` is different than `.Select()`. I haven't had to think about LINQ for a long time.

PR #273, opened 2021-02-02
Use System.CommandLine to parse cmdline args and allow provider to offer cmds

mjcheetham:
Replace our basic custom command-line handling code with the System.CommandLine library which provides simpler and easier handling.

Add the ability for host providers to register themselves as offering custom commands, under the provider ID name.

For example a provider with the ID 'foo' would be able to expose commands under the `git-credential-manager-core foo <..>` command.


src/shared/Microsoft.Git.CredentialManager/Application.cs
```diff
@@ -47,62 +50,60 @@ public void RegisterProvider(IHostProvider provider, HostProviderPriority priori
...
+            var eraseCommand = new EraseCommand(Context, _providerRegistry);
+            var configureCommand = new ConfigureCommand(Context, _configurationService);
+            var unconfigureCommand = new UnconfigureCommand(Context, _configurationService);
+
+            // Add standard commands
+            rootCommand.AddCommand(getCommand);
+            rootCommand.AddCommand(storeCommand);
+            rootCommand.AddCommand(eraseCommand);
+            rootCommand.AddCommand(configureCommand);
+            rootCommand.AddCommand(unconfigureCommand);
```
derrickstolee, line 79: nit: could this be of the pattern

```cs
rootCommand.AddCommand(new GetCommand(Context, _providerRegistry);
```

instead of creating locals that have no other purpose?

src/shared/Microsoft.Git.CredentialManager/Commands/ProviderCommand.cs
```diff
@@ -0,0 +1,24 @@
...
+    public interface ICommandProvider
+    {
+        /// <summary>
+        /// Configure a custom provider command.
+        /// </summary>
+        /// <param name="rootCommand">Root provider command.</param>
+        void ConfigureCommand(Command rootCommand);
+    }
+
+    internal class ProviderCommand : Command
```
derrickstolee, line 14: I'm a little confused as to why these are in the same file. They are not using each other in any visible way.

derrickstolee (approved):
There is certainly a lot going on here, but I appreciate that the tests get a bit simpler.

I hope this makes things easier as you extend into other features.

src/shared/Microsoft.Git.CredentialManager/Commands/ProviderCommand.cs
mjcheetham: That's because I'm an idiot 😜 

I changed the extension point design here a few times, but this doesn't "feel" right. I've moved to a factory style now where an `ICommandProvider` implementing host provider returns a new `ProviderCommand` instance from the `CreateCommand()` method.

src/shared/Microsoft.Git.CredentialManager/Application.cs
mjcheetham: Sure!

mjcheetham:
> There is certainly a lot going on here, but I appreciate that the tests get a bit simpler.

Yes, when we only had the `get|erase|store` commands for Git's consumption just a simple in-house parser was appropriate. Now we're adding more commands with options, arguments etc it feels the correct time to make this someone else's problem. 😉 

> I hope this makes things easier as you extend into other features.

That's the plan!

PR #274, opened 2021-02-02
Drop the native macOS ADAL-based Microsoft Authentication component

mjcheetham:
Remove the ADAL Objective-C based native authentication component for MSA/AAD. We now rely on the system browser flows within MSAL.NET only (or device code flow). If you're using Chrome or Safari you can also supply the correct Intune management certificate to support device CA policies in AAD.

At the same time remove the ability to use an external authentication helper for Microsoft authentication (AAD/MSA). We now always rely on the in-process MSAL library calls. This extension point was only used by the native helper component anyway.


derrickstolee (approved):
Delete all the things!

PR #277, opened 2021-02-02
Harmonise Microsoft Authentication APIs with AAD "v2" and MSAL concepts

mjcheetham:
Refactor the `MicrosoftAuthentication` component API signature to accept a set of scopes (`string[] scopes`) rather than a single "resource ID", and drop the no-longer used remote URL parameter.

Introduce a new wrapper type for the authentication result such that we can drop the dependency on the JWT library and not expose the MSAL `AuthenticationResult` type either. We only need the access token (as a string) and the account UPN.

Note that technically before we should not have been trying to parse the access token as a JWT because there is no guarantee that STS returns JWTs - they should be treated transparently.


derrickstolee (approved):
Code looks clean. As for the actual behavior, I'll need to trust your testing. ☑️

mjcheetham:
> Code looks clean. As for the actual behavior, I'll need to trust your testing. ☑️

There should be no behaviour changes at all. All I'm doing is removing a 'jump' method that turns the `string resourceId` in to a `string[] scope` with `new [] { $"resourceId/.default" }`, and introducing a simple transport/wrapper object for the access token + account ID (rather than trying to extract that from the access token ourselves).

PR #279, opened 2021-02-03
Add ability to override Microsoft Authentication parameters for Azure Repos

mjcheetham:
Add the ability to override the Microsoft authentication parameters in the Azure DevOps API and MSAuth components as used by the Azure Repos provider. This allows us to change the client ID, redirect URI, and the base authority at runtime as a developer (this is not documented or supported outside of development - the [same feature is available for the GitHub provider](https://github.com/microsoft/Git-Credential-Manager-Core/blob/66a573128898b28d46884a643eb6a446ab264562/src/shared/GitHub/GitHubOAuth2Client.cs#L28-L54)).

### DIFF
--- a/src/shared/Microsoft.AzureRepos/AzureDevOpsRestApi.cs
+++ b/src/shared/Microsoft.AzureRepos/AzureDevOpsRestApi.cs
@@ -33,14 +33,14 @@ namespace Microsoft.AzureRepos
         {
             EnsureArgument.AbsoluteUri(organizationUri, nameof(organizationUri));
 
-            const string authorityBase = "https://login.microsoftonline.com/";
-            const string commonAuthority = authorityBase + "common";
+            Uri authorityBase = GetAuthorityBaseUri();
+            var commonAuthority = new Uri(authorityBase, "common");
 
             // We should be using "/common" or "/consumer" as the authority for MSA but since
             // Azure DevOps uses MSA pass-through (an internal hack to support MSA and AAD
             // accounts in the same auth stack), which actually need to consult the "/organizations"
             // authority instead.
-            const string msaAuthority = authorityBase + "organizations";
+            var msaAuthority = new Uri(authorityBase, "organizations");
 
             _context.Trace.WriteLine($"HTTP: HEAD {organizationUri}");
             using (HttpResponseMessage response = await HttpClient.HeadAsync(organizationUri))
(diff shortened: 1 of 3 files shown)


src/shared/Microsoft.AzureRepos/AzureDevOpsRestApi.cs
```diff
@@ -33,14 +33,14 @@ public async Task<string> GetAuthorityAsync(Uri organizationUri)
         {
             EnsureArgument.AbsoluteUri(organizationUri, nameof(organizationUri));
 
-            const string authorityBase = "https://login.microsoftonline.com/";
-            const string commonAuthority = authorityBase + "common";
+            Uri authorityBase = GetAuthorityBaseUri();
+            var commonAuthority = new Uri(authorityBase, "common");
```
derrickstolee, line 37: A good addition, anyway.

src/shared/Microsoft.AzureRepos/AzureDevOpsRestApi.cs
```diff
@@ -72,22 +72,36 @@ public async Task<string> GetAuthorityAsync(Uri organizationUri)
...
+            return commonAuthority.ToString();
+        }
+
+        private Uri GetAuthorityBaseUri()
+        {
+            // Check for developer override value
+            if (_context.Settings.TryGetSetting(
+                AzureDevOpsConstants.EnvironmentVariables.DevAadAuthorityBaseUri,
+                Constants.GitConfiguration.Credential.SectionName, AzureDevOpsConstants.GitConfiguration.Credential.DevAadAuthorityBaseUri,
+                out string redirectUriStr) && Uri.TryCreate(redirectUriStr, UriKind.Absolute, out Uri authorityBase))
```
derrickstolee, line 99: nit: tabbing for the method parameters (at this width, they should also be on their own lines), then also split the `&&` across multiple lines. This applies to the similar `if` statements in other parts of this PR.

derrickstolee (approved):

src/shared/Microsoft.AzureRepos/AzureDevOpsRestApi.cs
mjcheetham: Ooh.. not sure what happened here in my style rules. Fixed.

derrickstolee (approved):

PR #283, opened 2021-02-09
Target .NET Framework 4.7.2 and fix missing binaries in Windows installers

mjcheetham:
When we added support for `System.CommandLine` I neglected to add the extra binaries it brings to the party in the Windows installers. D'oh!

Given the `System.CommandLine` is a .NET Standard targeting library, and we're currently targeting .NET Framework 4.6.1 on Windows, this brings in _a lot_ of compatibility and shim/type forwarding binaries (since `net461` doesn't perfectly implement `netstandard2.0`.. go figure). [We've asked for them to consider targeting a downlevel net461 TFM](https://github.com/dotnet/command-line-api/issues/1194), but in reality we can also just as easily increase our TFM to `net472`, which doesn't require any of those shim binaries.

.NET Framework 4.7.2 is supported on Windows versions all the way back to Windows 7 SP1, so we won't be removing any users from the support matrix. Git for Windows supports Vista, but since GCM Core is already targeting `net461`, it doesn't run on Vista today(!) Vista only supports 4.6(.0) as the latest Framework version.

We could also drop the `System.CommandLine` usage, but I'd rather not. It's worth it IMO.

### DIFF
--- a/.azure-pipelines/templates/windows/pack.yml
+++ b/.azure-pipelines/templates/windows/pack.yml
@@ -6,8 +6,8 @@ steps:
     displayName: Set version variables
 
   - script: |
-      xcopy "out\windows\Installer.Windows\bin\$(configuration)\net461"       "$(Build.StagingDirectory)\publish\"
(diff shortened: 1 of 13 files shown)


jeffhostetler (approved):
This is a rubber stamp.

Issue #261, opened 2021-01-10
Simplify usage in Azure Cloud Shell


aelij:
**Feature description**

Azure Cloud Shell sessions are ephemeral, so setting up PGP is not a great solution. I read that git has an in-memory credential store - https://git-scm.com/docs/git-credential-cache

> The stored credentials never touch the disk, and are forgotten after a configurable timeout. The cache is accessible over a Unix domain socket, restricted to the current user by filesystem permissions.

Would it be possible to configure the credential manager to use this store?

vtbassmatt:
@vtbassmatt to investigate whether [we're](https://github.com/microsoft/Git-Credential-Manager-Core/blob/master/docs/linuxcredstores.md) compatible with the built-in [credential cache](https://git-scm.com/docs/git-credential-cache), and document if so.

vtbassmatt:
Ah, after reading more about the credential cache, this isn't as trivial as I thought. Still not too bad, though:
1. Pull [this](https://github.com/microsoft/Git-Credential-Manager-Core/blob/307e9366da407f2d5b070f875f2cfc413214ddb9/src/shared/Microsoft.Git.CredentialManager/Authentication/AuthenticationBase.cs#L25-L67) out into a process invoker helper
2. Add a new option [here](https://github.com/microsoft/Git-Credential-Manager-Core/blob/66a573128898b28d46884a643eb6a446ab264562/src/shared/Microsoft.Git.CredentialManager/Interop/Linux/LinuxCredentialStore.cs) which delegates CRUD to `git credential-cache`.

PR #287, opened 2021-02-12
Add Git credential cache as backing store on Linux

vtbassmatt:
Fixes https://github.com/microsoft/Git-Credential-Manager-Core/issues/261

- [x] Factor out helper invoker
- [x] Invoke Git helpers using the correct instance of Git
- [x] Add config for using Git credential cache
- [x] Manual testing
  - [x] Simple case: does it store and reuse a credential?
  - [x] After running `git credential-cache exit`, are credentials dropped?
  - [x] Passing arguments like `--timeout` via environment variable works?
- [ ] Add unit tests
- [x] Include in documentation


src/shared/Microsoft.Git.CredentialManager/CredentialCacheStore.cs
```diff
@@ -0,0 +1,81 @@
...
+                "credential-cache get",
+                input
+            ).GetAwaiter().GetResult();
+
+            if (result.ContainsKey("username") && result.ContainsKey("password"))
+            {
+                return new GitCredential(result["username"], result["password"]);
+            }
+
+            return null;
```
mjcheetham, line 35: I wonder if throwing an error, or a warning would be more appropriate if there's no user/pass in the result?

Also makes me wonder what exactly `git-credential-cache` _does_ when it doesn't find a cached credential?

src/shared/Microsoft.Git.CredentialManager/CredentialCacheStore.cs
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        #endregion
+
+        private Dictionary<string, string> MakeGitCredentialsEntry(string service, string account)
+        {
+            var result = new Dictionary<string, string>();
+
+            result["url"] = service;
+            // TODO: "account" is unused, is that right?
```
mjcheetham, line 76: Account should be mapped to the `username` field

mjcheetham:
```suggestion
            result["username"] = account;
```

src/shared/Microsoft.Git.CredentialManager/Interop/Linux/LinuxCredentialStore.cs
```diff
@@ -80,11 +87,16 @@ private void EnsureBackingStore()
...
+                    ValidateCredentialCache();
+                    _backingStore = new CredentialCacheStore(_git, _helper);
+                    break;
+
                 case PlaintextStoreOption:
                     ValidatePlaintext(out string plainStoreRoot);
                     _backingStore = new PlaintextCredentialStore(_fileSystem, plainStoreRoot, ns);
                     break;
-
+
```
mjcheetham, line 99: Extra whitespace??

```suggestion

```

mjcheetham (commented):
Looks good so far.. some comments.

src/shared/Microsoft.Git.CredentialManager/CredentialCacheStore.cs
vtbassmatt: > I wonder if throwing an error [...] would be more appropriate

It's a good question. I was drafting off what the interface currently instructs:
https://github.com/microsoft/Git-Credential-Manager-Core/blob/791173e9dcd55ea7ba84645648a7aa06e8f7d166/src/shared/Microsoft.Git.CredentialManager/ICredentialStore.cs#L16

The primary callsite seems to be in `HostProvider` where it branches on `null`:
https://github.com/microsoft/Git-Credential-Manager-Core/blob/8a699dbc9af32277b3ff721e65d5e3efb7919603/src/shared/Microsoft.Git.CredentialManager/HostProvider.cs#L131

I guess the functional people would tell us to use a `Maybe` here, but since that's not really the style of this codebase, I think leaving it returning `null` is probably fine.

`credential-cache` is a credential helper, so it [should produce nothing](https://git-scm.com/docs/gitcredentials):
> A helper is free to produce a subset, or even no values at all if it has nothing useful to provide.

vtbassmatt:
Tested on an Ubuntu 20.04 machine against a private Azure Repos instance. I don't have a clear idea of how or what needs unit testing, and suggestions would be most welcomed.

src/shared/Microsoft.Git.CredentialManager/Git.cs
```diff
@@ -12,6 +15,20 @@ public interface IGit
         /// <param name="level">Configuration level filter.</param>
         /// <returns>Git configuration.</returns>
         IGitConfiguration GetConfiguration(GitConfigurationLevel level);
+
+        /// <summary>
+        /// Create a Git process that's ready to run.
+        /// </summary>
+        /// <returns>The Git process.</returns>
+        Process CreateProcess(string args);
```
mjcheetham, line 23: We can drop this from the interface now that we've got `InvokeHelperAsync` on the interface.. the `GitProcess` implementation can still call `CreateProcess`, and we don't need to worry about the `TestGit` implementation at all.

docs/linuxcredstores.md
```diff
@@ -87,7 +88,37 @@ export GPG_TTY=$(tty)
 **Note:** Using `/dev/tty` does not appear to work here - you must use the real
 TTY device path, as returned by the `tty` utility.
 
-## 3. Plaintext files
+## 3. Git's built-in [credential cache](https://git-scm.com/docs/git-credential-cache)
+
+```shell
+export GCM_CREDENTIAL_STORE=credentialcache
```
mjcheetham, line 94: Don't forget to update the table of values in both `docs/environment.md` and `docs/configuration.md` for the new option!

src/shared/Microsoft.Git.CredentialManager/Interop/Linux/LinuxCredentialStore.cs
```diff
@@ -11,29 +11,33 @@ public class LinuxCredentialStore : ICredentialStore
     {
         private const string SecretServiceStoreOption = "secretservice";
         private const string GpgStoreOption = "gpg";
+        private const string CredentialCacheOption = "credentialcache";
```
mjcheetham, line 14: I would say we should make this value be either the kebab-case form `credential-cache`, or simply just `cache`.

This leaves the possibility for us to allow arbitrary `git-<cmd>` or `git-credential-<cmd>` command forwarding at some point.. what do you think?

mjcheetham (commented):
Looks good to me! Left a question on the value of the store option, and also we're missing updates to the other settings/reference docs for the new options.

In terms of unit tests.. I am not sure it's worth the effort here, given this is largely around process creation and stdin/out forwarding.

docs/linuxcredstores.md
mjcheetham: ..and the new `_OPTIONS` command line args option too.

src/shared/Microsoft.Git.CredentialManager/Git.cs
vtbassmatt: Oh, sure thing. I added it for parallelism with `InvokeHelperAsync` but yeah, YAGNI.

docs/linuxcredstores.md
vtbassmatt: D'oh, good catch! Thanks!

src/shared/Microsoft.Git.CredentialManager/Interop/Linux/LinuxCredentialStore.cs
vtbassmatt: No opinion personally. `cache` is fine by me.

mjcheetham (approved):
:shipit:

PR #294, opened 2021-02-22
Stop using PATs for Azure Repos (Attempt 2: Electric Boogaloo)

mjcheetham:
Add a new setting that enables the Azure Repos provider to return the Azure Active Directory/Microsoft Account access tokens to Git directly, rather than exchanging them for an Azure DevOps Personal Access Token.

The default mode is to continue using PATs at the moment. The disable PATs set the following config:

```shell
git config --global credential.azreposPATMode false
```

New sub-commands are added to the `git-credential-manager-core azure-repos` command to list and override the binding of user accounts to Azure DevOps organisations.

To avoid overcomplicating things, and a combinatorial explosion of user binding options we support the following user bindings, in order of most preferred:

**1. Git remote URL**

```ini
[remote "origin"]
    url = https://user%40example.com@{org}.visualstudio.com/project/_git/repo
```

This must be set manually by the user as GCM doesn't have enough information about which remote was used for a particular request to be able to update the remote URL with the user.

GCM doesn't directly look this up (except when issuing a `azure-repos list --show-remotes` command) as Git itself will send us the username in the `get` request.

We _ignore_ the username value when the URL hostname is `dev.azure.com` as Azure DevOps decided to use the userinfo part of their clone URLs to include the organisation name (e.g. `https://org@dev.azure.com/org/...`) - as a result these username values cannot be trusted.

The user must encode any special characters like `@` as `%40`.

**2. Organisation binding in local Git configuration**

```ini
[credential "azrepos:org/{org}"]
    username = user@example.com
```

If no username value was provided by Git (from the remote URL), or we are unable to use the provided username value (because it's a `dev.azure.com` hostname), then we consult Git for any username bound to the Azure DevOps organisation name.

**3. Organisation binding in global Git configuration**

We explicitly check Git's local (above) and then global configurations for an Azure DevOps user-organisation binding because GCM will automatically update only these two config "levels" after a successful credential use (`store`).

**"Sign-in" (after a call to `store`)**

Here is a summary of the state of user-organisation bindings before and after signing in with a user "A":

&nbsp;|Description
-|-
A|User to sign-in
B |Another user
-|No user

Global|Local|:arrow_right:|Global|Local
-|-|-|-|-
-|-|:arrow_right:|A|-
-|A|:arrow_right:|A|-
-|B|:arrow_right:|A|-
A|-|:arrow_right:|A|-
A|A|:arrow_right:|A|-
A|B|:arrow_right:|A|-
B|-|:arrow_right:|B|A
B|A|:arrow_right:|B|A
B|B|:arrow_right:|B|A

**"Sign-out" (after a call to `erase`)**

Here is a summary of the state of user-organisation bindings before and after signing-out of an organisation:

&nbsp;|Description
-|-
U|User
X|Do not inherit (valid in local only)
-|No user

Global|Local|:arrow_right:|Global|Local
-|-|-|-|-
-|-|:arrow_right:|-|-
-|U|:arrow_right:|-|-
-|X|:arrow_right:|-|-
U|-|:arrow_right:|U|X
U|U|:arrow_right:|U|X
U|X|:arrow_right:|U|X

When we are called to `store`/`erase` and we're _not_ inside a Git repository (such as when called by some external tool that doesn't follow what Git does exactly) any writes to the `--local` configuration will silently fail.

This replaces [attempt 1](https://github.com/microsoft/Git-Credential-Manager-Core/pull/281) that uses our own file for storing data.


docs/configuration.md
```diff
@@ -333,3 +333,25 @@ Credential: "git:https://bob@github.com/example/myrepo" (user = bob)
 
    https://bob@github.com/example/myrepo
 ```
+
+---
+
+### credential.azreposPATMode
```
vtbassmatt, line 339: Can we get some mention of "legacy" in the variable name? While it defaults to `true` now, we'd like to default it to `false` in the not so distant future. (That might also be worth documenting here...)

docs/usage.md
```diff
@@ -30,3 +30,8 @@ Read the [Git manual](https://git-scm.com/docs/gitcredentials#_custom_helpers) a
 Set your user-level Git configuration (`~/.gitconfig`) to use GCM Core. If you pass
 `--system` to these commands, they act on the system-level Git configuration
 (`/etc/gitconfig`) instead.
+
+### azure-repos
```
vtbassmatt, line 34: Would be nice to have additional documentation about how manual binding works and any automated binding behaviors as a result of "simply" signing in.

vtbassmatt (approved):
Awesome! 🎉

docs/configuration.md
jeffhostetler: Would this be better as a set of values rather than a boolean?
I was initially thinking of just a set like `credential.azreposmode := "pat" | "azure" | ...`
but maybe instead a list of values (in case it is decided that we need to support
both old and new (and newer) ways at the same time.  (This may not make sense, but I'll ask the question.)

just in case the world changes again, or it is decided that we need to support both

src/shared/Microsoft.Git.CredentialManager/GitConfigurationKeyComparer.cs
```diff
@@ -18,12 +18,16 @@ public class GitConfigurationKeyComparer : StringComparer
     {
         public static readonly GitConfigurationKeyComparer Instance = new GitConfigurationKeyComparer();
 
+        public static readonly StringComparer SectionComparer = OrdinalIgnoreCase;
+        public static readonly StringComparer ScopeComparer = Ordinal;
```
bgavrilMS, line 22: Good catch, OAuth2 scopes are indeed case sensitive

src/shared/Microsoft.Git.CredentialManager/GitConfiguration.cs
```diff
@@ -111,7 +111,7 @@ internal GitProcessConfiguration(ITrace trace, GitProcess git, GitConfigurationL
         public void Enumerate(GitConfigurationEnumerationCallback cb)
         {
             string level = GetLevelFilterArg();
-            using (Process git = _git.CreateProcess($"config --null {level} --list"))
+            using (Process git = _git.CreateProcess($"config --null {level} --list --show-scope"))
```
bgavrilMS: Does this have unit tests?

docs/configuration.md
mjcheetham: Either are good ideas. `credential.useLegacyCredentials` for example?

Or `credential.azreposAuthMode := "pat" | "azure"` etc

src/shared/Microsoft.Git.CredentialManager/GitConfigurationKeyComparer.cs
mjcheetham: These are not OAuth2 scopes, but what I call "scopes" for Git configuration entries.. for example

`credential.https://github.com.useHttpPath` (`section.scope.property`)

(nit: I am aware that Git seems to call the different "levels" of config "scopes" - global, local etc)... naming is hard 😅

src/shared/Microsoft.Git.CredentialManager/GitConfiguration.cs
mjcheetham: It does not.. but probably should!

docs/usage.md
mjcheetham: Right. I can work on this.. will need some expert editing to make it understandable to humans!

docs/configuration.md
vtbassmatt: I think also to make the branding gods happy, we should spell out "azureReposAuthMode". Sorry for the wordiness.

mjcheetham: Updated to use @jeffhostetler 's idea. I didn't rename to the full `azureRepos` because we already have several options for the Azure Repos provider that are `azRepos` and consistency is more important than trying to please brand gods on a case that was lost before it was started. 😆

docs/usage.md
mjcheetham: @vtbassmatt I've added a new document outlining the differences between PATs and Azure tokens, as well as how users are bound (and how advanced users can play with them).

docs/azrepos-users-and-tokens.md
```diff
@@ -0,0 +1,226 @@
...
+as you are actively using them to perform Git operations.
+
+These tokens are also securely shared with other Microsoft developer tools
+including the Visual Studio IDE and Azure CLI. This means that as long as you're
+using Git or one of these tools with the same account, you'll never need to
+re-authenticate due to expired tokens!
+
+## User Accounts
+
+In versions of Git Credential Manager that support Azure Access Tokens, the user
```
vtbassmatt, line 42: Worth calling out from which version this is supported?

docs/azrepos-users-and-tokens.md
```diff
@@ -0,0 +1,226 @@
...
+account used to authenticate for a particular Azure DevOps organization will now
+be remembered.
+
+The first time you clone, fetch or push from/to an Azure DevOps organization you
+will be prompted to sign-in and select a user account. Git Credential Manager
```
vtbassmatt, line 47: I know on the Azure DevOps side, they say "identity" for the thing that represents a user. Genuine question, is that the same thing as "account" here (and if so, should we use their terminology)?

vtbassmatt (approved):

docs/azrepos-users-and-tokens.md
mjcheetham: We don't know until a release is made, and the version number is dynamically assigned based on Git commit height and hash 😉 

I'd be looking to change this after the release to include a version number.

docs/azrepos-users-and-tokens.md
mjcheetham: We can say "identity" if we want to be consistent with Azure DevOps.

In AAD there are users and accounts (an account != user.. one can span tenants and clouds, the other is in a particular cloud+tenant).

In Git there is only a "username".

mjcheetham: I guess technically, you're selecting an Azure _account_ when you sign-in, and not an Azure DevOps _identity_.. Just happens that Azure accounts are mapped to an AzDO identity.

Aside: there's also a _FUN_ thing that if you've never signed-in to the web front-end of a particular Azure DevOps organisation, you cannot call any AzDevOps APIs because the "identity" doesn't exist yet - it's only materialised on first sign-in on the web. 🙄

src/shared/Microsoft.Git.CredentialManager/GitConfiguration.cs
```diff
@@ -128,12 +128,63 @@ public void Enumerate(GitConfigurationEnumerationCallback cb)
...
+                    // Read config scope (null terminated)
+                    while (data[i] != '\0')
+                    {
+                        scope.Append(data[i++]);
+                    }
+
+                    // Skip the null terminator
+                    i++;
+
+                    // Read key name (LF terminated)
```
jeffhostetler, line 156: any chance there might be a CRLF present?

jeffhostetler (approved):

src/shared/Microsoft.Git.CredentialManager/GitConfiguration.cs
```diff
@@ -3,117 +3,121 @@
...
-    /// <param name="value">Value of the current configuration entry.</param>
+    /// <param name="entry">Current configuration entry.</param>
     /// <returns>True to continue enumeration, false to stop enumeration.</returns>
-    public delegate bool GitConfigurationEnumerationCallback(string name, string value);
+    public delegate bool GitConfigurationEnumerationCallback(GitConfigurationEntry entry);
 
     public enum GitConfigurationLevel
     {
         All,
-        ProgramData,
```
dscho, line 22: Whoa, that was still in there, wasn't it.

src/shared/Microsoft.Git.CredentialManager/GitConfiguration.cs
```diff
@@ -3,117 +3,121 @@
...
         System,
-        Xdg,
```
dscho, line 24: The `XDG_*` config level still exists...

src/shared/Microsoft.Git.CredentialManager/GitConfiguration.cs
```diff
@@ -3,117 +3,121 @@
...
-            _filterLevel = filterLevel;
         }
 
-        public void Enumerate(GitConfigurationEnumerationCallback cb)
+        public void Enumerate(GitConfigurationLevel level, GitConfigurationEnumerationCallback cb)
         {
-            string level = GetLevelFilterArg();
-            using (Process git = _git.CreateProcess($"config --null {level} --list"))
+            string levelArg = GetLevelFilterArg(level);
+            using (Process git = _git.CreateProcess($"config --null {levelArg} --list --show-scope"))
```
dscho, line 120: I agree that we require the information about the scope. The extensive changes required by that make be a _bit_ uneasy, but I think that cannot be helped.

dscho (approved):
I reviewed this PR purely with my Git for Windows maintainer hat on: how certain can we be that this PR minimizes the risk of breakages in case the feature flag is _not_ turned on (i.e. in case we run _without_ `credential.azreposPatMode=false`)?

Therefore, I did not review the tests, I did not review new methods/classes/interfaces, and I focused on the `IsPersonalAccessTokenMode()` code paths.

Theoretically, there would be a bit room to minimize above-mentioned risk: a couple of refactorings (such as renaming methods, extracting the `GetAzureAccessTokenAsync()` method from `GenerateCredentialAsync()`, etc) could be delayed, the authority cache could potentially be delayed, too.

At this point, however, I would be even warier of trying to disentangle those (good!) changes from this PR: I have tested (quite a bit) the resulting GCM Core with and without the feature flag turned on, and would be loathe to repeat that testing again (and would probably be less diligent about it due to sheer impatience about re-testing something I already tested, anyway).

So my vote is to go forward with it, merge it, release a new GCM Core version immediately. That way, we always have the escape hatch to downgrade GCM Core to GCM 2.0.374-beta (which was released, i.e. it is essentially GCM Core without this here PR).

src/shared/Microsoft.AzureRepos/AzureReposHostProvider.cs
```diff
@@ -67,61 +77,101 @@ public bool IsSupported(HttpResponseMessage response)
...
+            // for user account lookups when getting Azure Access Tokens.
+            Uri remoteUri = input.GetRemoteUri(includeUser: true);
+
+            if (UsePersonalAccessTokens())
             {
-                _context.Trace.WriteLine("Existing credential found.");
+                string service = GetServiceName(input);
+                string account = GetAccountNameForCredentialQuery(input);
+
+                _context.Trace.WriteLine($"Looking for existing PAT in store with service={service} account={account}...");
```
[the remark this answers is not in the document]
mjcheetham: Trace messaging is very much not "public API" (in fact, nothing in GCM is "public API" and is subject to change). The only thing we can guarantee against with being called by Git (via `git-credential` and friends).

I've changed trace messaging before, and no one has complained :)

src/shared/Microsoft.Git.CredentialManager/Application.cs
```diff
@@ -84,23 +87,12 @@ protected override async Task<int> RunInternalAsync(string[] args)
...
-                    WriteException(e);
-                }
+            var parser = new CommandLineBuilder(rootCommand)
+                .UseDefaults()
+                .UseExceptionHandler(OnException)
+                .Build();
 
-                return -1;
-            }
+            return await parser.InvokeAsync(args);
```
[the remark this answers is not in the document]
mjcheetham: The problem is the "current" release's handling of broken exceptions is "wrong" and different to the 2.30 bundled copy.
This change restores the "existing" behaviour before we added System.CommandLine.

src/shared/Microsoft.Git.CredentialManager/GitConfiguration.cs
mjcheetham: It still exists, but we never consider it explicitly in GCM. It'll still be read by Git when we do 'unscoped' reads.

src/shared/Microsoft.AzureRepos/AzureReposHostProvider.cs
```diff
@@ -304,7 +424,7 @@ public Task ConfigureAsync(ConfigurationTarget target)
             else
             {
                 _context.Trace.WriteLine("Setting Git configuration 'credential.useHttpPath' to 'true' for https://dev.azure.com...");
-                targetConfig.Set(useHttpPathKey, "true");
+                targetConfig.Set(configurationLevel, useHttpPathKey, "true");
```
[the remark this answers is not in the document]
mjcheetham: Correct. We still throw exceptions in the same way (when the config level is == `All`).

src/shared/Microsoft.AzureRepos/AzureReposHostProvider.cs
```diff
@@ -67,61 +77,101 @@ public bool IsSupported(HttpResponseMessage response)
...
-                _context.Trace.WriteLine("Creating new credential...");
-                credential = await GenerateCredentialAsync(input);
-                _context.Trace.WriteLine("Credential created.");
+                throw new Exception("Unencrypted HTTP is not supported for Azure Repos. Ensure the repository remote URL is using HTTPS.");
             }
-            else
+
+            // Include any user from the URL/request here so that we may use it as an override
+            // for user account lookups when getting Azure Access Tokens.
+            Uri remoteUri = input.GetRemoteUri(includeUser: true);
```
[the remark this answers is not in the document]
mjcheetham: > Given that we want to phase out PAT usage anyway

I'm thinking that we actually don't want to remove the PAT options, but make them (eventually) the non-default option.

PATs might still be preferable for some users, and may even be required (for various technical or environmental limitations with AAD tokens).

src/shared/Microsoft.Git.CredentialManager/GitConfiguration.cs
mjcheetham: Yep, but was totally unused in the GCM code anyway.. along with the `Xdg` member. Hence consolidating them to `Unknown` (again, just a name for values we don't care about).

src/shared/Microsoft.Git.CredentialManager/GitConfiguration.cs
mjcheetham: Yeah, this is very much a big change to how we parse config output. But it's the only way I can see to do this whilst also adding the required scope data.

docs/azrepos-users-and-tokens.md
```diff
@@ -0,0 +1,226 @@
...
+
+This is because `dev.azure.com`-style Azure DevOps remote URLs include the
+organization name in the user information part of the URL
+(`https://{org}@dev.azure.com/{org}/...`). Git Credential Manager is unable to
+differentiate between a URL with an organization name or a real user account for
+this style of remote URL.
+
+Bindings will always be shown as `(inherit)` for `dev.azure.com` remote URLs.
+
+---
```
mbarati0040, line 226: #

mbarati0040 (commented):
#

mbarati0040 (approved):

src/shared/Microsoft.AzureRepos/AzureReposHostProvider.cs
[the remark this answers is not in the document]
dscho: Sounds good!

src/shared/Microsoft.Git.CredentialManager/GitConfiguration.cs
```diff
@@ -128,12 +128,63 @@ public void Enumerate(GitConfigurationEnumerationCallback cb)
...
+                    while (data[i] != '\0')
+                    {
+                        scope.Append(data[i++]);
+                    }
+
+                    // Skip the null terminator
+                    i++;
+
+                    // Read key name (LF terminated)
+                    while (data[i] != '\n')
```
derrickstolee, line 157: How much should we be protecting against malformed data? Should we be `throw`ing if `i >= data.length`?

This feels like a pedantic complaint, but I'd hate to track down an `IndexOutOfBoundsException` instead of  a message saying "Output was malformed".

docs/configuration.md
```diff
@@ -333,3 +333,26 @@ Credential: "git:https://bob@github.com/example/myrepo" (user = bob)
 
    https://bob@github.com/example/myrepo
 ```
+
+---
+
+### credential.azreposCredentialType _(experimental)_
```
derrickstolee, line 339: hm. this is different than what I had configured. it must have changed. I'll update and re-install. Do you have an updated installer link?

docs/environment.md
```diff
@@ -424,3 +424,32 @@ export GCM_MSAUTH_FLOW="devicecode"
...
+
+```batch
+SET GCM_AZREPOS_CREDENTIALTYPE="false"
+```
+
+##### macOS/Linux
+
+```bash
+export GCM_AZREPOS_CREDENTIALTYPE="false"
+```
```
derrickstolee, line 453: Do you really intend for the environment variable to have different values to the config value?

src/shared/Microsoft.AzureRepos/AzureReposHostProvider.cs
```diff
@@ -278,6 +344,41 @@ private static string GetAccountNameForCredentialQuery(InputArguments input)
...
+        private bool UsePersonalAccessTokens()
+        {
+            // Default to using PATs whilst the Azure AT functionality is being tested
+            const bool defaultValue = true;
+
+            if (_context.Settings.TryGetSetting(
+                AzureDevOpsConstants.EnvironmentVariables.CredentialType,
+                KnownGitCfg.Credential.SectionName,
+                AzureDevOpsConstants.GitConfiguration.Credential.CredentialType,
+                out string valueStr))
```
derrickstolee, line 401: So, this loads from environment the same way it loads from the config? Then the docs are definitely wrong and should use `azure` as their value.

docs/azrepos-users-and-tokens.md
```diff
@@ -0,0 +1,210 @@
...
+Azure Access Tokens is the generic term for tokens issued by Azure Active
+Directory for either Work and School Accounts (AAD tokens) or Personal Accounts
+(Microsoft Account/MSA tokens).
+
+Azure DevOps supports Git authentication using Azure Access Tokens as well as
+PATs. Azure Access Tokens created by Git Credential Manager are scoped to Azure
+DevOps only.
+
+Unlike PATs, Azure Access Tokens get automatically refreshed and renewed as long
+as you are actively using them to perform Git operations.
```
derrickstolee, line 33: Am I reading this right that if we enable background maintenance, then the hourly prefetch action will keep the token alive indefinitely?

derrickstolee (changes requested):
One technical change and a few doc comments. What a major effort!

docs/azrepos-users-and-tokens.md
johnterickson: > if you've never signed-in to the web front-end of a particular Azure DevOps organisation, you cannot call any AzDevOps APIs

Let me know if you are seeing this - it means that someone forgot to turn on a feature flag.

docs/azrepos-users-and-tokens.md
johnterickson: AAD has access tokens and refresh tokens. It could either be helpful or confusing to call out the difference here - I'm not sure :)

https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-configurable-token-lifetimes#token-lifetime-policies-for-access-saml-and-id-tokens

docs/azrepos-users-and-tokens.md
```diff
@@ -0,0 +1,210 @@
...
+
+These tokens are also securely shared with other Microsoft developer tools
+including the Visual Studio IDE and Azure CLI. This means that as long as you're
+using Git or one of these tools with the same account, you'll never need to
+re-authenticate due to expired tokens!
+
+## User Accounts (experimental)
+
+In versions of Git Credential Manager that support Azure Access Tokens, the user
```
johnterickson, line 42: I'm not sure `Azure Access Tokens` is correct - many Azure properties have token concepts.   Perhaps 'Microsoft identity platform access tokens'? That's from the official docs, but is a mouthful! https://docs.microsoft.com/en-us/azure/active-directory/develop/access-tokens

docs/environment.md
mjcheetham: Whoops! Nope. Will fix.

docs/azrepos-users-and-tokens.md
mjcheetham: The document is simplifying somewhat (on purpose). The access tokens only last for an hour (typically) and are continually recreated from the longer-lived refresh token. The refresh token is _also_ renewed automagically by MSAL + ESTS as long as you're using the RT to mint a new AT. As the RT is getting close to expiry, ESTS will return a new RT along with the new AT.

> Am I reading this right that if we enable background maintenance, then the hourly prefetch action will keep the token alive indefinitely?

So yes, as long as we're using the ATs (and therefore using the RT) periodically, we'll never have an expired token.

The only cases where tokens would "expire" would be if the tenant administrator decided to enforce a conditional access policy to limit RT renewal lifetime, or make it conditional on something like being Intune/AAD joined.

docs/azrepos-users-and-tokens.md
mjcheetham: Yeah.. but maaaan is that a mouthful! @bgavrilMS do you have any recommendations here for a shorthand of "Microsoft identity platform access token"?

docs/configuration.md
mjcheetham: Not yet! Will send one over shortly.

src/shared/Microsoft.AzureRepos/AzureReposHostProvider.cs
```diff
@@ -67,61 +77,101 @@ public bool IsSupported(HttpResponseMessage response)
...
             }
             else
             {
-                _context.Trace.WriteLine("No credential was erased.");
+                _context.Trace.WriteLine($"Signing out of organization '{orgName}'...");
+                _bindingManager.SignOut(orgName);
             }
 
+            // Clear the authority cache in case this was the reason for failure
+            _authorityCache.EraseAuthority(orgName);
```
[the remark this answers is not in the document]
mjcheetham: Done! Authority cache, and user bindings is only done on the non-PAT code paths!

docs/azrepos-users-and-tokens.md
mjcheetham: I know this was the case back when GCM Core was first out. I remember hitting this exact issue and having to include support for surfacing human readable error messages (telling you do go log in via the web first) from Azure DevOps' REST API endpoints.

src/shared/Microsoft.Git.CredentialManager/GitConfiguration.cs
mjcheetham: @derrickstolee I've added one more commit to check for unexpected end of data and to trace and break from the parsing loop.

derrickstolee (approved):
Thanks for the parsing updates. This satisfies my only concern.

PR #299, opened 2021-03-01
Add support for the `NO_PROXY` bypass environment variable

### LINKED ISSUE #297: Respect NO_PROXY environment variable / configuration
**Feature description**

Hello, 

same report as in https://github.com/microsoft/Git-Credential-Manager-for-Windows/issues/876, it appear there's no `NO_PROXY` support in GCMC (whatever the source is): https://github.com/microsoft/Git-Credential-Manager-Core/blob/v2.0.318-beta/src/shared/Microsoft.Git.CredentialManager/Settings.cs#L341

So git is properly handling requests, but GCMC can leaks informations through the proxy while it shouldn't.

Could GCMC handle this environment variable the same way gitdoes, which is https://curl.haxx.se/docs/manual.html (§Environment Variables, as per https://github.com/git/git/search?q=no_proxy&unscoped_q=no_proxy)?

Best,

mjcheetham:
Add support for the cURL environment variable (also respected by Git) to list a set of host names that should connect directly (bypass), and not connect via the configured proxy.

Fixes #297


docs/netconfig.md
```diff
@@ -55,6 +55,22 @@ GCM Core supports other ways of configuring a proxy for convenience and compatib
    - `ALL_PROXY`
 1. `GCM_HTTP_PROXY` environment variable (_**only** respected by GCM; **deprecated**_)
 
+### Bypassing addresses
+
+In some circumstances you may wish to bypass a configured proxy for specific
+addresses. GCM Core supports the cURL environment variable `NO_PROXY` for this
+scenariom, as does Git itself.
+
+The `NO_PROXY` environment variable should contain a comma (`,`) seperated list
```
dscho, line 64:
```suggestion
The `NO_PROXY` environment variable should contain a comma (`,`) separated list
```

src/shared/Microsoft.Git.CredentialManager/Settings.cs
```diff
@@ -358,6 +397,30 @@ bool TryGetUriSetting(string envarName, string section, string property, out Uri
...
+                }.Uri;
+
+                // Extract the username and password from the URI if present
+                uri.TryGetUserInfo(out string userName, out string password);
+
+                // Get the proxy bypass host names
+                var bypassHosts = new List<string>();
+                if (_environment.Variables.TryGetValue(KnownEnvars.CurlNoProxy, out string noProxyStr))
+                {
+                    bypassHosts = noProxyStr.Split(',').ToList();
```
dscho, line 418: Curl actually _also_ allows separating by spaces: https://github.com/curl/curl/blob/313faa1bcbd2bcc957e660d98360a82c15fe7c07/lib/url.c#L2181. Do we want to do the same?

dscho (approved):
Looks good to me!

src/shared/Microsoft.Git.CredentialManager/Settings.cs
mjcheetham: @dscho Is it mutually exclusive or split on either?

```text
foo.com,bar.com example.com

-->

foo.com
bar.com
example.com
```

dscho: As far as I read the code, it's either. I.e. something like `NO_PROXY.split(/[ ,]+/)`.

mjcheetham: Added!

dscho (approved):
Nice!

PR #306, opened 2021-03-10
Tell MSAL about any parent window handles GCM knows about

mjcheetham:
MSAL supports parenting windows it creates to another window.
GCM also supports being told about a parent window by the caller.

Pass through this parent window to MSAL if we have it!

### DIFF
--- a/src/shared/Microsoft.Git.CredentialManager/Authentication/MicrosoftAuthentication.cs
+++ b/src/shared/Microsoft.Git.CredentialManager/Authentication/MicrosoftAuthentication.cs
@@ -197,6 +197,14 @@ namespace Microsoft.Git.CredentialManager.Authentication
                 appBuilder.WithLogging(OnMsalLogMessage, LogLevel.Verbose, enablePiiLogging, false);
             }
 
+            // If we have a parent window ID we should tell MSAL about it so it can parent any authentication dialogs
+            // correctly. We only support this on Windows right now as MSAL only supports embedded/dialogs on Windows.
+            if (PlatformUtils.IsWindows() && !string.IsNullOrWhiteSpace(Context.Settings.ParentWindowId) &&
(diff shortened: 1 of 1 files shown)


jeffhostetler (approved):

src/shared/Microsoft.Git.CredentialManager/Authentication/MicrosoftAuthentication.cs
```diff
@@ -197,6 +197,14 @@ private async Task<IPublicClientApplication> CreatePublicClientApplicationAsync(
                 appBuilder.WithLogging(OnMsalLogMessage, LogLevel.Verbose, enablePiiLogging, false);
             }
 
+            // If we have a parent window ID we should tell MSAL about it so it can parent any authentication dialogs
+            // correctly. We only support this on Windows right now as MSAL only supports embedded/dialogs on Windows.
+            if (PlatformUtils.IsWindows() && !string.IsNullOrWhiteSpace(Context.Settings.ParentWindowId) &&
+                int.TryParse(Context.Settings.ParentWindowId, out int hWndInt) && hWndInt > 0)
```
dscho, line 203: Nice. This handles `INVALID_HANDLE_VALUE` as well: It is `-1`, cast to `HANDLE`: https://github.com/git-for-windows/git-sdk-64/blob/8be5d7823baeb3ff55a4aba9254fd3ca4e96e662/mingw64/x86_64-w64-mingw32/include/handleapi.h#L16

dscho (approved):
Nice!

src/shared/Microsoft.Git.CredentialManager/Authentication/MicrosoftAuthentication.cs
mjcheetham: MSAL only wants to be told about "real" handles. Some library versions don't handle bad/invalid handles well.

PR #308, opened 2021-03-10
Run postinstall and uninstall configure commands as current user

### LINKED ISSUE #186: Installing Cask on macOS changes ~/.gitconfig ownership to root
The `git-credential-manager-core` Cask requires installation as root because root is required to install the internal .pkg file.
GCM Core has a postinstall action that calls the `configure` command, that internally calls `git config --global ... ...` which, being run as root, will change the ownership of the `~/.gitconfig` file.

We need a way to have these post install actions _not_ run as root **OR** change the Cask to run `configure --system` which will update the `git config --system` config file (`/etc/gitconfig`) instead.

The latter suggested change of course will make GCM Core a machine-level installation, and not per user. Perhaps this is the 'correct' design however.

Also see https://github.com/microsoft/scalar/issues/443.

mjcheetham:
In the macOS installer postinstall script we run the `configure` command to get GCM to configure the current user's credential helper as GCM.

However, the postinstall script is often run as root (because that's how `installer` works out of the box), meaning although GCM will be writing to the ~/.gitconfig file, it will be doing so from a process running as root.

To avoid having root take ownership of ~/.gitconfig we run `sudo -u` to run the `configure` command as the real user (not root).

Also update the uninstall script to do the same thing (run `unconfigure` as the non-root user).

Fixes #186

### DIFF
--- a/src/osx/Installer.Mac/uninstall.sh
+++ b/src/osx/Installer.Mac/uninstall.sh
@@ -10,9 +10,9 @@ then
 	exit $?
 fi
 
-# Unconfigure
+# Unconfigure (as the current user)
 echo "Unconfiguring credential helper..."
-"$GCMBIN" unconfigure
+sudo -u `/usr/bin/logname` "$GCMBIN" unconfigure
 
 # Remove symlink
 if [ -L /usr/local/bin/git-credential-manager-core ]
--- a/src/osx/Installer.Mac/scripts/postinstall
+++ b/src/osx/Installer.Mac/scripts/postinstall
@@ -30,7 +30,8 @@ fi
 mkdir -p /usr/local/bin
 /bin/ln -Fs "$INSTALL_DESTINATION/git-credential-manager-core" /usr/local/bin/git-credential-manager-core
 
-# Configure GCM for the current user
-"$INSTALL_DESTINATION/git-credential-manager-core" configure
+# Configure GCM for the current user (running as the current user to avoid root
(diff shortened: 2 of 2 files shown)


dscho (approved):

src/osx/Installer.Mac/uninstall.sh
```diff
@@ -10,9 +10,9 @@ then
 	exit $?
 fi
 
-# Unconfigure
+# Unconfigure (as the current user)
 echo "Unconfiguring credential helper..."
-"$GCMBIN" unconfigure
+sudo -u `logname` "$GCMBIN" unconfigure
```
dscho, line 15: According to https://ss64.com/osx/logname.html, this is the right thing to do. It's too bad that this is a bit asymmetric from the "install" case, where we use `${USER}` instead...

jeffhostetler: Should we be careful here?  That is, use /usr/bin/logname inside the backticks.

mjcheetham: Updated to the absolute path.

src/osx/Installer.Mac/uninstall.sh
```diff
@@ -10,9 +10,9 @@ then
 	exit $?
 fi
 
-# Unconfigure
+# Unconfigure (as the current user)
```
elliottmoore31-sudo: Stop

Issue #313, opened 2021-03-17
fatal: Failed to enumerate all Git configuration entries [129] error: unknown option `show-scope'


eisber:
**Which version of GCM Core are you using?**

2.0.394-beta+3fc6791abf

Also: 

Linux version 5.4.72-microsoft-standard-WSL2 (oe-user@oe-host) (gcc version 8.2.0 (GCC)) #1 SMP Wed Oct 28 23:40:43 UTC 2020
git version 2.25.1 (latest available on Ubuntu)

**Which Git host provider are you trying to connect to?**

* [x] Azure DevOps
* [ ] Azure DevOps Server (TFS/on-prem)
* [ ] GitHub
* [ ] GitHub Enterprise
* [ ] Bitbucket
* [ ] Other - please describe

**Can you access the remote repository directly in the browser using the remote URL?**

From a terminal, run `git remote -v` to see your remote URL.

* [x] Yes
* [ ] No, I get a permission error
* [x] No, for a different reason - please describe

---

**_[Azure DevOps only]_ What format is your remote URL?**

* [ ] Not applicable
* [ ] https://dev.azure.com/`{org}`/...
* [ ] https://`{org}`@dev.azure.com/`{org}`/...
* [x] https://`{org}`.visualstudio.com/...

**_[Azure DevOps only]_ If the account picker shows more than one identity as you authenticate, check that you selected the same one that has access on the web.**

* [x] Not applicable
* [ ] I only see one identity
* [ ] I checked each identity and none worked

---

**Expected behavior**

I am authenticated and my Git operation completes successfully.

**Actual behavior**

fatal: Failed to enumerate all Git configuration entries [129]
error: unknown option `show-scope'
usage: git config [<options>]

Config file location
    --global              use global config file
    --system              use system config file
    --local               use repository config file
    --worktree            use per-worktree config file
    -f, --file <file>     use given config file
    --blob <blob-id>      read config from given blob object

Action
    --get                 get value: name [value-regex]
    --get-all             get all values: key [value-regex]
    --get-regexp          get values for regexp: name-regex [value-regex]
    --get-urlmatch        get value specific for the URL: section[.var] URL
    --replace-all         replace all matching variables: name value [value_regex]
    --add                 add a new variable: name value
    --unset               remove a variable: name [value-regex]
    --unset-all           remove all matches: name [value-regex]
    --rename-section      rename section: old-name new-name
    --remove-section      remove a section: name
    -l, --list            list all
[...]

PR #318, opened 2021-03-24
Drop the git config --show-scope option usage

mjcheetham:
The Git config `--show-scope` option was only introduced in Git from
version 2.26 onwards. The latest version of Git available in some
distributions of Linux (or macOS) is often older.

We only needed to know the scope of configuration values in one
particular call site: reading all Azure Repos user bindinds.

Replace the single `IGitConfiguration::Enumerate` call with two calls to
`Enumerate`, one for the global scope, and one for the local one.

Drop the --show-scope option parsing.

Fixes: #313


dscho (approved):
That's quite elegant! No unnecessary parsing because you simply use the scope that you wanted to have in the first place. I like it.

PR #323, opened 2021-04-09
Add support for Windows broker (WAM)

### LINKED ISSUE #211: Support FIDO and features like Windows Hello
This is a tracking issue for MSAL's feature work to add support for "WAM" (that would enable things like Windows Hello directly without needing to use the system browser).

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2124
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2125
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2126
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1339

mjcheetham:
Add support for broker-assisted authentication on Windows using "WAM" (Web Authentication Manager) as provided by the MSAL.Desktop library.

The `GCM_MSAUTH_USEBROKER` environment variable or the `credential.msauthUseBroker` configuration option will control if WAM is enabled or not. ~By default WAM _is_ enabled.~

This ended up being simpler then I expected. Looks like the VS AAD app registration has now been updated to have the required WAM redirect URI which was previously a technical blocker.

Fixes #211 

**Update:** WAM is now default _disabled_ due to concerns about the maturity of the UX and technology. cc: @bgavrilMS

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -256,11 +256,12 @@ git config --global credential.plaintextStorePath /mnt/external-drive/credential
 
 Specify which authentication flow should be used when performing Microsoft authentication and an interactive flow is required.
 
-Defaults to the value `auto`.
+Defaults to `auto`.
 
 **Note:** If [`credential.msauthUseBroker`](#credentialmsauthusebroker) is set
 to `true` and the operating system authentication broker is available, all flows
-will be delegated to the broker; this setting has no effect.
+will be delegated to the broker. If both of those things are true, then the
+value of `credential.msauthFlow` has no effect.
 
 Value|Authentication Flow
 -|-
@@ -283,7 +284,7 @@ git config --global credential.msauthFlow devicecode
 
 Use the operating system account manager where available.
 
-Defaults to the value `true`.
+Defaults to `true`.
 
 Value|Description
 -|-
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -284,17 +284,17 @@ git config --global credential.msauthFlow devicecode
 
 Use the operating system account manager where available.
 
-Defaults to `true`.
+Defaults to `false`. This default is subject to change in the future.
 
 Value|Description
 -|-
-`true` _(default)_|Use the operating system account manager as an authentication broker.
-`false`|Do not use the broker.
+`true`|Use the operating system account manager as an authentication broker.
+`false` _(default)_|Do not use the broker.
 
 #### Example
 
 ```shell
-git config --global credential.msauthUseBroker false
+git config --global credential.msauthUseBroker true
 ```
 
 **Also see: [GCM_MSAUTH_USEBROKER](environment.md#GCM_MSAUTH_USEBROKER)**
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -258,6 +258,10 @@ Specify which authentication flow should be used when performing Microsoft authe
 
 Defaults to the value `auto`.
 
+**Note:** If [`credential.msauthUseBroker`](#credentialmsauthusebroker) is set
+to `true` and the operating system authentication broker is available, all flows
+will be delegated to the broker; this setting has no effect.
+
 Value|Authentication Flow
 -|-
 `auto` _(default)_|Select the best option depending on the current environment and platform.
@@ -272,6 +276,27 @@ git config --global credential.msauthFlow devicecode
 ```
 
 **Also see: [GCM_MSAUTH_FLOW](environment.md#GCM_MSAUTH_FLOW)**
+
+---
+
+### credential.msauthUseBroker
+
+Use the operating system account manager where available.
+
+Defaults to the value `true`.
+
+Value|Description
+-|-
+`true` _(default)_|Use the operating system account manager as an authentication broker.
+`false`|Do not use the broker.
+
+#### Example
+
+```shell
+git config --global credential.msauthUseBroker false
+```
+
+**Also see: [GCM_MSAUTH_USEBROKER](environment.md#GCM_MSAUTH_USEBROKER)**
 
 ---
 
--- a/docs/environment.md
+++ b/docs/environment.md
@@ -436,17 +436,17 @@ export GCM_MSAUTH_FLOW="devicecode"
 
 Use the operating system account manager where available.
 
-Defaults to `true`.
+Defaults to `false`. This default is subject to change in the future.
 
 Value|Description
 -|-
-`true` _(default)_|Use the operating system account manager as an authentication broker.
-`false`|Do not use the broker.
+`true`|Use the operating system account manager as an authentication broker.
+`false` _(default)_|Do not use the broker.
 
 ##### Windows
 
 ```batch
-SET GCM_MSAUTH_USEBROKER="false"
+SET GCM_MSAUTH_USEBROKER="true"
 ```
 
 ##### macOS/Linux
--- a/docs/environment.md
+++ b/docs/environment.md
@@ -402,11 +402,12 @@ export GCM_PLAINTEXT_STORE_PATH=/mnt/external-drive/credentials
 
 Specify which authentication flow should be used when performing Microsoft authentication and an interactive flow is required.
 
-Defaults to the value `auto`.
+Defaults to `auto`.
 
 **Note:** If [`GCM_MSAUTH_USEBROKER`](#gcm_msauth_usebroker) is set to `true`
 and the operating system authentication broker is available, all flows will be
-delegated to the broker; this setting has no effect.
+delegated to the broker. If both of those things are true, then the value of
+`GCM_MSAUTH_FLOW` has no effect.
 
 Value|Authentication Flow
 -|-
@@ -435,7 +436,7 @@ export GCM_MSAUTH_FLOW="devicecode"
 
 Use the operating system account manager where available.
 
-Defaults to the value `true`.
+Defaults to `true`.
 
 Value|Description
 -|-
--- a/docs/environment.md
+++ b/docs/environment.md
@@ -404,6 +404,10 @@ Specify which authentication flow should be used when performing Microsoft authe
 
 Defaults to the value `auto`.
 
+**Note:** If [`GCM_MSAUTH_USEBROKER`](#gcm_msauth_usebroker) is set to `true`
+and the operating system authentication broker is available, all flows will be
+delegated to the broker; this setting has no effect.
+
 Value|Authentication Flow
 -|-
 `auto` _(default)_|Select the best option depending on the current environment and platform.
@@ -424,6 +428,33 @@ export GCM_MSAUTH_FLOW="devicecode"
 ```
 
 **Also see: [credential.msauthFlow](configuration.md#credentialmsauthflow)**
+
+---
+
+### GCM_MSAUTH_USEBROKER
+
+Use the operating system account manager where available.
+
+Defaults to the value `true`.
+
+Value|Description
+-|-
+`true` _(default)_|Use the operating system account manager as an authentication broker.
+`false`|Do not use the broker.
+
+##### Windows
+
+```batch
+SET GCM_MSAUTH_USEBROKER="false"
+```
+
+##### macOS/Linux
+
+```bash
+export GCM_MSAUTH_USEBROKER="false"
+```
+
+**Also see: [credential.msauthUseBroker](configuration.md#credentialmsauthusebroker)**
 
 ---
 
--- a/src/shared/Microsoft.Git.CredentialManager/Authentication/MicrosoftAuthentication.cs
+++ b/src/shared/Microsoft.Git.CredentialManager/Authentication/MicrosoftAuthentication.cs
@@ -400,26 +400,29 @@ namespace Microsoft.Git.CredentialManager.Authentication
 
         private bool CanUseBroker()
         {
-            // We only support the broker on Windows 10 and .NET Framework
 #if NETFRAMEWORK
-            if (Context.SessionManager.IsDesktopSession && PlatformUtils.IsWindows10())
+            // We only support the broker on Windows 10 and require an interactive session
+            if (!Context.SessionManager.IsDesktopSession || !PlatformUtils.IsWindows10())
             {
-                // Default to using the OS broker
-                const bool defaultValue = true;
+                return false;
+            }
 
-                if (Context.Settings.TryGetSetting(Constants.EnvironmentVariables.MsAuthUseBroker,
+            // Default to using the OS broker
+            const bool defaultValue = true;
+
+            if (Context.Settings.TryGetSetting(Constants.EnvironmentVariables.MsAuthUseBroker,
                     Constants.GitConfiguration.Credential.SectionName,
                     Constants.GitConfiguration.Credential.MsAuthUseBroker,
                     out string valueStr))
-                {
-                    return valueStr.ToBooleanyOrDefault(defaultValue);
-                }
-
-                return defaultValue;
+            {
+                return valueStr.ToBooleanyOrDefault(defaultValue);
             }
-#endif
 
+            return defaultValue;
+#else
+            // OS broker requires .NET Framework right now until we migrate to .NET 5.0 (net5.0-windows10.x.y.z)
             return false;
+#endif
         }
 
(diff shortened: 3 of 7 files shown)


docs/configuration.md
```diff
@@ -258,6 +258,10 @@ Specify which authentication flow should be used when performing Microsoft authe
 
 Defaults to the value `auto`.
 
+**Note:** If [`credential.msauthUseBroker`](#credentialmsauthusebroker) is set
+to `true` and the operating system authentication broker is available, all flows
+will be delegated to the broker; this setting has no effect.
```
dscho, line 263: By "this setting" you mean `credential.msauthFlow`, and by "has no effect" you mean that `.msauthUseBroker` overrides `.msauthFlow`, right?

docs/environment.md
```diff
@@ -404,6 +404,10 @@ Specify which authentication flow should be used when performing Microsoft authe
 
 Defaults to the value `auto`.
 
+**Note:** If [`GCM_MSAUTH_USEBROKER`](#gcm_msauth_usebroker) is set to `true`
+and the operating system authentication broker is available, all flows will be
+delegated to the broker; this setting has no effect.
```
dscho, line 409: By "this setting" you mean `GCM_MSAUTH_FLOW`, correct? And by "has no effect" you mean to imply that `_USEBROKER` overrides `_FLOW`, correct?

docs/environment.md
```diff
@@ -427,6 +431,33 @@ export GCM_MSAUTH_FLOW="devicecode"
...
+SET GCM_MSAUTH_USEBROKER="false"
+```
+
+##### macOS/Linux
+
+```bash
+export GCM_MSAUTH_USEBROKER="false"
+```
+
+**Also see: [credential.msauthUseBroker](configuration.md#credentialmsauthusebroker)**
```
dscho, line 457: Does the environment variable override the config setting, as is custom in Git?

src/shared/Microsoft.Git.CredentialManager/Authentication/MicrosoftAuthentication.cs
```diff
@@ -82,48 +97,61 @@ public MicrosoftAuthentication(ICommandContext context)
...
+                            Context.Trace.WriteLine("Performing interactive auth with device code...");
+                            // We don't have a way to display a device code without a terminal at the moment
+                            // TODO: introduce a small GUI window to show a code if no TTY exists
+                            ThrowIfTerminalPromptsDisabled();
+                            result = await app.AcquireTokenWithDeviceCode(scopes, ShowDeviceCodeInTty).ExecuteAsync();
+                            break;
+
+                        default:
+                            goto case MicrosoftAuthenticationFlowType.Auto;
+                    }
```
dscho, line 154: The indentation change makes this harder to review than I'd like, but that cannot be helped, I guess.

src/shared/Microsoft.Git.CredentialManager/Authentication/MicrosoftAuthentication.cs
```diff
@@ -205,6 +234,15 @@ private async Task<IPublicClientApplication> CreatePublicClientApplicationAsync(
                 appBuilder.WithParentActivityOrWindow(() => new IntPtr(hWndInt));
             }
 
+            // On Windows 10 & .NET Framework try and use the WAM broker
+            if (enableBroker && PlatformUtils.IsWindows10())
+            {
+#if NETFRAMEWORK
+                appBuilder.WithExperimentalFeatures();
```
dscho, line 241: Experimental? That sounds safe... 
image

src/shared/Microsoft.Git.CredentialManager/Authentication/MicrosoftAuthentication.cs
```diff
@@ -360,6 +398,30 @@ public HttpClient GetHttpClient()
...
+                    out string valueStr))
+                {
+                    return valueStr.ToBooleanyOrDefault(defaultValue);
+                }
+
+                return defaultValue;
+            }
+#endif
+
+            return false;
```
dscho, line 422: Personal preference: if there are ways to return from a function early, do so, as it reduces cognitive load for the reader. In this instance:

```c#
if (!!context.SessionManager.IsDesktopSession || !PlatformUtils.IsWindows10())
    return false;

[...]
```

src/shared/Microsoft.Git.CredentialManager/PlatformUtils.cs
```diff
@@ -20,6 +20,26 @@ public static PlatformInformation GetPlatformInformation()
...
+            if (!IsWindows())
+            {
+                return false;
+            }
+
+            // Implementation of version checking was taken from:
+            // https://github.com/dotnet/runtime/blob/6578f257e3be2e2144a65769706e981961f0130c/src/libraries/System.Private.CoreLib/src/System/Environment.Windows.cs#L110-L122
+            //
+            // Note that we cannot use Environment.OSVersion in .NET Framework (or Core versions less than 5.0) as
+            // The implementation in those versions "lies" about Windows versions > 8.1 if there is no application manifest.
```
dscho, line 34: s/The/the/

dscho (approved):
Do we need to be worried about [these warnings](https://github.com/microsoft/Git-Credential-Manager-Core/pull/323/commits/c81bd22191c2c31915cf7af4dbd25397d46c25f8#diff-d32249dde0548f66cf569037cf9a6fbe6d56f9637c5e454c09ba052294d41c85R339) (`StorageCreationPropertiesBuilder()` being obsolete)?

docs/configuration.md
mjcheetham: Yeah. Grammar here is hard. Any suggestions to make it clearer?

docs/environment.md
mjcheetham: Yes. Take a look at aka.ms/gcmcore-env:

image

src/shared/Microsoft.Git.CredentialManager/Authentication/MicrosoftAuthentication.cs
mjcheetham: Yeah. This is unfortunate 😢

src/shared/Microsoft.Git.CredentialManager/Authentication/MicrosoftAuthentication.cs
mjcheetham: Updated.

APACGAMONDE:
Congrats M. CHEETHAM!

docs/configuration.md
vtbassmatt: Suggestion: truncate existing sentence at the `;`. Then add a second sentence:
> If both of those things are true, then the value of `credential.msauthFlow` has no effect.

docs/configuration.md
```diff
@@ -275,6 +279,27 @@ git config --global credential.msauthFlow devicecode
 
 ---
 
+### credential.msauthUseBroker
+
+Use the operating system account manager where available.
+
+Defaults to the value `true`.
```
vtbassmatt, line 286: Nit suggestion:
> Defaults to `true`.

docs/environment.md
vtbassmatt: Same sentence suggestion as the other instance.

vtbassmatt (approved):

src/shared/Microsoft.Git.CredentialManager/Authentication/MicrosoftAuthentication.cs
```diff
@@ -28,7 +32,7 @@ public enum MicrosoftAuthenticationFlowType
         Auto = 0,
         EmbeddedWebView = 1,
         SystemWebView = 2,
-        DeviceCode = 3
+        DeviceCode = 3,
```
derrickstolee, line 35: Unrelated, but OK.

src/shared/Microsoft.Git.CredentialManager/Authentication/MicrosoftAuthentication.cs
```diff
@@ -9,6 +9,10 @@
 using Microsoft.Identity.Client;
 using Microsoft.Identity.Client.Extensions.Msal;
 
+#if NETFRAMEWORK
+using Microsoft.Identity.Client.Desktop;
+#endif
```
derrickstolee, line 14: Is this going to be supported in .NET 5? Does adding this feature add more debt towards upgrading to .NET 5?

src/shared/Microsoft.Git.CredentialManager/Authentication/MicrosoftAuthentication.cs
derrickstolee: image

Reloading to ignore whitespace changes works great!

image

derrickstolee (approved):

src/shared/Microsoft.Git.CredentialManager/Authentication/MicrosoftAuthentication.cs
mjcheetham: For a possible future move to .NET 5 on Windows, we'd be using the `net5.0-windows*` TFMs, of which the MSAL library includes the WAM pieces already.

The `MS.ID.Client.Desktop` package is there to bring those Windows-only dependencies in for .NET Framework (or .NET Core 3.x and below) without adding them to the core, x-plat `MS.ID.Client` library. Several of their customers had hard blockers on introducing things like a dependency on WPF or WinRT APIs in say, a use of MSAL in an ASP.NET application.

I've been round the houses on getting us on .NET 5 on Windows, but it's fraught with problems. Primarily around: a) trimming/size of the resulting application (bigger than the rest of G4W!), b) multi-targeting and down-level Windows support (win7+8.x), c) UI 'helper' assemblies.

cc: @bgavrilMS

src/shared/Microsoft.Git.CredentialManager/Authentication/MicrosoftAuthentication.cs
mjcheetham: @bgavrilMS is this still needed for WAM support to be available?

bgavrilMS: Experimental = public preview :). We're waiting to get some more traction on this feature before moving it to GA. I will discuss directly with Matthew if we make any breaking changes. IMO the only difficulties are around supporting Personal accounts.

src/shared/Microsoft.Git.CredentialManager/Authentication/MicrosoftAuthentication.cs
bgavrilMS: Yeah, ideally WAM support would be inside MSAL on all frameworks, however the depedencies on nuget packages that call into WinRT makes it impossible. In MSAL 5 we're thinking of making WAM the default option, and thus we'd need to separate MSAL into an MSAL for public client and MSAL for confidential client.

src/shared/Microsoft.Git.CredentialManager/Authentication/MicrosoftAuthentication.cs
```diff
@@ -205,6 +234,15 @@ private async Task<IPublicClientApplication> CreatePublicClientApplicationAsync(
                 appBuilder.WithParentActivityOrWindow(() => new IntPtr(hWndInt));
             }
 
+            // On Windows 10 & .NET Framework try and use the WAM broker
+            if (enableBroker && PlatformUtils.IsWindows10())
```
bgavrilMS, line 238: Broker is also available on Win Server 2016 and Win Server 2019, but the support there is a little iffy.

src/shared/Microsoft.Git.CredentialManager/Authentication/MicrosoftAuthentication.cs
```diff
@@ -205,6 +234,15 @@ private async Task<IPublicClientApplication> CreatePublicClientApplicationAsync(
...
+            {
+#if NETFRAMEWORK
+                appBuilder.WithExperimentalFeatures();
+                appBuilder.WithWindowsBroker();
+#endif
+            }
+
```
bgavrilMS, line 245: for `StorageCreationPropertiesBuilder(cacheFileName, cacheDirectory, clientId)` just use `StorageCreationPropertiesBuilder(cacheFileName, cacheDirectory)`. The `clientId` param is only needed for situations where you want to monitor for accounts created / deleted via an event. This is used by VS to update their UI whenver an app (e.g. GCM) adds or removes an account. It's not being used by GCM.

src/shared/Microsoft.Git.CredentialManager/PlatformUtils.cs
```diff
@@ -20,6 +20,26 @@ public static PlatformInformation GetPlatformInformation()
...
 
+        public static bool IsWindows10()
+        {
+            if (!IsWindows())
+            {
+                return false;
+            }
+
+            // Implementation of version checking was taken from:
+            // https://github.com/dotnet/runtime/blob/6578f257e3be2e2144a65769706e981961f0130c/src/libraries/System.Private.CoreLib/src/System/Environment.Windows.cs#L110-L122
```
bgavrilMS, line 31: Very cool,. we use a nastier approach based on Registry for netframework.

src/shared/Microsoft.Git.CredentialManager/Authentication/MicrosoftAuthentication.cs
```diff
@@ -360,6 +398,33 @@ public HttpClient GetHttpClient()
 
         #region Auth flow capability detection
 
+        private bool CanUseBroker()
+        {
+#if NETFRAMEWORK
+            // We only support the broker on Windows 10 and require an interactive session
+            if (!Context.SessionManager.IsDesktopSession || !PlatformUtils.IsWindows10())
```
bgavrilMS, line 405: There is a helper for this in MSAL that we've added in collaboration with you: 

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/2cebdaec10f1991482ed7109a2fa34389a037299/src/client/Microsoft.Identity.Client/AppConfig/PublicClientApplicationBuilder.cs#L318

bgavrilMS (approved):

src/shared/Microsoft.Git.CredentialManager/Authentication/MicrosoftAuthentication.cs
mjcheetham: Updated to remove this use of the deprecated ctor.

src/shared/Microsoft.Git.CredentialManager/Authentication/MicrosoftAuthentication.cs
mjcheetham: Yeah, looking at this however we won't be able to use this sadly.. it's hanging off of the PCABuilder, and we need it before and after the builder exists 😢

src/shared/Microsoft.Git.CredentialManager/PlatformUtils.cs
mjcheetham: Yeah I thought there has to be a better solution given the .NET folks invented all those `OperatingSystem.IsWindowsVersionOrGreater` etc APIs, and they correctly identify Windows 10 without manifests.

src/shared/Microsoft.Git.CredentialManager/Authentication/MicrosoftAuthentication.cs
mjcheetham: NT major version 10 should also capture Windows Server SKUs based on Windows 10.

mjcheetham:
> Do we need to be worried about [these warnings](https://github.com/microsoft/Git-Credential-Manager-Core/pull/323/commits/c81bd22191c2c31915cf7af4dbd25397d46c25f8#diff-d32249dde0548f66cf569037cf9a6fbe6d56f9637c5e454c09ba052294d41c85R339) (`StorageCreationPropertiesBuilder()` being obsolete)?

@dscho I've removed the warnings by dropping that extra (not needed) argument from the constructor in question.
We updated the dependency and in the new version, that API is marked obsolete.

PR #332, opened 2021-04-28
Updating release-winget according to manifest schema changes

ldennington:
It appears that changes have been made to the winget schema since this workflow was created, so I've made the necessary updates here. Additionally, we've fixed a couple issues in the `mjcheetham/update-winget` task that had been causing this workflow to fail (see [this](https://github.com/mjcheetham/update-winget/pull/92) and [this](https://github.com/mjcheetham/update-winget/pull/94)). 

With these updates, we should be good to go to publish gcm core to winget automatically with the next release. Exciting times!

I validated this workflow locally by successfully opening a [PR](https://github.com/ldennington/winget-playground/pull/5) against my winget-playground repo and running `winget validate` to ensure the generated manifest adheres to winget's required schema.

### DIFF
--- a/.github/workflows/release-winget.yaml
+++ b/.github/workflows/release-winget.yaml
@@ -12,13 +12,13 @@ jobs:
     steps:
     - id: update-winget
       name: Update winget repository
-      branch: master
       uses: mjcheetham/update-winget@v1.1
       with:
         id: Microsoft.GitCredentialManagerCore
         token: ${{ secrets.WINGET_TOKEN }}   
         releaseAsset: gcmcore-win-x86-(.*)\.exe
         repo: ldennington/winget-playground
+        branch: master
         manifestText: |
           PackageIdentifier: {{id}}
           PackageVersion: {{version}}
--- a/.github/workflows/release-winget.yaml
+++ b/.github/workflows/release-winget.yaml
@@ -9,11 +9,12 @@ jobs:
     steps:
     - id: update winget
       name: Update winget repository
-      uses: mjcheetham/update-winget@v1.1
+      uses: mjcheetham/update-winget@v1.2
       with:
         id: Microsoft.GitCredentialManagerCore
         token: ${{ secrets.WINGET_TOKEN }}   
         releaseAsset: gcmcore-win-x86-(.*)\.exe
+        repo: ldennington/winget-playground
         manifestText: |
           PackageIdentifier: {{id}}
           PackageVersion: {{version}}
@@ -33,3 +34,5 @@ jobs:
           ManifestType: singleton
           ManifestVersion: 1.0.0
         alwaysUsePullRequest: true
+        releaseRepo: 'microsoft/Git-Credential-Manager-Core'
+        releaseTag: 'v2.0.194-beta'
--- a/.github/workflows/release-winget.yaml
+++ b/.github/workflows/release-winget.yaml
@@ -2,6 +2,9 @@ name: "release-winget"
 on:
   release:
     types: [released]
+  push:
+    branches:
+      - ldennington/update-winget-deployments
 
 jobs:
   release:
--- a/.github/workflows/release-winget.yaml
+++ b/.github/workflows/release-winget.yaml
@@ -15,7 +15,6 @@ jobs:
       uses: mjcheetham/update-winget@v1.1
       with:
         id: Microsoft.GitCredentialManagerCore
-        version: 2.0.194.40577
         token: ${{ secrets.WINGET_TOKEN }}   
         releaseAsset: gcmcore-win-x86-(.*)\.exe
         repo: ldennington/winget-playground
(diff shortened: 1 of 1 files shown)


.github/workflows/release-winget.yaml
```diff
@@ -7,26 +7,30 @@ jobs:
...
+    - id: update-winget
+      name: Update winget repository
+      uses: mjcheetham/update-winget@v1.2
       with:
-        token: ${{ secrets.WINGET_TOKEN }}
-        repo: microsoft/winget-pkgs
         id: Microsoft.GitCredentialManagerCore
+        token: ${{ secrets.WINGET_TOKEN }}   
         releaseAsset: gcmcore-win-x86-(.*)\.exe
+        repo: ldennington/winget-playground
```
derrickstolee, line 17: We'll want to change this before merging.

.github/workflows/release-winget.yaml
```diff
@@ -7,26 +7,30 @@ jobs:
...
           Publisher: Microsoft Corporation
-          AppMoniker: git-credential-manager-core
-          Homepage: https://aka.ms/gcmcore
-          Tags: "gcm, gcmcore, git, credential"
+          Moniker: git-credential-manager-core
+          PackageUrl: https://aka.ms/gcmcore
+          Tags: [ gcm, gcmcore, git, credential ]
           License: Copyright (C) Microsoft Corporation
-          Description: Secure, cross-platform Git credential storage with authentication to GitHub, Azure Repos, and other popular Git hosting services.
+          ShortDescription: Secure, cross-platform Git credential storage with authentication to GitHub, Azure Repos, and other popular Git hosting services.
```
derrickstolee, line 27: Is there now a longer `Description:` that we should populate somehow?

derrickstolee (commented):
Wow, they really broke backwards compatibility with these manifest changes. Do you have a link to the full spec? I wonder if there are some other fields we might want to populate (do `Description` or `LongDescription` exist?).

.github/workflows/release-winget.yaml
ldennington: [`Description` does still exist](https://github.com/microsoft/winget-cli/blob/cf6f8e58195bf033a1a5872f7b800866cd27a586/schemas/JSON/manifests/v1.0.0/manifest.singleton.1.0.0.json#L545-L555). However, for reasons that aren't super clear, `ShortDescription` is now required, while `Description` is not. I'm happy to take a stab at adding a longer `Description` though!

.github/workflows/release-winget.yaml
ldennington: Phew! Good catch.

ldennington:
> Wow, they really broke backwards compatibility with these manifest changes. Do you have a link to the full spec? I wonder if there are some other fields we might want to populate (do `Description` or `LongDescription` exist?).

> Yep. I was pretty shocked tbh. Here's the [link to the spec](https://github.com/microsoft/winget-cli/blob/master/schemas/JSON/manifests/v1.0.0/manifest.singleton.1.0.0.json), and some [documentation around it](https://github.com/microsoft/winget-cli/blob/master/doc/ManifestSpecv1.0.md) that I felt was helpful. I'm happy to add whichever extra fields we feel are helpful!

derrickstolee:
> > Wow, they really broke backwards compatibility with these manifest changes. Do you have a link to the full spec? I wonder if there are some other fields we might want to populate (do `Description` or `LongDescription` exist?).
> 
> Yep. I was pretty shocked tbh. Here's the [link to the spec](https://github.com/microsoft/winget-cli/blob/master/schemas/JSON/manifests/v1.0.0/manifest.singleton.1.0.0.json), and some [documentation around it](https://github.com/microsoft/winget-cli/blob/master/doc/ManifestSpecv1.0.md) that I felt was helpful. I'm happy to add whichever extra fields we feel are helpful!

Ah, there seems to be [a `Description` item](https://github.com/microsoft/winget-cli/blob/cf6f8e58195bf033a1a5872f7b800866cd27a586/schemas/JSON/manifests/v1.0.0/manifest.singleton.1.0.0.json#L445-L456) in addition to the `ShortDescription`.

.github/workflows/release-winget.yaml
ldennington: Is there anything specific we'd want to call out in a longer `Description`?

derrickstolee: I think we can wait on `Description` until we know more about where it will show up.

.github/workflows/release-winget.yaml
```diff
@@ -7,26 +7,29 @@ jobs:
...
     runs-on: ubuntu-latest
     steps:
-    - name: Update winget repository
-      uses: mjcheetham/update-winget@v1.0
+    - id: update-winget
+      name: Update winget repository
+      uses: mjcheetham/update-winget@v1.2
       with:
-        token: ${{ secrets.WINGET_TOKEN }}
-        repo: microsoft/winget-pkgs
```
derrickstolee, line 14: I think we still want `repo: microsoft/winget-pkgs`, right? Or, is that built-in to `v1.2`?

ldennington: It's the [default](https://github.com/mjcheetham/update-winget/blob/bbac470a681ce6fbd56c54b25238dfa15f463d53/action.yml#L50-L53)! However, we can keep if it's better to be explicit.

derrickstolee: Thanks! Using the default works for me.

derrickstolee (approved):

.github/workflows/release-winget.yaml
APACGAMONDE: @APACGAMONDE

APACGAMONDE:
[CGAMONDBANK](url)

PR #337, opened 2021-05-05
AzureReposHostProviderTests: fix warnings around orgName

derrickstolee:
We keep getting warnings from the actions bot saying:

  The variable 'orgName' is assigned but its value is never used

However, these variables _are_ used, just as parameters to lambdas
within LINQ methods.

By extracting this org string into a static readonly, we should be able
to avoid these warnings.


mjcheetham (approved):
:shipit: 🤫 🤖

PR #343, opened 2021-05-12
Add support for OAuth with GitHub AE

mjcheetham:
Add support for detecting GitHub AE (GHAE) and enabling the OAuth authentication mode with the GitHub provider.

We assume that all instances of GHAE have the GCM OAuth application deployed.

https://docs.github.com/en/github-ae@latest/admin/overview/about-github-ae

### DIFF
--- a/src/shared/GitHub/GitHubConstants.cs
+++ b/src/shared/GitHub/GitHubConstants.cs
@@ -33,6 +33,11 @@ namespace GitHub
         public static readonly Version MinimumEnterpriseOAuthVersion = new Version("99.99.99");
 
         /// <summary>
+        /// The version string returned from the meta API endpoint for GitHub AE instances.
+        /// </summary>
+        public const string GitHubAeVersionString = "GitHub AE";
+
+        /// <summary>
         /// Supported authentication modes for GitHub.com.
         /// </summary>
         /// <remarks>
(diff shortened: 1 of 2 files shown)


src/shared/GitHub/GitHubConstants.cs
```diff
@@ -32,6 +32,11 @@ public static class GitHubConstants
         // TODO: update this with a real version number once the GCM OAuth application has been deployed to GHE
         public static readonly Version MinimumEnterpriseOAuthVersion = new Version("99.99.99");
```
mjcheetham, line 33: Once we know which version of GHES has the OAuth app deployed, we can update this constant too.

vtbassmatt: Will this ever cause confusion since "Enterprise" is a part of the name of three different GitHub products? `MinimumOnPremOAuthVersion` maybe? (This may be pointless bikeshedding on my part - the question is genuine but not particularly important.)

mjcheetham: Feel free to submit the rename PR to that OnPrem suggestion ;)

derrickstolee (approved):

src/shared/GitHub/GitHubConstants.cs
vtbassmatt: Ha 😁

vtbassmatt: Whoops, can't resolve, since you had left this as a note to yourself.

PR #344, opened 2021-05-13
Add WAM workaround for elevated processes

mjcheetham:
Add a workaround to a broker bug whereby the account control does not appear when running from an elevated process.

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2560

The underlying issue is to do with COM and the OS account control not being able to call-back in to the elevated process.

The workaround is to set the process COM security to "none" iif we are on Windows 10, the process is elevated, and the user hasn't disabled the broker.

It is possible the call to `CoInitializeSecurity` may fail, as this can only be called once in the lifetime of a process, and must be called before any COM interactions occur. The CLR may perform some COM interop before we even get to the Main method(!)

We try our best here and call the `CoInitializeSecurity` function as soon as we reasonably can in the lifetime of our process.

### DIFF
--- a/src/shared/Git-Credential-Manager/Program.cs
+++ b/src/shared/Git-Credential-Manager/Program.cs
@@ -20,7 +20,11 @@ namespace Microsoft.Git.CredentialManager
             {
                 // Workaround for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2560
                 if (MicrosoftAuthentication.CanUseBroker(context))
-                    try { MicrosoftAuthentication.InitializeBroker(); }
+                {
+                    try
+                    {
+                        MicrosoftAuthentication.InitializeBroker();
+                    }
                     catch (Exception ex)
                     {
                         context.Streams.Error.WriteLine(
@@ -28,6 +32,7 @@ namespace Microsoft.Git.CredentialManager
                             Environment.NewLine, ex.Message
                         );
                     }
+                }
 
                 // Register all supported host providers at the normal priority.
                 // The generic provider should never win against a more specific one, so register it with low priority.
--- a/src/shared/Git-Credential-Manager/Program.cs
+++ b/src/shared/Git-Credential-Manager/Program.cs
@@ -3,10 +3,10 @@
 using System;
 using System.IO;
 using System.Reflection;
-using System.Runtime.InteropServices;
 using Atlassian.Bitbucket;
 using GitHub;
 using Microsoft.AzureRepos;
+using Microsoft.Git.CredentialManager.Authentication;
 
 namespace Microsoft.Git.CredentialManager
 {
@@ -18,6 +18,17 @@ namespace Microsoft.Git.CredentialManager
             using (var context = new CommandContext(appPath))
             using (var app = new Application(context))
             {
+                // Workaround for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2560
+                if (MicrosoftAuthentication.CanUseBroker(context))
+                    try { MicrosoftAuthentication.InitializeBroker(); }
+                    catch (Exception ex)
+                    {
+                        context.Streams.Error.WriteLine(
+                            "warning: broker initialization failed{0}{1}",
+                            Environment.NewLine, ex.Message
+                        );
+                    }
+
                 // Register all supported host providers at the normal priority.
                 // The generic provider should never win against a more specific one, so register it with low priority.
                 app.RegisterProvider(new AzureReposHostProvider(context), HostProviderPriority.Normal);
--- a/src/shared/Microsoft.Git.CredentialManager/Authentication/MicrosoftAuthentication.cs
+++ b/src/shared/Microsoft.Git.CredentialManager/Authentication/MicrosoftAuthentication.cs
@@ -49,15 +49,24 @@ namespace Microsoft.Git.CredentialManager.Authentication
 
         public static void InitializeBroker()
         {
-            if (IsBrokerInitialized) return;
+            if (IsBrokerInitialized)
+            {
+                return;
+            }
 
             IsBrokerInitialized = true;
 
             // Broker is only supported on Windows 10
-            if (!PlatformUtils.IsWindows10()) return;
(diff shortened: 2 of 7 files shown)


bgavrilMS (approved):

src/shared/Microsoft.Git.CredentialManager/Authentication/MicrosoftAuthentication.cs
```diff
@@ -44,19 +43,65 @@ public class MicrosoftAuthentication : AuthenticationBase, IMicrosoftAuthenticat
...
+
+        public static void InitializeBroker()
+        {
+            if (IsBrokerInitialized) return;
+
+            // Broker is only supported on Windows 10
+            if (!PlatformUtils.IsWindows10()) return;
+
+            // Nothing to do when not an elevated user
+            if (!PlatformUtils.IsElevatedUser()) return;
```
bgavrilMS, line 58: @pmaytak - I think we should have this (lines 58 - 83) as a helper in MSAL

src/shared/Git-Credential-Manager/Program.cs
```diff
@@ -18,6 +18,17 @@ public static void Main(string[] args)
             using (var context = new CommandContext(appPath))
             using (var app = new Application(context))
             {
+                // Workaround for https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2560
+                if (MicrosoftAuthentication.CanUseBroker(context))
+                    try { MicrosoftAuthentication.InitializeBroker(); }
```
derrickstolee, line 23: nit: expand braces into multiple lines.

src/shared/Microsoft.Git.CredentialManager/Authentication/MicrosoftAuthentication.cs
```diff
@@ -44,19 +43,65 @@ public class MicrosoftAuthentication : AuthenticationBase, IMicrosoftAuthenticat
...
+        {
+            if (IsBrokerInitialized) return;
+
+            IsBrokerInitialized = true;
+
+            // Broker is only supported on Windows 10
+            if (!PlatformUtils.IsWindows10()) return;
+
+            // Nothing to do when not an elevated user
+            if (!PlatformUtils.IsElevatedUser()) return;
```
derrickstolee, line 60: Your time working in the Git codebase is showing.

```suggestion
            if (IsBrokerInitialized)
            {
                        return;
            }

            IsBrokerInitialized = true;

            // Broker is only supported on Windows 10
            if (!PlatformUtils.IsWindows10())
            {
                        return;
            }

            // Nothing to do when not an elevated user
            if (!PlatformUtils.IsElevatedUser())
            {
                        return;
            }
```

src/shared/Microsoft.Git.CredentialManager/PlatformUtils.cs
```diff
@@ -136,6 +137,24 @@ public static void EnsurePosix()
...
 
+        public static bool IsElevatedUser()
+        {
+            if (IsWindows())
+            {
+#if NETFRAMEWORK
+                var identity = System.Security.Principal.WindowsIdentity.GetCurrent();
+                var principal = new System.Security.Principal.WindowsPrincipal(identity);
+                return principal.IsInRole(System.Security.Principal.WindowsBuiltInRole.Administrator);
+#endif
```
derrickstolee, line 148: Just a question: is this available in the new, "unified" .NET 5? How much longer are we forced to use the .NET Framework?

derrickstolee (approved):

src/shared/Microsoft.Git.CredentialManager/Authentication/MicrosoftAuthentication.cs
mjcheetham: There are select parts in the GCM codebase that I've used this style of `if` without braces, and it's usually exclusively been for `if (condition) return` early exits.

But you're right. I probably should keep the codebase "clean" :)

derrickstolee: I guess if it is an existing pattern in this codebase, then that is probably fine. Any C# linter would probably complain, though.

src/shared/Microsoft.Git.CredentialManager/PlatformUtils.cs
mjcheetham: The API surface is in .NET Framwork and .NET 5 (and even in .NET Core 3.x and 1.x), but just not .NET _Standard_ 2.0. If we unified on .NET 5 then this `#if` could become just a runtime `if (PlatformUtils.IsWindows())`.

> How much longer are we forced to use the .NET Framework?

Sadly this looks to be much longer. There is one final problem with moving from .NET Framework to .NET 5 on Windows: File size and file count bloat.

Going from < 3MB and a dozen files.. to 100MB+ and 80+ files(!) which @dscho has shown some shock and horror at for bundling with Git for Windows (and rightly so!). Even with assembly trimming and linking, we're looking at maybe 40-50MB and then the last time I tried it, 1/2 of the UI features broke 😢

derrickstolee: Thanks for giving me some context. At minimum, this change doesn't add to the friction of a future switch.

src/shared/Microsoft.Git.CredentialManager/Authentication/MicrosoftAuthentication.cs
mjcheetham: `grep -rn "if\s(.*;$" src` returns 37 matches (including the ones added in this PR)

src/shared/Microsoft.Git.CredentialManager/PlatformUtils.cs
mjcheetham: Yeah, in most cases the `#if`s are working around differences between what is and isn't in .NET Standard.
Moving to any concrete runtime is mostly just _deleting_ these conditions.

PR #374, opened 2021-06-21
Fix app entry path computation on Windows

mjcheetham:
The way we located the application (GCM) path was broken on .NET Framework (it worked OK for .NET Core/5.0+).

### DIFF
--- a/src/shared/Microsoft.Git.CredentialManager/ApplicationBase.cs
+++ b/src/shared/Microsoft.Git.CredentialManager/ApplicationBase.cs
@@ -86,24 +86,31 @@ namespace Microsoft.Git.CredentialManager
 
         public static string GetEntryApplicationPath()
         {
-            // Assembly::Location always returns an empty string if the application was published as a single file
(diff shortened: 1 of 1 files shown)


src/shared/Microsoft.Git.CredentialManager/ApplicationBase.cs
```diff
@@ -86,18 +86,25 @@ public Task<int> RunAsync(string[] args)
 
         public static string GetEntryApplicationPath()
         {
-            // Assembly::Location always returns an empty string if the application was published as a single file
+            // Single file publishing does not exist with .NET Framework so
+            // we can just use reflection to get the entry assembly path.
+#if NETFRAMEWORK
+            return Assembly.GetEntryAssembly().Location;
+#endif
```
derrickstolee, line 93: Will this cause "unreachable code" errors in .NET Framework? Should we use an `#else`/`#endif` block to contain the rest of the code?

derrickstolee (approved):

PR #373, opened 2021-06-18
Issue-267 Add fingerprint header, to autodetect Bitbucket DC instances.

mminns:
Addressed https://github.com/microsoft/Git-Credential-Manager-Core/issues/267

Uses the Atlassian specific HTTP Header "X-AREQUESTID" to fingerprint a Bitbucket on-premise instance.

Added a simple test, I notice there are currently no tests for Bitbucket, so I might look at that more generally in a separate PR.

It's been a while since I contributed, nice to be back. :)

PR #379, opened 2021-06-22
Avoid extra restore in macOS installer creation (fix CI hangs)

mjcheetham:
Avoid an extra `dotnet restore` during the `dotnet publish` commands in `layout.sh` when building the macOS installer.
For some reason, the CI machines (and sometimes locally) hang during this step with:

```
Building Installer.Mac
Copying uninstall script...
Microsoft (R) Build Engine version 16.10.0-preview-21181-07+073022eb4 for .NET
Copyright (C) Microsoft Corporation. All rights reserved.

  Determining projects to restore...

```

Avoiding the restore during publish should be OK since the hosting `Installer.Mac.csproj` project will have already built (and restored) packages for the dependent projects.

Issues noticed for example in #373

### DIFF
--- a/src/osx/Installer.Mac/layout.sh
+++ b/src/osx/Installer.Mac/layout.sh
@@ -74,6 +74,7 @@ cp "$INSTALLER_SRC/uninstall.sh" "$PAYLOAD" || exit 1
 # Publish core application executables
 echo "Publishing core application..."
 dotnet publish "$GCM_SRC" \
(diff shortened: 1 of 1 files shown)


derrickstolee (approved):

mjcheetham:
`/m:1` seems to fix this.. probably some deadlocking situation where MSBuild is waiting on MSBuild 🙄

PR #398, opened 2021-07-20
Update linuxcredstores.md

hickford:
Clarify that is necessary to run `git-credential-manager-core configure`.


mjcheetham (approved):
Thank you! 🎉

PR #401, opened 2021-07-20
Issue 329 force bitbucket oauth

mminns:
The following options allow a user to force the GCMC to use Basic Auth and/or OAuth for interactions with bitbucket.org

By default a combination of Basic Auth and OAuth is used.

See the revised docs in this PR on how to configure.

**Testing**

Assuming there are no existing stored credentials and the git 'host' is identified as bitbucket.org then

AuthModes = "" or AuthModes = "basic, oauth"

- using Basic Auth a username/password is requested and used to retrieve the user profile 
  - if this works the username/password are passed back to git.
  - if this fails with a 403 response, this indicates 2FA is on and OAuth is required
    - The user is prompted to run the OAuth flow
      - If this is successful the access_token is passed back to git

AuthModes = "basic"

- using Basic Auth a username/password is requested and used to retrieve the user profile 
  - if this works the username/password are passed back to git.
- the OAuth flow is never triggered 

AuthModes = "oauth"

- The Basic Auth step is skipped
- The user is prompted to run the OAuth flow
      - If this is successful the access_token is passed back to git
      
     
      
  **TBC testing against bitbucket server instances.**

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -148,6 +148,32 @@ git config --global credential.httpsProxy http://john.doe:password@proxy.contoso
 
 ---
 
+### credential.bitbucketAuthModes
+
+Override the available authentication modes presented during Bitbucket authentication.
+If this option is not set, then the available authentication modes will be automatically detected.
+
+
+**Note:** This setting only applies to Bitbucket.org not Server or DC instances.
+
+**Note:** This setting supports multiple values separated by commas.
+
+Value|Authentication Mode
+-|-
+_(unset)_|Automatically detect modes
+`oauth`|OAuth-based authentication
+`basic`|Basic/PAT-based authentication
+
+#### Example
+
+```shell
+git config --global credential.bitbucketAuthModes "oauth,basic"
+```
+
+**Also see: [GCM_BITBUCKET_AUTHMODES](environment.md#GCM_BITBUCKET_AUTHMODES)**
+
+---
+
 ### credential.gitHubAuthModes
 
 Override the available authentication modes presented during GitHub authentication.
--- a/docs/environment.md
+++ b/docs/environment.md
@@ -283,6 +283,37 @@ export GCM_HTTP_PROXY=http://john.doe:password@proxy.contoso.com
 
 ---
 
+### GCM_BITBUCKET_AUTHMODES
+
+Override the available authentication modes presented during Bitbucket authentication.
+If this option is not set, then the available authentication modes will be automatically detected.
+
+**Note:** This setting only applies to Bitbucket.org not Server or DC instances.
+
+**Note:** This setting supports multiple values separated by commas.
+
+Value|Authentication Mode
+-|-
+_(unset)_|Automatically detect modes
+`oauth`|OAuth-based authentication
+`basic`|Basic/PAT-based authentication
+
+##### Windows
+
+```batch
+SET GCM_BITBUCKET_AUTHMODES="oauth,basic"
+```
+
+##### macOS/Linux
+
+```bash
+export GCM_BITBUCKET_AUTHMODES="oauth,basic"
+```
+
+**Also see: [credential.bitbucketAuthModes](configuration.md#credential.bitbucketAuthModes)**
+
+---
+
 ### GCM_GITHUB_AUTHMODES
 
 Override the available authentication modes presented during GitHub authentication.
@@ -308,7 +339,7 @@ SET GCM_GITHUB_AUTHMODES="oauth,basic"
 export GCM_GITHUB_AUTHMODES="oauth,basic"
 ```
 
-**Also see: [credential.gitHubAuthModes](configuration.md#credentialgitHubAuthModes)**
+**Also see: [credential.gitHubAuthModes](configuration.md#credential.gitHubAuthModes)**
 
 ---
 
--- a/src/shared/Atlassian.Bitbucket/BitbucketAuthentication.cs
+++ b/src/shared/Atlassian.Bitbucket/BitbucketAuthentication.cs
@@ -12,6 +12,17 @@ using Microsoft.Git.CredentialManager.Authentication.OAuth;
 
 namespace Atlassian.Bitbucket
 {
+
+    [Flags]
+    public enum AuthenticationModes
+    {
+        None = 0,
+        Basic = 1,
+        OAuth = 1 << 1,
+        //Pat = 1 << 2,
+
+        All = Basic | OAuth //| Pat
+    }
     public interface IBitbucketAuthentication : IDisposable
     {
         Task<ICredential> GetBasicCredentialsAsync(Uri targetUri, string userName);
--- a/src/shared/Atlassian.Bitbucket/BitbucketConstants.cs
+++ b/src/shared/Atlassian.Bitbucket/BitbucketConstants.cs
@@ -34,6 +34,7 @@ namespace Atlassian.Bitbucket
             public const string DevOAuthClientId = "GCM_DEV_BITBUCKET_CLIENTID";
             public const string DevOAuthClientSecret = "GCM_DEV_BITBUCKET_CLIENTSECRET";
             public const string DevOAuthRedirectUri = "GCM_DEV_BITBUCKET_REDIRECTURI";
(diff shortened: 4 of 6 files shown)


mminns:
My local test all seems to work OK.

Very happy if anyone else tests it out too.

src/shared/Atlassian.Bitbucket/BitbucketHostProvider.cs
```diff
@@ -197,6 +212,47 @@ public async Task<ICredential> GetCredentialAsync(InputArguments input)
...
+        private static bool SupportsBasicAuth(AuthenticationModes authModes)
+        {
+            return (authModes & AuthenticationModes.Basic) != 0;
+        }
+
+        public async Task<AuthenticationModes> GetSupportedAuthenticationModesAsync(Uri targetUri)
+        {
+            if(!IsBitbucketOrg(targetUri))
+            {
+                // Bitbucket.org should use Basic only
```
mjcheetham, line 229: Don't you mean Bitbucket server here?

src/shared/Atlassian.Bitbucket/BitbucketHostProvider.cs
```diff
@@ -197,6 +212,47 @@ public async Task<ICredential> GetCredentialAsync(InputArguments input)
...
+        {
+            return (authModes & AuthenticationModes.OAuth) != 0;
+        }
+
+        private static bool SupportsBasicAuth(AuthenticationModes authModes)
+        {
+            return (authModes & AuthenticationModes.Basic) != 0;
+        }
+
+        public async Task<AuthenticationModes> GetSupportedAuthenticationModesAsync(Uri targetUri)
```
mjcheetham, line 225: Don't make this an `async` method until/unless it will make a network call.

src/shared/Atlassian.Bitbucket/BitbucketHostProvider.cs
```diff
@@ -176,6 +167,30 @@ public async Task<ICredential> GetCredentialAsync(InputArguments input)
...
+
+            // Store the refreshed RT
+            _context.Trace.WriteLine("Storing new refresh token...");
+            _context.CredentialStore.AddOrUpdate(refreshTokenService, input.UserName, refreshResult.RefreshToken);
+
+            // Return new access token
+            return new GitCredential(refreshUserName, refreshResult.AccessToken);
+        }
+
+        private async Task<ICredential> GetOAuthCredentialsViaManualFlow(Uri targetUri, string refreshTokenService)
```
mjcheetham, line 192: I'd rename this method to `GetOAuthCredentialInteractive` rather than "manual".

We should maybe also throw an exception if interaction has been disable here via `GCM_INTERACTIVE` or `credential.interactive` being `false`...

https://github.com/microsoft/Git-Credential-Manager-Core/blob/7fe5379e11a4f9296e6ced4661d72434fa358545/src/shared/Microsoft.Git.CredentialManager/Authentication/AuthenticationBase.cs#L80

At this point I also think it'll probably be worth extracting some of the "auth" parts from the provider to a `BitbucketAuthentication` class/type, like we do [for GitHub](https://github.com/microsoft/Git-Credential-Manager-Core/blob/7fe5379e11a4f9296e6ced4661d72434fa358545/src/shared/GitHub/GitHubAuthentication.cs).

src/shared/Atlassian.Bitbucket/BitbucketConstants.cs
```diff
@@ -44,7 +45,18 @@ public static class Credential
...
                 public const string DevOAuthClientSecret = "bitbucketDevClientSecret";
                 public const string DevOAuthRedirectUri = "bitbucketDevRedirectUri";
+                public const string AuthenticationModes = "bitbucketAuthModes";
             }
         }
+
+        /// <summary>
+        /// Supported authentication modes for Bitbucket.org
+        /// </summary>
+        public const AuthenticationModes DotOrgAuthenticationModes = AuthenticationModes.Basic | AuthenticationModes.OAuth; //| AuthenticationModes.Pat;
```
mjcheetham, line 55: If we're not adding an explicit `Pat` option (and just use the basic password field to paste a PAT in, which is perfectly OK), then please drop the commented out options. 😁

src/shared/Atlassian.Bitbucket/BitbucketAuthentication.cs
```diff
@@ -12,6 +12,17 @@
...
 namespace Atlassian.Bitbucket
 {
+
+    [Flags]
+    public enum AuthenticationModes
+    {
+        None = 0,
+        Basic = 1,
+        OAuth = 1 << 1,
+        //Pat = 1 << 2,
```
mjcheetham, line 22: Drop commented out/unused options.

docs/environment.md
```diff
@@ -308,7 +339,7 @@ SET GCM_GITHUB_AUTHMODES="oauth,basic"
 export GCM_GITHUB_AUTHMODES="oauth,basic"
 ```
 
-**Also see: [credential.gitHubAuthModes](configuration.md#credentialgitHubAuthModes)**
+**Also see: [credential.gitHubAuthModes](configuration.md#credential.gitHubAuthModes)**
```
mjcheetham, line 342: GitHub link fragments ignore periods `.` in headings, so this change breaks the link.

docs/environment.md
```diff
@@ -283,6 +283,37 @@ export GCM_HTTP_PROXY=http://john.doe:password@proxy.contoso.com
...
+SET GCM_BITBUCKET_AUTHMODES="oauth,basic"
+```
+
+##### macOS/Linux
+
+```bash
+export GCM_BITBUCKET_AUTHMODES="oauth,basic"
+```
+
+**Also see: [credential.bitbucketAuthModes](configuration.md#credential.bitbucketAuthModes)**
```
mjcheetham, line 313: Check this link works; should drop the `.`.

docs/environment.md
```diff
@@ -283,6 +283,37 @@ export GCM_HTTP_PROXY=http://john.doe:password@proxy.contoso.com
 
 ---
 
+### GCM_BITBUCKET_AUTHMODES
+
+Override the available authentication modes presented during Bitbucket authentication.
+If this option is not set, then the available authentication modes will be automatically detected.
+
+**Note:** This setting only applies to Bitbucket.org not Server or DC instances.
```
mjcheetham, line 291:
```suggestion
**Note:** This setting only applies to Bitbucket.org, and not Server or DC instances.
```

docs/configuration.md
```diff
@@ -148,6 +148,32 @@ git config --global credential.httpsProxy http://john.doe:password@proxy.contoso
 
 ---
 
+### credential.bitbucketAuthModes
+
+Override the available authentication modes presented during Bitbucket authentication.
+If this option is not set, then the available authentication modes will be automatically detected.
+
+
+**Note:** This setting only applies to Bitbucket.org not Server or DC instances.
```
mjcheetham, line 157:
```suggestion
**Note:** This setting only applies to Bitbucket.org, and not Server or DC instances.
```

mjcheetham (changes requested):
Thanks for the effort here! I have a few asks to rename some methods and extract some of the `GetCreds*` to a new `BitbucketAuthentication` class.

mminns:
Nio problem. Thanks for the feedback I'll address it in the next few days.

src/shared/Atlassian.Bitbucket/BitbucketHostProvider.cs
mjcheetham: > There is already a BitbucketAuthentication class [..]

Ahh so there is! Doh! Sorry about that. 🤦 

Yes, I think all that is needed is to add a `ThrowIfUserInteractionDisabled()` call right at the start of the `BitbucketAuthentication.CreateOAuthCredentialsAsync()` method. The method will be the one launching the web browser to do OAuth, which is an interaction.

You can drop the extra `void ThrowIfUserInteractionDisabled();` method signature on the `IBitbucketAuthentication` interface and the call in `BitbucketHostProvider.GetOAuthCredentialsInteractive()`.

Sorry again for making you add that in the first place.. I should learn to look outside of the files diff in GitHub!

Jan11532 (commented):
> 

> ``

PR #404, opened 2021-08-05
Add ability to specify default settings values from the registry on Windows

mjcheetham:
Introduce the concept of another layer of settings (lowest precedence; default values only) below the existing environment variable and Git configuration file mechanisms.

Implement on Windows using the Registry, under key: `HKLM\SOFTWARE\GitCredentialManager\Configuration`

The primary target for this feature are enterprise system administrators who wish to set defaults for GCM via MDM technologies such as Group Policy on Windows.

**Note:** We believe the user should _always_ be at liberty to configure Git and GCM exactly as they wish. By preferring environment variables and Git configuration files over OS store values, these only act as _default values_ that can always be overridden by the user in the usual ways.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -3,6 +3,7 @@
 [Git Credential Manager Core](usage.md) works out of the box for most users.
 
 Git Credential Manager Core (GCM Core) can be configured using Git's configuration files, and follows all of the same rules Git does when consuming the files.
+
 Global configuration settings override system configuration settings, and local configuration settings override global settings; and because the configuration details exist within Git's configuration files you can use Git's `git config` utility to set, unset, and alter the setting values. All of GCM Core's configuration settings begin with the term `credential`.
 
 GCM Core honors several levels of settings, in addition to the standard local \> global \> system tiering Git uses.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -8,7 +8,7 @@ Global configuration settings override system configuration settings, and local 
 GCM Core honors several levels of settings, in addition to the standard local \> global \> system tiering Git uses.
 URL-specific settings or overrides can be applied to any value in the `credential` namespace with the syntax below.
 
-Additionally, GCM Core respects several GCM-specific [environment variables](environment.md) **which take precedence over configuration options.**
+Additionally, GCM Core respects several GCM-specific [environment variables](environment.md) **which take precedence over configuration options**. System administrators may also configure [default values](enterprise-config.md) for many settings used by GCM Core.
 
 GCM Core will only be used by Git if it is installed and configured. Use `git config --global credential.helper manager-core` to assign GCM Core as your credential helper. Use `git config credential.helper` to see the current configuration.
 
--- a/docs/enterprise-config.md
+++ b/docs/enterprise-config.md
@@ -8,34 +8,45 @@ different mechanisms. In order of preference, those mechanisms are:
    1. Repository/local configuration (`.git/config`)
    2. User/global configuration (`$HOME/.gitconfig` or `%HOME%\.gitconfig`)
    3. Installation/system configuration (`etc/gitconfig`)
-3. Operating system specific configuration store
+3. Enterprise system administrator defaults
+4. Compiled default values
 
 This model largely matches what Git itself supports, namely environment
 variables that take precedence over Git configuration files.
 
-The addition of the operating system specific configuration store enables system
-administrators to configure defaults for many settings that may be required in
-an enterprise or corporate setting.
+The addition of the enterprise system administrator defaults enables those
+administrators to configure many GCM settings using familiar MDM tooling, rather
+than having to modify the Git installation configuration files.
 
----
+### User Freedom
 
-**Important:** We believe the user should _always_ be at liberty to configure
+We believe the user should _always_ be at liberty to configure
 Git and GCM exactly as they wish. By prefering environment variables and Git
-configuration files over OS store values, these only act as _default values_
+configuration files over system admin values, these only act as _default values_
 that can always be overriden by the user in the usual ways.
-
----
 
 ## Windows
 
 Default setting values come from the Windows Registry, specifically the
-`HKEY_LOCAL_MACHINE` hive under the following key:
+following keys:
+
+**32-bit Windows**
 
 ```text
-SOFTWARE\GitCredentialManager\Configuration
+HKEY_LOCAL_MACHINE\SOFTWARE\GitCredentialManager\Configuration
 ```
 
-By using the Windows Registry system administrators can use Group Policy to
+**64-bit Windows**
+
+```text
+HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\GitCredentialManager\Configuration
+```
+
+> GCM Core is a 32-bit executable on Windows. When running on a 64-bit
+installation of Windows registry access is transparently redirected to the
+`WOW6432Node` node.
+
+By using the Windows Registry, system administrators can use Group Policy to
 easily set defaults for GCM Core's settings.
 
 The names and possible values of all settings under this key are the same as
@@ -44,13 +55,6 @@ those of the [Git configuration](configuration.md) settings.
 The type of each registry key can be either `REG_SZ` (string) or `REG_DWORD`
 (integer).
 
-**Note:** GCM Core is a 32-bit executable on Windows. When running on a 64-bit
-installation of Windows registry access is transparently redirected to the
-`WOW6432Node`. This means the actual registry key on 64-bit Windows is:
-
-```text
-HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\GitCredentialManager\Configuration
-```
 
 ## macOS/Linux
 
--- a/docs/enterprise-config.md
+++ b/docs/enterprise-config.md
@@ -1,0 +1,57 @@
+# Enterprise configuration defaults
+
+Git Credential Manager Core (GCM Core) can be configured using multiple
+different mechanisms. In order of preference, those mechanisms are:
+
+1. [Environment variables](environment.md)
+2. [Standard Git configuration files](configuration.md)
+   1. Repository/local configuration (`.git/config`)
+   2. User/global configuration (`$HOME/.gitconfig` or `%HOME%\.gitconfig`)
+   3. Installation/system configuration (`etc/gitconfig`)
+3. Operating system specific configuration store
+
+This model largely matches what Git itself supports, namely environment
+variables that take precedence over Git configuration files.
+
+The addition of the operating system specific configuration store enables system
+administrators to configure defaults for many settings that may be required in
+an enterprise or corporate setting.
+
+---
+
+**Important:** We believe the user should _always_ be at liberty to configure
+Git and GCM exactly as they wish. By prefering environment variables and Git
+configuration files over OS store values, these only act as _default values_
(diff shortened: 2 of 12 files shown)


vtbassmatt:
This is cool. My only question is, how certain are we that this is the correct area of the registry? (Genuine question -- you've probably done your homework, I just don't know how MDM systems expect to configure the registry.)

Also, as a future thought -- when we build some diagnostics, it would be nice to emit configuration and where those config values came from. Env vars are obvious, Git config can be slightly obscure, but the registry is really hidden from the average user's view (even a power user). Doesn't block this work _at all_, thought.

vtbassmatt (approved):
Definitely appreciate the additional documentation, as well!

mjcheetham:
@vtbassmatt:

> Also, as a future thought -- when we build some diagnostics, it would be nice to emit configuration and where those config values came from

This was my immediate thought as well! :wink: I'll add a read out of all keys in the upcoming `gcm diagnose` command.

mjcheetham:
> My only question is, how certain are we that this is the correct area of the registry?

Group Policy and/or Intune can set any arbitrary registry value from what I understand.

The Local Machine hive feels like the correct place based on MDM policies applying to the entire machine (not just the current user).

Since the settings pertain to GCM, the settings should logically exist under a GCM namespace in the registry. Applications should place their root key under the `SOFTWARE` key (under a company subkey if applicable, like `SOFTWARE\Microsoft\Edge`).

docs/enterprise-config.md
```diff
@@ -0,0 +1,57 @@
...
+
+Git Credential Manager Core (GCM Core) can be configured using multiple
+different mechanisms. In order of preference, those mechanisms are:
+
+1. [Environment variables](environment.md)
+2. [Standard Git configuration files](configuration.md)
+   1. Repository/local configuration (`.git/config`)
+   2. User/global configuration (`$HOME/.gitconfig` or `%HOME%\.gitconfig`)
+   3. Installation/system configuration (`etc/gitconfig`)
+3. Operating system specific configuration store
```
jeffhostetler, line 11: maybe "Operating system specific enterprise configuration store"
to tie it back to the title of the section/document ??

I guess I'm looking for a clarifying term to distinguish [3] from [2.3].

docs/enterprise-config.md
```diff
@@ -0,0 +1,57 @@
...
+
+This model largely matches what Git itself supports, namely environment
+variables that take precedence over Git configuration files.
+
+The addition of the operating system specific configuration store enables system
```
jeffhostetler, line 16: and again here.  (maybe)

src/shared/Microsoft.Git.CredentialManager/Interop/Windows/WindowsSettings.cs
```diff
@@ -0,0 +1,55 @@
...
+        }
+
+        protected override bool TryGetExternalDefault(string section, string property, out string value)
+        {
+            value = null;
+
+#if NETFRAMEWORK
+            // Check for machine (HKLM) registry keys that match the Git configuration name.
+            // These can be set by system administrators via Group Policy, so make useful defaults.
+            using (Win32.RegistryKey configKey = Win32.Registry.LocalMachine.OpenSubKey(Constants.WindowsRegistry.HKConfigurationPath))
```
vdye, line 29: Why is this limited to `NETFRAMEWORK` projects? It looks like both `RegistryKey` and `Registry` exist for .NET 5.

docs/enterprise-config.md
```diff
@@ -0,0 +1,57 @@
...
+administrators to configure defaults for many settings that may be required in
+an enterprise or corporate setting.
```
jeffhostetler, line 18: maybe add something like "[3] may allow administrators to use normal enterprise tooling instead of [2.3] to simplify their configuration tasks."

I guess I'm looking for a motivation for adding [3] given that we already have [2.3].

vdye (approved):
Looks great! Just had one question, mostly for my own understanding.

jeffhostetler (approved):

docs/configuration.md
```diff
@@ -8,7 +8,7 @@ Global configuration settings override system configuration settings, and local
 GCM Core honors several levels of settings, in addition to the standard local \> global \> system tiering Git uses.
 URL-specific settings or overrides can be applied to any value in the `credential` namespace with the syntax below.
 
-Additionally, GCM Core respects several GCM-specific [environment variables](environment.md) **which take precedence over configuration options.**
+Additionally, GCM Core respects several GCM-specific [environment variables](environment.md) **which take precedence over configuration options**. System administrators may also configure [default values](enterprise-config.md) for many settings used by GCM Core.
```
derrickstolee, line 11: Should we make it _really_ explicit here that the defaults are overridden by config values and then environment variables?

```suggestion
GCM Core has multiple sources that provide customization of configured values. The following sources are used, in priority order:

1. GCM-specific [environment variables](environment.md)
2. Values available in `git config`. Git prioritizes local config settings over global and system config settings.
3. [System administrator-set default values](enterprise-config.md).
4. Compiled defaults in the GCM codebase.
```

Not sure this is better, but it at least makes the order very clear. Also, the two preceding paragraphs would want edits to work more closely with this format.

docs/enterprise-config.md
```diff
@@ -0,0 +1,57 @@
...
+
+---
+
+**Important:** We believe the user should _always_ be at liberty to configure
+Git and GCM exactly as they wish. By prefering environment variables and Git
+configuration files over OS store values, these only act as _default values_
+that can always be overriden by the user in the usual ways.
```
derrickstolee, line 25: I appreciate this caveat. Could we stick it in a `###`-header so we can link directly to it when someone complains either that:

1. their system admin gets too much control over their machine (they don't), or
2. the system admin complains that users are overriding their settings.

docs/enterprise-config.md
```diff
@@ -0,0 +1,57 @@
...
+The type of each registry key can be either `REG_SZ` (string) or `REG_DWORD`
+(integer).
+
+**Note:** GCM Core is a 32-bit executable on Windows. When running on a 64-bit
+installation of Windows registry access is transparently redirected to the
+`WOW6432Node`. This means the actual registry key on 64-bit Windows is:
+
+```text
+HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\GitCredentialManager\Configuration
+```
```
derrickstolee, line 53: Should we assume 64-bit windows and list only this configuration setting?

Is there any chance that we upgrade GCM Core to 64-bit in the future? Should system admins set both keys?

docs/environment.md
```diff
@@ -2,7 +2,7 @@
 
 [Git Credential Manager Core](usage.md) works out of the box for most users. Configuration options are available to customize or tweak behavior.
 
-Git Credential Manager Core (GCM Core) can be configured using environment variables. **Environment variables take precedence over [configuration](configuration.md) options.**
+Git Credential Manager Core (GCM Core) can be configured using environment variables. **Environment variables take precedence over [configuration](configuration.md) options**. System administrators may also configure [default values](enterprise-config.md) for many settings used by GCM Core.
```
derrickstolee, line 5: Perhaps separate this from the earlier discussion by just saying that environment variables have highest priority?

```suggestion
Git Credential Manager Core (GCM Core) can be configured using environment variables. **Environment variables take precedence over any other configuration options.**
```

src/shared/Microsoft.Git.CredentialManager/Interop/Windows/WindowsSettings.cs
mjcheetham: Although the `Registry*` types exist in .NET 5 (/.NET Core), they exist in a separate assembly/NuGet package that must be referenced to use. For the .NET Framework target the `Registry*` types are included by default in the framework references.

Since this feature only pertains to Windows, and we only release on .NET Framework for Windows there's no need to bring the extra assembly along for the ride in all the other targets (Mac and Linux).
If/once we move to .NET 5 (or 6) on Windows then there'll need to be a review of all `#if NETFRAMEWORK` code, at which point we should hopefully have better support from the linker/trimmer to keep distribution bloat to a minimum on other platforms.

derrickstolee (approved):
I mostly have comments on the docs. The existing discussion of config settings could be reorganized to make the new landscape a bit clearer.

I built this branch locally and confirmed that I was able to load the config setting from a registry key. Excellent work!

docs/enterprise-config.md
```diff
@@ -0,0 +1,57 @@
...
+Git and GCM exactly as they wish. By prefering environment variables and Git
+configuration files over OS store values, these only act as _default values_
+that can always be overriden by the user in the usual ways.
+
+---
+
+## Windows
+
+Default setting values come from the Windows Registry, specifically the
+`HKEY_LOCAL_MACHINE` hive under the following key:
```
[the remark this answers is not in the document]
mjcheetham: Yeah, that's the usual "shorthand" for the hives: HKLM, HKCU, HKU, etc

docs/enterprise-config.md
mjcheetham: > Should we assume 64-bit windows and list only this configuration setting?

We do run on 32-bit Windows (ask @dscho about the download numbers for Git for Windows and mingit).

--

> Is there any chance that we upgrade GCM Core to 64-bit in the future? Should system admins set both keys?

We may very well move to 64-bit in the future, but I don't see that being in the near or mid- future. A move to .NET 5/6 would make it easier to release x86, x86-64, and ARM64 bits natively.. but right now x86 binaries works on all three of those architectures via WoW.

docs/enterprise-config.md
mjcheetham: How about "Enterprise system administrator defaults"

docs/enterprise-config.md
mjcheetham: Done!

mjcheetham:
@derrickstolee & @jeffhostetler I've made some edits to the documentation and naming of the feature to stress the audience and scope: "enterprise system administrator defaults".

jeffhostetler:
Is there a feature in GCM to list the various settings and where they came from (something like `git config --list --show-origin` ?    (But with nice sorting and showing the precedence.)

The Admin could use to ensure that they got everything the way they wanted (especially WRT the 32/64 registry hive stuff).

And later, the Admin could ask individual users to run it and see what they're getting whenever there's a problem (or maybe we include it in `diagnose`).

Just asking.  It doesn't have to be a part of this.

jeffhostetler (approved):

mjcheetham:
> Is there a feature in GCM to list the various settings and where they came from (something like `git config --list --show-origin` ? (But with nice sorting and showing the precedence.)
> 
> The Admin could use to ensure that they got everything the way they wanted (especially WRT the 32/64 registry hive stuff).

That might be a good idea actually. I can throw together an issue tracking this idea.

docs/configuration.md
mjcheetham: I don't think we need to be explicit here and present a list. We're presenting a list in the `enterprise-config.md` file, and the text in `configuration.md` does say these are "default values" only so that shouldn't cause confusion or concern.

docs/enterprise-config.md
dscho: The biggest user of 32-bit Git for Windows is actually Visual Studio (it bundles i686 MinGit for compatibility with i686, x86_64 and ARM64 setups).

PR #405, opened 2021-08-05
Add UI helpers to .deb package & change install location w/ symlinking

### LINKED ISSUE #387: Debian package is missing new UI helper executables
Although the Linux tarball (.tar.gz) release assets contain all the required executables, the Debian package (.deb) only contains the main `git-credential-manager-core` executable and is missing the UI helpers such as `GitHub.UI` and `Atlassian.Bitbucket.UI`.

The issue is here..
https://github.com/microsoft/Git-Credential-Manager-Core/blob/4706801d29038f429bc15353eec4c85f0ac4241b/src/linux/Packaging.Linux/build.sh#L175-L176

..we're not copying all the binaries.

While we're here, we might want to consider installing the binaries somewhere else (like in `/usr/[local/]bin/share/gcm-core` like we do on macOS) and instead create _symlinks_ to `/usr/bin`.

vdye:
Fixes #387 

## Changes
- Add `Atlassian.Bitbucket.UI` and `GitHub.UI` to created .deb package
- Move the binary install location to `/usr/local/bin/share/gcm-core`, with symlinks in the original location (`/usr/bin`)

## Questions
- [x] Is `/usr/local/bin/share/gcm-core` the right install directory to use?
  - Answer: use `/usr/local/share/gcm-core`, symlinked to `/usr/local/bin` 
- [ ] Right now the symlinks are relative paths (see [here](https://github.com/vdye/Git-Credential-Manager-Core/runs/3304256829?check_suite_focus=true#step:3:72)) - should they be absolute path links instead?

### DIFF
--- a/src/linux/Packaging.Linux/build.sh
+++ b/src/linux/Packaging.Linux/build.sh
@@ -154,8 +154,9 @@ tar -czvf "$SYMTARBALL" * || exit 1
 popd
 
 # Build .deb
-INSTALL_TO="$DEBROOT/usr/bin/"
-mkdir -p "$DEBROOT/DEBIAN" "$INSTALL_TO" || exit 1
+INSTALL_TO="$DEBROOT/usr/local/share/gcm-core/"
+LINK_TO="$DEBROOT/usr/local/bin/"
+mkdir -p "$DEBROOT/DEBIAN" "$INSTALL_TO" "$LINK_TO" || exit 1
 
 # make the debian control file
 cat >"$DEBROOT/DEBIAN/control" <<EOF
@@ -167,8 +168,8 @@ Architecture: $ARCH
 Depends:
 Maintainer: GCM-Core <gcmsupport@microsoft.com>
 Description: Cross Platform Git Credential Manager Core command line utility.
- GCM Core supports authentication with a number of Git hosting providers 
- including GitHub, BitBucket, and Azure DevOps. 
+ GCM Core supports authentication with a number of Git hosting providers
+ including GitHub, BitBucket, and Azure DevOps.
  For more information see https://aka.ms/gcmcore
 EOF
 
@@ -176,6 +177,14 @@ EOF
 cp "$PAYLOAD/git-credential-manager-core" "$INSTALL_TO" || exit 1
 cp "$PAYLOAD/Atlassian.Bitbucket.UI" "$INSTALL_TO" || exit 1
 cp "$PAYLOAD/GitHub.UI" "$INSTALL_TO" || exit 1
+
+# Create symlinks
+ln -s -r "$INSTALL_TO/git-credential-manager-core" \
+    "$LINK_TO/git-credential-manager-core" || exit 1
+ln -s -r "$INSTALL_TO/Atlassian.Bitbucket.UI" \
+    "$LINK_TO/Atlassian.Bitbucket.UI" || exit 1
+ln -s -r "$INSTALL_TO/GitHub.UI" \
+    "$LINK_TO/GitHub.UI" || exit 1
 
 dpkg-deb --build "$DEBROOT" "$DEBPKG" || exit 1
 
--- a/src/linux/Packaging.Linux/build.sh
+++ b/src/linux/Packaging.Linux/build.sh
@@ -172,8 +172,10 @@ Description: Cross Platform Git Credential Manager Core command line utility.
  For more information see https://aka.ms/gcmcore
 EOF
 
-# Copy single binary to target installation location
+# Copy GCM Core & UI helper binaries to target installation location
 cp "$PAYLOAD/git-credential-manager-core" "$INSTALL_TO" || exit 1
+cp "$PAYLOAD/Atlassian.Bitbucket.UI" "$INSTALL_TO" || exit 1
+cp "$PAYLOAD/GitHub.UI" "$INSTALL_TO" || exit 1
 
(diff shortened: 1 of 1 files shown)


mjcheetham:
> s this a test that occasionally fails, or is this change interacting with the tests somehow?

I've not seen this failure in CI before, but I've heard of the same underlying issue affecting some customers who use the Keychain from other C# projects, but only sporadically.

mjcheetham:
@vdye I've pushed a commit (1cefe5947ded9dbdc93de8183afa81b83f6a32bf) that should help side-step the flakey Keychain test for macOS. If you rebase this work on the latest `main` branch, you should be good!

mjcheetham:
> Is /usr/local/bin/share/gcm-core the right install directory to use?

We had a similar question when first putting the Linux support in: to use `/usr` or `/usr/local`. In the end we opted for `/usr` (even though on macOS we're using `/usr/local`).

See <https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04.html>

Re-reading the FHS I'd argue that `/usr/local` should be the correct place, and we're currently placing bits in the wrong place:

> 4.9.1. Purpose
> The /usr/local hierarchy is for use by the system administrator when installing software locally. It needs to be safe from being overwritten when the system software is updated. It may be used for programs and data that are shareable amongst a group of hosts, but not found in /usr.
>
>Locally installed software must be placed within /usr/local rather than /usr unless it is being installed to replace or upgrade software in /usr.

Thoughts @derrickstolee @jeffhostetler @dscho ?

derrickstolee:
> Re-reading the FHS I'd argue that `/usr/local` should be the correct place, and we're currently placing bits in the wrong place:
> 
> > 4.9.1. Purpose
> > The /usr/local hierarchy is for use by the system administrator when installing software locally. It needs to be safe from being overwritten when the system software is updated. It may be used for programs and data that are shareable amongst a group of hosts, but not found in /usr.
> > Locally installed software must be placed within /usr/local rather than /usr unless it is being installed to replace or upgrade software in /usr.

I agree that `/usr/local/` is the right place. That's where we put `microsoft/git` on install, too.

The biggest question is: how do we get users who currently have it in `/usr` out of their current state? `/usr` might take priority, so if we simply start adding it to `/usr/local` we will have those users stuck on an older version. (Right?)

mjcheetham:
> `/usr` might take priority, so if we simply start adding it to `/usr/local` we will have those users stuck on an older version. (Right?)

I believe you'd want `/usr/local` to take priority over `/usr`. Logically your local install should take precedence over any distribution binaries, no (hence why we put `microsoft/git` there)?

I know on macOS the default `PATH` puts `/usr/local/bin` ahead of `/usr/[s]bin` which is ahead of `/[s]bin`.

In lieu of more concrete defaults docs, here's my findings of default `PATH` values for various distros:

Distribution|Version|Default `PATH`
-|-|-
Alpine|3.14|`/sbin:/usr/sbin:/bin:/usr/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin`
CentOS|8|`/home/$USER/.local/bin:/home/$USER/bin:/usr/local/bin:/usr/local/sbin:/usr/bin:/usr/sbin`
openSUSE|Leap 15|`/home/$USER/bin:/usr/local/bin:/usr/bin:/bin`
Ubuntu|18.04|`/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin`

Using the `getconf` tool on Ubuntu 18.04 however you get this:

```shell
> getconf PATH
/bin:/usr/bin
```

..which is the opposite, and doesn't list `local` at all(!)

jeffhostetler:
On my Mac:

```
$ getconf PATH
/usr/bin:/bin:/usr/sbin:/sbin
```

whereas my $PATH has (somehow) accumulated `/usr/local/bin`.

mjcheetham:
> whereas my $PATH has (somehow) accumulated /usr/local/bin.

Mac has the `path_helper` tool and its additional way of setting up the `PATH`.

I really don't think there is anything akin to an default ordering, defacto or otherwise. 😢

src/linux/Packaging.Linux/build.sh
```diff
@@ -154,8 +154,9 @@ tar -czvf "$SYMTARBALL" * || exit 1
 popd
 
 # Build .deb
-INSTALL_TO="$DEBROOT/usr/bin/"
-mkdir -p "$DEBROOT/DEBIAN" "$INSTALL_TO" || exit 1
+INSTALL_TO="$DEBROOT/usr/local/bin/share/gcm-core/"
```
mjcheetham, line 157:
```suggestion
INSTALL_TO="$DEBROOT/usr/local/share/gcm-core/"
```

..to match GCM Core on macOS.

src/linux/Packaging.Linux/build.sh
```diff
@@ -154,8 +154,9 @@ tar -czvf "$SYMTARBALL" * || exit 1
...
+LINK_TO="$DEBROOT/usr/bin/"
```
mjcheetham, line 158:
```suggestion
LINK_TO="$DEBROOT/usr/local/bin/"
```

..if we're placing binaries in `/usr/local/share` then the symlinks should match and be in `/usr/local/bin`

mjcheetham (changes requested):

mjcheetham (approved):
:shipit:

Issue #242, opened 2020-12-01
Respect `http.sslCAInfo` when set and validate TLS certs against it


mjcheetham:
**Feature description**

GCM Core uses the .NET HTTP stack, which itself uses the native OS crypto stacks (schannel/CNG on Windows, Apple's Security Framework as of .NET Core 2.0, and OpenSSL on Linux).

When using self-signed certificates over HTTPS remotes with Git on Windows or macOS with the default options, the `http.sslCAInfo` option has no effect, since this is only used by OpenSSL.

Currently the only option is for the user to install/trust their certificates in the system store (Windows Certificate Store, or the macOS Keychain) to have GCM allow connections using the certificates (or disable TLS verification outright via `http.sslVerify = false`).

We should be able to use the same extension point we're currently using for the TLS verification disabling to respect the `http.sslCAInfo` setting.

(see: https://github.com/dotnet/announcements/issues/21)

dscho:
If you want to be _fully_ consistent with Git for Windows, you should also ignore `http.sslCAInfo` if `http.backend == "schannel" && (http.schannelUseSSLCAInfo == unset || http.schannelUseSSLCAInfo == "false")`.

vtbassmatt:
@mjcheetham I might be missing something - is there a need to trust a self-signed cert only in GCM Core but not in the system's cert store? The latter feels like the right thing to me.

mjcheetham:
@dscho is it correct that Git supports some special magic values for config paths, like `%PREFIX/` or something?
We should probably also support that in GCM's handling so we match behaviour between GCM and Git.

dscho:
> 
> 
> @dscho is it correct that Git supports some special magic values for config paths, like `%PREFIX/` or something?
> We should probably also support that in GCM's handling so we match behaviour between GCM and Git.

Yes, current Git for Windows [special-cases paths in the config which start with a single `/`](https://github.com/git-for-windows/git/commit/40c063fb720bbee9caaa802e3d48ed51303318f9#diff-dd5e9dc8fe1f52308750c5b22086433b63835d5a70fc1d0b910fba8b1aceb694): They are treated as being relative to the installation location. (It is actually slightly more complicated: if you install Git for Windows into `C:\Program Files\Git`, then `git.exe` will live in `C:\Program Files\Git\mingw64\bin\git.exe` (for 64-bit) and the path is interpreted to be relative to that `mingw64` directory, the "prefix").

[...]

PR #411, opened 2021-08-11
242 respect ssl ca info

vdye:
Implements #242 

## Changes
- Add parsing for `sslCAInfo` git config parameter (and corresponding environment variable)
  - Ignored if `sslBackend` is `schannel` and `schannelUseSSLCAInfo` is false or not set
- If `sslCAInfo` _is_ set, custom cert validation callback is called for HTTP client
  - If there are any non-certificate chain errors with the server cert, **fail** (see `RemoteCertificateNameMismatch` and `RemoteCertificateNotAvailable` [here](https://docs.microsoft.com/en-us/dotnet/api/system.net.security.sslpolicyerrors?view=net-5.0))
  - If there are any SSL chain errors _other than_ untrusted root, **fail** (see flags listed [here](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509chainstatusflags?view=net-5.0))
  - Check the root of the server certificate 
    - If the server cert root matches the cert pointed to by `sslCAInfo`, **success**
    - If the certs do not match, **fail**
- Unit tests for the `sslCAInfo` setting
- Minor whitespace fixes from my last pull request 😜 

## Testing
For all tests, ran the `get` command as described in "[Development and Debugging](https://github.com/microsoft/Git-Credential-Manager-Core/blob/d9d62c7ce9100c9b1866a5ca3ed390151ff6e7be/docs/development.md#debugging)" (changing the protocol from "http" to "https" for `github.com`)

- **Before starting**, ran `sudo dpkg-reconfigure ca-certificates` to interactively (and temporarily) disable the root cert used by `github.com`
- Do not set `GIT_SSL_CAINFO`
  - Result: 
    ```
    fatal: The SSL connection could not be established, see inner exception.
    fatal: The remote certificate is invalid because of errors in the certificate chain: UntrustedRoot
    ```
- Set `GIT_SSL_CAINFO` to path to trusted root certificate
  - Result: successful `get` execution
- Set `GIT_SSL_CAINFO` to path to an unrelated trusted root certificate for GitHub 
  - Result: 
    ```
    fatal: The SSL connection could not be established, see inner exception.
    fatal: The remote certificate was rejected by the provided RemoteCertificateValidationCallback.
    ```
- Set `GIT_SSL_CAINFO` to path to a non-existent file
  - Result:  failure with the following printout (caused by uncaught exception in callback)
    ```
    fatal: Custom certificate bundle not found at path: <path to fake cert>
    ```

## Questions
- [x] Should this check whether the `sslCAInfo` file exists, or let the program fail?
  - Answer: Check first!
- [x] Should the other `X509ChainStatusFlags` be allowed to continue on to the cert match check? 
  - Answer: nope! be conservative with errors
  - Because the HTTP client first validates the cert against your local root store (and I didn't remove GitHub's actual root CA from my System Roots when testing), I wasn't able to verify that the Untrusted Root error is the _only_ one to come through with an otherwise-valid self-signed cert

### DIFF
--- a/src/shared/Microsoft.Git.CredentialManager/HttpClientFactory.cs
+++ b/src/shared/Microsoft.Git.CredentialManager/HttpClientFactory.cs
@@ -2,10 +2,13 @@
 // Licensed under the MIT license.
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
 using System.Net.Http.Headers;
+using System.Net.Security;
+using System.Security.Cryptography.X509Certificates;
 
 namespace Microsoft.Git.CredentialManager
 {
@@ -34,16 +37,19 @@ namespace Microsoft.Git.CredentialManager
 
     public class HttpClientFactory : IHttpClientFactory
     {
+        private readonly IFileSystem _fileSystem;
         private readonly ITrace _trace;
         private readonly ISettings _settings;
         private readonly IStandardStreams _streams;
 
-        public HttpClientFactory(ITrace trace, ISettings settings, IStandardStreams streams)
+        public HttpClientFactory(IFileSystem fileSystem, ITrace trace, ISettings settings, IStandardStreams streams)
         {
+            EnsureArgument.NotNull(fileSystem, nameof(fileSystem));
             EnsureArgument.NotNull(trace, nameof(trace));
             EnsureArgument.NotNull(settings, nameof(settings));
             EnsureArgument.NotNull(streams, nameof(streams));
 
+            _fileSystem = fileSystem;
             _trace = trace;
             _settings = settings;
             _streams = streams;
@@ -70,6 +76,7 @@ namespace Microsoft.Git.CredentialManager
                 handler = new HttpClientHandler();
             }
 
+            // IsCertificateVerificationEnabled takes precedence over custom TLS cert verification
             if (!_settings.IsCertificateVerificationEnabled)
             {
                 _trace.WriteLine("TLS certificate verification has been disabled.");
@@ -82,6 +89,95 @@ namespace Microsoft.Git.CredentialManager
                 ServicePointManager.ServerCertificateValidationCallback = (req, cert, chain, errors) => true;
 #elif NETSTANDARD
                 handler.ServerCertificateCustomValidationCallback = (req, cert, chain, errors) => true;
+#endif
+            }
+            // If schannel is the TLS backend, custom certificate usage must be explicitly enabled
+            else if (!string.IsNullOrWhiteSpace(_settings.CustomCertificateBundlePath) &&
+                ((_settings.TlsBackend != TlsBackend.Schannel) || _settings.UseCustomCertificateBundleWithSchannel))
+            {
+                string certBundlePath = _settings.CustomCertificateBundlePath;
+                _trace.WriteLine($"Custom certificate verification has been enabled with certificate bundle at {certBundlePath}");
+
+                // Throw exception if cert bundle file not found
+                if (!_fileSystem.FileExists(certBundlePath))
+                {
+                    throw new FileNotFoundException($"Custom certificate bundle not found at path: {certBundlePath}", certBundlePath);
+                }
+
+                Func<X509Certificate2, X509Chain, SslPolicyErrors, bool> validationCallback = (cert, chain, errors) =>
+                {
+                    // Fail immediately if there are non-chain issues with the remote cert
+                    if ((errors & ~SslPolicyErrors.RemoteCertificateChainErrors) != 0)
+                    {
+                        return false;
+                    }
+
+                    // Import the custom certs
+                    X509Certificate2Collection certBundle = new X509Certificate2Collection();
+                    certBundle.Import(certBundlePath);
+
+                    try
+                    {
+                        // Add the certs to the chain
+                        chain.ChainPolicy.ExtraStore.AddRange(certBundle);
+
+                        // Rebuild the chain
+                        if (chain.Build(cert))
+                        {
+                            return true;
+                        }
+
+                        // Manually handle case where only error is UntrustedRoot
+                        if (chain.ChainStatus.All(status => status.Status == X509ChainStatusFlags.UntrustedRoot))
+                        {
+                            // Verify root is contained within the certBundle
+                            X509Certificate2 rootCert = chain.ChainElements[chain.ChainElements.Count - 1].Certificate;
+                            var matchingCerts = certBundle.Find(X509FindType.FindByThumbprint, rootCert.Thumbprint, false);
+                            if (matchingCerts.Count > 0)
+                            {
+                                // Check the content of the first matching cert found (do
+                                // not try others if mismatched - mirrors OpenSSL:
+                                // https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_default_verify_paths.html#WARNINGS)
+                                return rootCert.RawData.SequenceEqual(matchingCerts[0].RawData);
+                            }
+                            else
+                            {
+                                // Untrusted root not found in custom cert bundle
+                                return false;
+                            }
+                        }
+
+                        // Fail for errors other than UntrustedRoot
+                        return false;
+                    }
+                    finally
+                    {
+                        // Dispose imported cert bundle
+                        for (int i = 0; i < certBundle.Count; i++)
+                        {
+                            certBundle[i].Dispose();
+                        }
+                    }
+                };
+
+                // Set the custom server certificate validation callback.
+                // NOTE: this is executed after the default platform server certificate validation is performed
+#if NETFRAMEWORK
+                ServicePointManager.ServerCertificateValidationCallback = (_, cert, chain, errors) =>
+                {
+                    // Fail immediately if the cert or chain isn't present
+                    if (cert is null || chain is null)
+                    {
+                        return false;
+                    }
+
+                    using (X509Certificate2 cert2 = new X509Certificate2(cert))
+                    {
+                        return validationCallback(cert2, chain, errors);
+                    }
+                };
+#elif NETSTANDARD
+                handler.ServerCertificateCustomValidationCallback = (_, cert, chain, errors) => validationCallback(cert, chain, errors);
 #endif
             }
 
--- a/src/shared/Microsoft.Git.CredentialManager/Settings.cs
+++ b/src/shared/Microsoft.Git.CredentialManager/Settings.cs
@@ -119,6 +119,25 @@ namespace Microsoft.Git.CredentialManager
         /// Credential backing store override.
         /// </summary>
         string CredentialBackingStore { get; }
+
+        /// <summary>
+        /// Optional path to a file containing one or more certificates that should
+        /// be used *exclusively* when verifying server certificate chains.
+        /// </summary>
+        /// <remarks>The default value is null if unset.</remarks>
+        string CustomCertificateBundlePath { get; }
+
+        /// <summary>
+        /// The SSL/TLS backend.
+        /// </summary>
+        TlsBackend TlsBackend { get; }
+
+        /// <summary>
+        /// True if, when using an schannel backend, using certificates from the
+        /// CustomCertificateBundlePath is allowed.
+        /// </summary>
+        /// <remarks>The default value is false if unset.</remarks>
+        bool UseCustomCertificateBundleWithSchannel { get; }
     }
 
     public class ProxyConfiguration
@@ -161,6 +180,13 @@ namespace Microsoft.Git.CredentialManager
         /// List of host names that should not be proxied.
         /// </summary>
         public ICollection<string> BypassHosts { get; }
+    }
+
+    public enum TlsBackend
+    {
+        OpenSsl,
+        Schannel,
+        Other,
     }
 
     public class Settings : ISettings
@@ -391,6 +417,18 @@ namespace Microsoft.Git.CredentialManager
                 return true;
             }
         }
+
+        public string CustomCertificateBundlePath =>
+            TryGetSetting(KnownEnvars.GitSslCaInfo, KnownGitCfg.Http.SectionName, KnownGitCfg.Http.SslCaInfo, out string value) ? value : null;
+
+        public TlsBackend TlsBackend =>
+            TryGetSetting(null, KnownGitCfg.Http.SectionName, KnownGitCfg.Http.SslBackend, out string config)
+                ? (Enum.TryParse(config, true, out TlsBackend backend) ? backend : CredentialManager.TlsBackend.Other)
+                : default(TlsBackend);
+
+        public bool UseCustomCertificateBundleWithSchannel =>
+            TryGetSetting(null, KnownGitCfg.Http.SectionName, KnownGitCfg.Http.SchannelUseSslCaInfo, out string schannelUseSslCaInfo) &&
+                schannelUseSslCaInfo.ToBooleanyOrDefault(false);
 
         public ProxyConfiguration GetProxyConfiguration()
         {
--- a/src/shared/Microsoft.Git.CredentialManager.Tests/HttpClientFactoryTests.cs
+++ b/src/shared/Microsoft.Git.CredentialManager.Tests/HttpClientFactoryTests.cs
@@ -15,7 +15,7 @@ namespace Microsoft.Git.CredentialManager.Tests
         [Fact]
         public void HttpClientFactory_GetClient_SetsDefaultHeaders()
         {
-            var factory = new HttpClientFactory(Mock.Of<ITrace>(), Mock.Of<ISettings>(), new TestStandardStreams());
+            var factory = new HttpClientFactory(Mock.Of<IFileSystem>(), Mock.Of<ITrace>(), Mock.Of<ISettings>(), new TestStandardStreams());
 
             HttpClient client = factory.CreateClient();
 
@@ -27,7 +27,7 @@ namespace Microsoft.Git.CredentialManager.Tests
         [Fact]
         public void HttpClientFactory_GetClient_MultipleCalls_ReturnsNewInstance()
         {
-            var factory = new HttpClientFactory(Mock.Of<ITrace>(), Mock.Of<ISettings>(), new TestStandardStreams());
+            var factory = new HttpClientFactory(Mock.Of<IFileSystem>(), Mock.Of<ITrace>(), Mock.Of<ISettings>(), new TestStandardStreams());
 
             HttpClient client1 = factory.CreateClient();
             HttpClient client2 = factory.CreateClient();
@@ -47,7 +47,7 @@ namespace Microsoft.Git.CredentialManager.Tests
                 RemoteUri = repoRemoteUri,
                 RepositoryPath = repoPath
             };
-            var httpFactory = new HttpClientFactory(Mock.Of<ITrace>(), settings, Mock.Of<IStandardStreams>());
+            var httpFactory = new HttpClientFactory(Mock.Of<IFileSystem>(), Mock.Of<ITrace>(), settings, Mock.Of<IStandardStreams>());
 
             bool result = httpFactory.TryCreateProxy(out IWebProxy proxy);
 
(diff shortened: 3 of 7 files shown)


src/shared/Microsoft.Git.CredentialManager/HttpClientFactory.cs
```diff
@@ -85,6 +87,35 @@ public HttpClient CreateClient()
 #endif
             }
 
+            string sslCaInfo = _settings.SslCaInfo;
+            if (!string.IsNullOrWhiteSpace(sslCaInfo))
```
mjcheetham, line 91: Just above this we check if `GIT_SSL_NO_VERIFY` (or the git config equivalent) has been set and set the certificate validation callback(s) to return `true`.

If both `GIT_SSL_NO_VERIFY` **_and_** `http.sslCAInfo` have been set, what should the outcome be (which callback is run) if the root certificate doesn't match? Does "no verify" win? We should match Git's behaviour.

Should probably join these blocks together in an `if-elseif`, and also validate _here_ if we should use `sslCAInfo` based on `sslBackend` (not in `ISettings`).

src/shared/Microsoft.Git.CredentialManager/Settings.cs
```diff
@@ -119,6 +119,12 @@ public interface ISettings : IDisposable
         /// Credential backing store override.
         /// </summary>
         string CredentialBackingStore { get; }
+
+        /// <summary>
+        /// Optional path to a custom root certificate authority to trust
```
mjcheetham, line 124: nit:

```suggestion
        /// Optional path to a custom root certificate authority to trust.
```

src/shared/Microsoft.Git.CredentialManager/Settings.cs
```diff
@@ -392,6 +398,30 @@ public bool IsCertificateVerificationEnabled
...
+            get
+            {
+                // First, verify that, if schannel is the HTTP backend, sslCaInfo must be explicitly enabled to be used
+                if (TryGetSetting(null, KnownGitCfg.Http.SectionName, KnownGitCfg.Http.SslBackend, out string backend) &&
+                    StringComparer.OrdinalIgnoreCase.Equals(backend, "schannel") &&
+                    (!TryGetSetting(null, KnownGitCfg.Http.SectionName, KnownGitCfg.Http.SchannelUseSslCaInfo, out string schannelUseSslCaInfo) ||
+                    !schannelUseSslCaInfo.ToBooleanyOrDefault(false)))
+                {
+                    return null;
+                }
```
mjcheetham, line 412: Personally I would say the `SslCaInfo` property on `ISettings` should return the _actual_ resolved value of `GIT_SSL_CAINFO/http.sslCAInfo`, and _the caller_ should be responsible for enforcing the "only if the backend is not schannel then use the value".

Would mean exposing both the `sslBackend` and `schannelUseSSLCAInfo` settings as additional properties on `ISettings`, which is probably a good idea.

src/shared/Microsoft.Git.CredentialManager/Settings.cs
```diff
@@ -392,6 +398,30 @@ public bool IsCertificateVerificationEnabled
             }
         }
 
+        public string SslCaInfo
+        {
+            get
+            {
+                // First, verify that, if schannel is the HTTP backend, sslCaInfo must be explicitly enabled to be used
+                if (TryGetSetting(null, KnownGitCfg.Http.SectionName, KnownGitCfg.Http.SslBackend, out string backend) &&
```
mjcheetham, line 406: Just checking, does `http.sslBackend` not have an environment variable equivalent? Just curious; I thought most Git settings had both.

src/shared/Microsoft.Git.CredentialManager/Settings.cs
```diff
@@ -392,6 +398,30 @@ public bool IsCertificateVerificationEnabled
...
+                    StringComparer.OrdinalIgnoreCase.Equals(backend, "schannel") &&
```
mjcheetham, line 407: Should these constants (`"schannel"` and `"openssl"`) be `const string`s in _Constants.cs_?

src/shared/Microsoft.Git.CredentialManager/HttpClientFactory.cs
```diff
@@ -85,6 +87,35 @@ public HttpClient CreateClient()
...
+                {
+                    if ((errors & ~SslPolicyErrors.RemoteCertificateChainErrors) != 0 ||
+                        chain.ChainStatus.Any(status => status.Status != X509ChainStatusFlags.UntrustedRoot))
+                    {
+                        // Fail validation if remote cert has any errors other than untrusted root
+                        return false;
+                    }
+
+                    // Get the custom CA root from http.sslCAInfo
+                    X509Certificate2 caBundle = new X509Certificate2(sslCaInfo);
```
mjcheetham, line 103: `X509Certificate2` [implements `IDisposable` starting from .NET Framework 4.6](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509certificate2?view=netframework-4.7.2#remarks), so we should wrap this in a `using` block (or call `Dispose` in the `finally` block of a `try-finally`).

It might also be good measure to test what the happens if an exception is thrown in this validation callback; does it go unhandled?

src/shared/Microsoft.Git.CredentialManager/HttpClientFactory.cs
```diff
@@ -85,6 +87,35 @@ public HttpClient CreateClient()
...
+
+                    // Read the root of the remote certificate chain
+                    X509Certificate2 remoteRoot = chain.ChainElements[chain.ChainElements.Count - 1].Certificate;
```
mjcheetham, line 106: Do we want to check only the root certificate in the chain matches the user supplied one, or should we be happy that the provided certificate is _anywhere_ in the chain?

For example if the user provides cert `X` as the `sslCAInfo`, and the chain we're validating looks like this:

```text
A -> B -> X -> C -> D
```

..should this chain be "ok"?

mjcheetham (changes requested):
A couple of questions and changes regarding validation and what `ISettings` should return.

> Should this check whether the sslCAInfo file exists, or let the program fail?

`HttpClientFactory` should probably test if the file exists before setting the callback. It should then either warn or die (throw an exception), depending on what Git's behaviour is when `http.sslCAInfo` points to an invalid path.

> Should the other X509ChainStatusFlags be allowed to continue on to the cert match check?

We should be conservative and fail for any certificate error we encounter (and manually check the root matches the supplied one.

> Because the HTTP client first validates the cert against your local root store (and I didn't remove GitHub's actual root CA from my System Roots when testing), I wasn't able to verify that the Untrusted Root error is the only one to come through with an otherwise-valid self-signed cert

This is an interesting behaviour. So setting that callback means the "normal" .NET validation still happens before us? That probably deserves a comment in the code for future readers.

src/shared/Microsoft.Git.CredentialManager/Settings.cs
vdye: As far as I can tell, it doesn't - it's not listed in its config entry in the [Git documentation](https://git-scm.com/docs/git-config#Documentation/git-config.txt-httpsslBackend), and a quick search through the code didn't show a corresponding environment variable.

mjcheetham: Yeah, I just did a short search too and didn't find anything. Will resolve this comment.

src/shared/Microsoft.Git.CredentialManager/HttpClientFactory.cs
vdye: According to the Git implementation and [documentation](https://curl.se/libcurl/c/CURLOPT_CAINFO.html) of the cURL parameters it uses, `GIT_SSL_NO_VERIFY`/`http.sslVerify` should take precedence. I'll update the blocks to use an `if-elseif` structure accordingly.

src/shared/Microsoft.Git.CredentialManager/Settings.cs
vdye: Yes for `schannel`, but would you also want `openssl` included (even though it's not used anywhere here)?

src/shared/Microsoft.Git.CredentialManager/HttpClientFactory.cs
vdye: > X509Certificate2 implements IDisposable starting from .NET Framework 4.6, so we should wrap this in a using block (or call Dispose in the finally block of a try-finally).

Will do!

> It might also be good measure to test what the happens if an exception is thrown in this validation callback; does it go unhandled?

The "non-existent file" test I noted actually leads to an uncaught exception (the `X509Certificate2` constructor reads the file from disk - if it's not found, an exception is thrown. The result is this output:

```
fatal: The SSL connection could not be established, see inner exception.
fatal: Could not find file '<path to the fake cert>'.
```

src/shared/Microsoft.Git.CredentialManager/HttpClientFactory.cs
vdye: After looking into this, I actually got a lot more clarification on what exactly OpenSSL does with this setting ([link to the documentation](https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_default_verify_paths.html#NOTES)):

- Like you suggested, it should allow _any_ of the certs in the CA file, _anywhere_ in the certificate chain
- The specified file can actually contain multiple certificates (so it should probably be an `X509Certificate2Collection`)

Based on that (relating to my earlier question on chain errors) I think the way the certificate is currently verified isn't quite correct. My thought now is:

- Add the set of certificates at `sslCAInfo` to the callback `chain`'s "[ExtraStore](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509chainpolicy.extrastore?view=net-5.0)"
- Rebuild the chain with "[Build](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509chain.build?view=net-5.0)"
- Return the result

It looks like this might still fail with "Untrusted Root" (if the root of the chain is in the file at `sslCAInfo`) - [link](https://github.com/dotnet/runtime/issues/20302) to a relevant discussion, where a fix was unfortunately only just implemented in .NET 5.0. Without that update, I think we'd need to do approximately what I have now, which is to ignore "Untrusted Root" errors (as long as the root certificate matches one in `sslCAInfo`). That'll end up needing some redundant computation, but I should be able to mimic what OpenSSL is doing (check thumbprints of certificates in `sslCAInfo` until a match is found, then compare raw data).

Sorry for the brain dump - just want to get all of this written down so I don't forget it 🙂 Also, let me know if I missed something or if you can think of a better way to handle these things.

src/shared/Microsoft.Git.CredentialManager/Settings.cs
vdye: I can think of 3 potential ways to handle the setting:

1. What you're suggesting - the caller handles compatibility checks with `schannel`
2. Create/rename to a "new" setting like `CustomSslCertificateFile` (similar `IsCertificateVerificationEnabled`, which isn't quite the same as `GIT_SSL_NO_VERIFY`) that internally handles the `schannel` compatibility
3. Instead of a property (as in 2), create a method like `GetCustomSslCertificateFile()` (similar to `GetProxyConfiguration()`)

I like the testability of 2 & 3 (and that they represent more of a "black box" setting that is guaranteed to match Git behavior without any extra checks in potentially multiple places), but if there are scenarios where you'd want the directly-from-config value, then 1 would probably be the best way to go.

mjcheetham: I think the responsibility of applying the cert validation rules lies with the `HttpClientFactory` (since that's where the validation callback also lives). The fact that sslNoVerify wins out to the sslCAInfo is also encoded in how the factory evaluates the settings.

You are right of course that `IsCertificateVerificationEnabled` is not exactly like NoVerify (it's a boolean `NOT` of the underlying setting). I guess that should have been called `IsCertificateVerificationIgnored` really.

The more complex proxy settings implementation is because we're trying to hide somewhat the back-compat of now deprecated proxy setting mechanisms, but also want to warn the user about the specific values mechanisms being used.

I think I'd still prefer option 1, with the following properties, and possibly an `enum` for the backend type:

```csharp
// With this enum we can avoid having to create externally visible constants for "schannel" etc
enum TlsBackend
{
    OpenSsl,
    Schannel,
}

interface ISettings
{
    TlsBackend TlsBackend { get; }
    string CustomRootCertificatePath { get; }
    bool UseCustomRootCertificateWithSchannel { get; }
}
```

Really it's TLS these days, so let's stick with "TLS" in the interface and up.

vdye:
>> Because the HTTP client first validates the cert against your local root store (and I didn't remove GitHub's actual root CA from my System Roots when testing), I wasn't able to verify that the Untrusted Root error is the only one to come through with an otherwise-valid self-signed cert

> This is an interesting behaviour. So setting that callback means the "normal" .NET validation still happens before us? That probably deserves a comment in the code for future readers.

The [official documentation](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclienthandler.servercertificatecustomvalidationcallback?view=net-5.0) is pretty unclear, but I _think_  [the .NET source code](https://github.com/dotnet/runtime/blob/dfd618dc648ba9b11dd0f8034f78113d69f223cd/src/libraries/System.Net.Security/src/System/Net/Security/SecureChannel.cs#L937-L999) confirms it (assuming I traced the through it properly). I'll add a comment explicitly mentioning that in the callback.

src/shared/Microsoft.Git.CredentialManager/Settings.cs
vdye: Because this setting is just a passthrough to libcurl, there are [more options](https://curl.se/libcurl/c/curl_global_sslset.html) than `schannel` and `openssl` (depending on platform). If `schannel` is the only one used here (so far, at least), I think I'd prefer to only include that for now (it keeps us from needing to keep up with mirroring all of libcurl's available options).

src/shared/Microsoft.Git.CredentialManager/Settings.cs
```diff
@@ -392,6 +398,30 @@ public bool IsCertificateVerificationEnabled
...
+                    (!TryGetSetting(null, KnownGitCfg.Http.SectionName, KnownGitCfg.Http.SchannelUseSslCaInfo, out string schannelUseSslCaInfo) ||
```
dscho, line 408: Is there any chance that we can call `git config --type=path http.sslCAInfo`? The [`--type=path` option](https://git-scm.com/docs/git-config#Documentation/git-config.txt---typelttypegt) would ensure that we benefit from Git's own path interpolation.

src/shared/Microsoft.Git.CredentialManager/Settings.cs
vdye: This should be updated now! But please double-check that what's there now approximately matches what you were thinking (especially the `TlsBackend` enum).

src/shared/Microsoft.Git.CredentialManager/HttpClientFactory.cs
vdye: Update on this - because the file is now imported as an `X509Certificate2Collection` (since it can contain multiple certs), it's no longer an `IDisposable` type. Should the collection elements be manually disposed (e.g., in a loop), or left without manual disposal?

src/shared/Microsoft.Git.CredentialManager/Settings.cs
vdye: This would involve some changes to [GitConfiguration.cs](https://github.com/microsoft/Git-Credential-Manager-Core/blob/main/src/shared/Microsoft.Git.CredentialManager/GitConfiguration.cs) (like adding an optional `isPath` argument to `TryGet(...)` or creating a `TryGetPath(...)` method), but I really like the benefits of using Git's path interpolation logic (and not replicating it here). @mjcheetham what do you think?

Does this handle the environment variable version of the config as well (in this case, `GIT_SSL_CAINFO`)? Or is `%(prefix)` (and any other "magic" path interpolation) not supported with environment variables?

dscho: This does _not_ handle the environment variables (but neither does Git interpolate prefix-relative paths in environment variables, so I think we're good).

vdye:
I figured out a way to test this more accurately on Ubuntu (manually disabling the root certificate used by GitHub, then copying that cert to a custom location to use it as a "self-signed" cert) - results have been updated in the "Testing" section of the PR description.

src/shared/Microsoft.Git.CredentialManager/HttpClientFactory.cs
```diff
@@ -84,6 +91,100 @@ public HttpClient CreateClient()
...
+                    {
+                        // Add the certs to the chain
+                        chain.ChainPolicy.ExtraStore.AddRange(certBundle);
+
+                        // Rebuild the chain
+                        if (chain.Build(cert))
+                        {
+                            return true;
+                        }
+                        else if (chain.ChainStatus.Any(status => status.Status != X509ChainStatusFlags.UntrustedRoot))
```
mjcheetham, line 129: nit: don't need the `else` here since we'd have exited the block if this wasn't true.

```suggestion

                       if (chain.ChainStatus.Any(status => status.Status != X509ChainStatusFlags.UntrustedRoot))
```

src/shared/Microsoft.Git.CredentialManager/HttpClientFactory.cs
```diff
@@ -84,6 +91,100 @@ public HttpClient CreateClient()
...
+                        {
+                            // Fail for any errors other than untrusted root
+                            return false;
+                        }
+                        else if (chain.ChainStatus.Any(status => status.Status == X509ChainStatusFlags.UntrustedRoot))
+                        {
+                            // Manually verify root is contained within the certBundle
+                            X509Certificate2 rootCert = chain.ChainElements[chain.ChainElements.Count - 1].Certificate;
```
mjcheetham, line 137: Is it ever possible to have a chain with zero `ChainElements`? (such that `.Count - 1 < 0`)

src/shared/Microsoft.Git.CredentialManager/Settings.cs
```diff
@@ -163,6 +183,13 @@ public class ProxyConfiguration
         public ICollection<string> BypassHosts { get; }
     }
 
+    public enum TlsBackend
+    {
+        Schannel,
+        OpenSsl,
+        Other,
+    }
```
mjcheetham, line 191: I wonder if we include a `TlsBackend.Default = 0` option, or just put `TlsBackend.OpenSsl` at position `0` (and therefore it becomes the `default(TlsBackend)` value)?

@dscho is `http.sslBackend` ever likely to change the default from `openssl`? Is there a way to ask Git what the default value is for a configuration option when unset?

mjcheetham (approved):
This looks good to me! Good job @vdye! Just one nit (if you want to fix.. don't worry if not!) and a question to @dscho about default values for the `http.sslBackend`.

src/shared/Microsoft.Git.CredentialManager/Settings.cs
dscho: `http.sslBackend`'s default is likely to remain unchanged, barring any compelling reason, of course.

src/shared/Microsoft.Git.CredentialManager/HttpClientFactory.cs
vdye: Apparently, [it is possible](https://social.msdn.microsoft.com/Forums/en-US/d7f72468-d8fb-4a37-8be2-82c5ca62485c/httpclienthandlerservercertificatecustomvalidationcallback-receives-empty-certchain?forum=xamarinios) (although, to me, what's described there sounds more like a bug than an intentional feature, and I only saw a reference to it for iOS). However, because this block is predicated on having an Untrusted Root error thrown by the previous chain build, I think it would need at least one cert (the root) in the chain.

src/shared/Microsoft.Git.CredentialManager/Settings.cs
vdye: If `openssl` is the default backend used by Git, would it be more appropriate to use a non-nullable `TlsBackend` for the setting, setting the value to `openssl` if it's unspecified in the user's config?

src/shared/Microsoft.Git.CredentialManager/HttpClientFactory.cs
vdye: Updated - was able to simplify this by rearranging some of the conditions as well!

src/shared/Microsoft.Git.CredentialManager/HttpClientFactory.cs
mjcheetham: Sounds good to me!  Should be fine to leave as is.

src/shared/Microsoft.Git.CredentialManager/Settings.cs
mjcheetham: > @mjcheetham what do you think?

I think this should be discussed in a separate PR/issue, so let's leave this as is today for this change.

PR #420, opened 2021-08-19
address feedback from Azure Identity team

vtbassmatt:


vtbassmatt:
Noting here, because it was brought up in the other PR (and because GitHub won't let me add a comment in an unchanged section of the file).

> If you haven't connected any Microsoft accounts to Windows before, the first account you connect will cause the local Windows user account to be converted to a connected account.

The Azure Identity team was surprised to hear that there was no way to deny this behavior, and suggested that there's a workaround. We're unable to repro the workaround. It seems to stem from an Azure DevOps server-side configuration issue, which we requested to be addressed several months ago but have seen no movement on. So for now, this statement will have to stay -- but if someone can help get the "new" GCM Core client ID trusted on Azure DevOps, I'd love to update this section.

mjcheetham (approved):

vtbassmatt:
I'm leaving out of the doc the fact that MSA sign-in is broken with 21H2, as presumably they'll fix that shortly. Noting it here for anyone spelunking PRs later.

Issue #136, opened 2020-06-18
Port WPF-based helper GUIs to Avalonia for Mac/Linux


mjcheetham:
We currently do not have any GUI for Bitbucket and GitHub on macOS (and Linux) and require a TTY be present and available. 

IDEs and other graphical-only tools that shell out to `git` will either hang (if they fail to set `GIT_TERMINAL_PROMPT` to `0`) or error immediately as there is not TTY available and there are no supported GUI prompts.

The current Windows-only GUIs for Bitbucket and GitHub are written in WPF/XAML. We should be able to port these to Avalonia with minimal effort and get Mac and Linux GUIs "for free".

mminns:
If it is any help to bootstrap this update here is a prototype I built
https://github.com/itofinity/Atlassian-Authentication-Helper

PR #431, opened 2021-08-29
Remove planned features section, completed feature

karlhorky:
First of all, thanks for this project! I've seen it make things much smoother for some of our bootcamp students. Considering also including this in our [macOS System Setup guide](https://github.com/upleveled/system-setup/blob/main/macos.md) to avoid the problems associated with password-based CLI authentication failing (due to the [new changes to remove password-based auth for Git operations](https://github.blog/2020-12-15-token-authentication-requirements-for-git-operations/))

The native UI feature from issue https://github.com/microsoft/Git-Credential-Manager-Core/issues/136 was completed in https://github.com/microsoft/Git-Credential-Manager-Core/pull/336 by https://github.com/mjcheetham .

cc @mjcheetham

---

PS. Side question: Is my assumption in the first paragraph above correct? If I get the students to install the Git Credential Manager Core on their macOS systems via Homebrew, will this avoid them getting the dreaded error below?

```
remote: Support for password authentication was removed on August 13, 2021. Please use a personal access token instead.
```

My assumption is that Git Credential Manager Core will avoid this by generating an auth token for them, but I haven't tested this yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,10 +35,6 @@ Proxy support|&#10003;|&#10003;|&#10003;
 
 (\*\*) Debian package offered but not yet available on an official Microsoft feed.
 
-### Planned features
-
-- [ ] macOS/Linux native UI ([#136](https://github.com/microsoft/Git-Credential-Manager-Core/issues/136))
-
 ## Download and Install
 
 ### macOS Homebrew


mjcheetham (approved):
Hi @karlhorky!

Thanks for the PR to remove the now completed planned features list.

> PS. Side question: Is my assumption in the first paragraph above correct? If I get the students to install the Git Credential Manager Core on their macOS systems via Homebrew, will this avoid them getting the dreaded error below?

Yes, you are correct. Installing and configuring GCM on macOS (such as via Homebrew) will mean when doing a `git clone/fetch/pull`, GCM will prompt users and create a token on their behalf (and store it for later use). You will not see such a "passwords are not supported" error anymore.

karlhorky:
Great, thanks for the merge and the confirmation! I think that I will add this to our System Setup guide then, awesome!

PR #436, opened 2021-09-02
Split macOS and Linux implementations and fix password echo bug

### LINKED ISSUE #432: Hide password when supplying basic credentials
Currently, the following happens:

```console
$ git-credential-manager-core --version
2.0.498+7ad55fb809
$ git config credential.helper
/usr/bin/git-credential-manager-core
$ git clone https://review.gerrithub.io/a/felipecrs/gerrit-jenkins-test
Cloning into 'gerrit-jenkins-test'...
Enter basic credentials for 'https://review.gerrithub.io/':
Username: felipecrs
Password: test1234
```

Ideally, the password should be hidden.

mjcheetham:
Previously the `PosixTerminal` class used the POSIX `termios` structure and `tcgetattr`/`tcsetattr` functions for both macOS and Linux systems.

However, it transpires that the definition of the `termios` flags fields differs between Mac and Linux. On Linux the fields are `unsigned int` (4 bytes), but on macOS the fields are `unsigned long` (8 bytes).

This change splits the `termios` structure into Mac and Linux flavours and sets the correct field offsets.

This also fixes a bug with the bit fiddling.. previously we correctly cleared the `echo` bit when needed, but failed to set the bit if needed. This was not really an issue since echo is usually default "on" anyway.

Fixes #432

### DIFF
--- a/src/shared/Microsoft.Git.CredentialManager/CommandContext.cs
+++ b/src/shared/Microsoft.Git.CredentialManager/CommandContext.cs
@@ -110,7 +110,7 @@ namespace Microsoft.Git.CredentialManager
                 SessionManager    = new MacOSSessionManager();
                 SystemPrompts     = new MacOSSystemPrompts();
                 Environment       = new PosixEnvironment(FileSystem);
-                Terminal          = new PosixTerminal(Trace);
+                Terminal          = new MacOSTerminal(Trace);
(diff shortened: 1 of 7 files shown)


dscho (approved):
Looks good!

derrickstolee (approved):

PR #439, opened 2021-09-02
Use built-in `git` path interpolation for config settings

vdye:
## Changes
- Added optional config type specification ([documentation](https://git-scm.com/docs/git-config#Documentation/git-config.txt---typelttypegt)) to `GitConfiguration` getter functions
- Added `TryGetPathSetting` to `Settings`, including handling URL-scoped config settings
- Updated `sslCAInfo` & credential helper settings to use path type from git config

## Notes
- Internal to `GitConfiguration`, an enum (`GitConfigurationType`) is used to specify the type of the config setting, but the `TryGet` extension & functions in `Settings` instead use an `isPath` boolean. 
  - My main reasoning for using the boolean is that paths are the only config type that seem particularly useful for settings in GCM Core, and abstracting that away has some precedent with `GitConfigurationLevel`.

### DIFF
--- a/src/shared/Microsoft.Git.CredentialManager.Tests/GitConfigurationTests.cs
+++ b/src/shared/Microsoft.Git.CredentialManager.Tests/GitConfigurationTests.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.IO;
 using Microsoft.Git.CredentialManager.Tests.Objects;
 using Xunit;
 using static Microsoft.Git.CredentialManager.Tests.GitTestUtilities;
@@ -137,7 +138,7 @@ namespace Microsoft.Git.CredentialManager.Tests
             var git = new GitProcess(trace, gitPath, repoPath);
             IGitConfiguration config = git.GetConfiguration();
 
-            bool result = config.TryGet("user.name", out string value);
+            bool result = config.TryGet("user.name", false, out string value);
             Assert.True(result);
             Assert.NotNull(value);
             Assert.Equal("john.doe", value);
@@ -154,9 +155,80 @@ namespace Microsoft.Git.CredentialManager.Tests
             IGitConfiguration config = git.GetConfiguration();
 
             string randomName = $"{Guid.NewGuid():N}.{Guid.NewGuid():N}";
-            bool result = config.TryGet(randomName, out string value);
+            bool result = config.TryGet(randomName, false, out string value);
             Assert.False(result);
             Assert.Null(value);
+        }
+
+        [Fact]
+        public void GitConfiguration_TryGet_IsPath_True_ReturnsCanonicalPath()
+        {
+            string repoPath = CreateRepository(out string workDirPath);
+            ExecGit(repoPath, workDirPath, "config --local example.path ~/test").AssertSuccess();
+
+            string homeDirectory = Environment.GetFolderPath(Environment.SpecialFolder.UserProfile);
+            string gitPath = GetGitPath();
+            var trace = new NullTrace();
+            var git = new GitProcess(trace, gitPath, repoPath);
+            IGitConfiguration config = git.GetConfiguration();
+
+            bool result = config.TryGet("example.path", true, out string value);
+            Assert.True(result);
+            Assert.NotNull(value);
+            Assert.Equal(Path.GetFullPath($"{homeDirectory}/test").Replace("\\", "/"), value);
+        }
+
+        [Fact]
+        public void GitConfiguration_TryGet_IsPath_False_ReturnsRawConfig()
+        {
+            string repoPath = CreateRepository(out string workDirPath);
+            ExecGit(repoPath, workDirPath, "config --local example.path ~/test").AssertSuccess();
+
+            string gitPath = GetGitPath();
+            var trace = new NullTrace();
+            var git = new GitProcess(trace, gitPath, repoPath);
+            IGitConfiguration config = git.GetConfiguration();
+
+            bool result = config.TryGet("example.path", false, out string value);
+            Assert.True(result);
+            Assert.NotNull(value);
+            Assert.Equal($"~/test", value);
+        }
+
+        [Fact]
+        public void GitConfiguration_TryGet_BoolType_ReturnsCanonicalBool()
+        {
+            string repoPath = CreateRepository(out string workDirPath);
+            ExecGit(repoPath, workDirPath, "config --local example.bool fAlSe").AssertSuccess();
+
+            string gitPath = GetGitPath();
+            var trace = new NullTrace();
+            var git = new GitProcess(trace, gitPath, repoPath);
+            IGitConfiguration config = git.GetConfiguration();
+
+            bool result = config.TryGet(GitConfigurationLevel.Local, GitConfigurationType.Bool,
+                "example.bool", out string value);
+            Assert.True(result);
+            Assert.NotNull(value);
+            Assert.Equal("false", value);
+        }
+
+        [Fact]
+        public void GitConfiguration_TryGet_BoolWithoutType_ReturnsRawConfig()
+        {
+            string repoPath = CreateRepository(out string workDirPath);
+            ExecGit(repoPath, workDirPath, "config --local example.bool fAlSe").AssertSuccess();
+
+            string gitPath = GetGitPath();
+            var trace = new NullTrace();
+            var git = new GitProcess(trace, gitPath, repoPath);
+            IGitConfiguration config = git.GetConfiguration();
+
+            bool result = config.TryGet(GitConfigurationLevel.Local, GitConfigurationType.Raw,
+                "example.bool", out string value);
+            Assert.True(result);
+            Assert.NotNull(value);
+            Assert.Equal("fAlSe", value);
         }
 
         [Fact]
--- a/src/shared/Microsoft.Git.CredentialManager/GitConfiguration.cs
+++ b/src/shared/Microsoft.Git.CredentialManager/GitConfiguration.cs
@@ -110,6 +110,8 @@ namespace Microsoft.Git.CredentialManager
 
     public class GitProcessConfiguration : IGitConfiguration
     {
+        private static readonly GitVersion TypeConfigMinVersion = new GitVersion(2, 18, 0);
+
         private readonly ITrace _trace;
         private readonly GitProcess _git;
 
@@ -448,14 +450,26 @@ namespace Microsoft.Git.CredentialManager
             }
         }
 
-        private static string GetCanonicalizeTypeArg(GitConfigurationType type)
+        private string GetCanonicalizeTypeArg(GitConfigurationType type)
         {
-            return type switch
+            if (_git.Version >= TypeConfigMinVersion)
             {
-                GitConfigurationType.Bool         => "--bool",
-                GitConfigurationType.Path         => "--path",
-                _                                 => null
-            };
+                return type switch
+                {
+                    GitConfigurationType.Bool   => "--type=bool",
+                    GitConfigurationType.Path   => "--type=path",
+                    _                           => null
+                };
+            }
+            else
+            {
+                return type switch
+                {
+                    GitConfigurationType.Bool   => "--bool",
+                    GitConfigurationType.Path   => "--path",
+                    _                           => null
+                };
+            }
         }
 
         public static string QuoteCmdArg(string str)
--- a/src/shared/Microsoft.Git.CredentialManager/GitConfiguration.cs
+++ b/src/shared/Microsoft.Git.CredentialManager/GitConfiguration.cs
@@ -21,6 +21,17 @@ namespace Microsoft.Git.CredentialManager
         Unknown,
     }
 
+    public enum GitConfigurationType
+    {
+        Raw,
+        Bool,
+        Int,
+        BoolOrInt,
+        Path,
+        ExpiryDate,
+        Color
+    }
+
     public interface IGitConfiguration
     {
         /// <summary>
@@ -34,10 +45,11 @@ namespace Microsoft.Git.CredentialManager
         /// Try and get the value of a configuration entry as a string.
         /// </summary>
         /// <param name="level">Filter to the specific configuration level.</param>
+        /// <param name="type">Type constraint to which the config value should be canonicalized.</param>
         /// <param name="name">Configuration entry name.</param>
         /// <param name="value">Configuration entry value.</param>
         /// <returns>True if the value was found, false otherwise.</returns>
-        bool TryGet(GitConfigurationLevel level, string name, out string value);
+        bool TryGet(GitConfigurationLevel level, GitConfigurationType type, string name, out string value);
 
         /// <summary>
         /// Set the value of a configuration entry.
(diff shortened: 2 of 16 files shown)


src/shared/Microsoft.Git.CredentialManager/GitConfiguration.cs
```diff
@@ -34,10 +45,11 @@ public interface IGitConfiguration
         /// Try and get the value of a configuration entry as a string.
         /// </summary>
         /// <param name="level">Filter to the specific configuration level.</param>
+        /// <param name="type">Type constraint to which the config value should be canonicalized.</param>
         /// <param name="name">Configuration entry name.</param>
         /// <param name="value">Configuration entry value.</param>
         /// <returns>True if the value was found, false otherwise.</returns>
-        bool TryGet(GitConfigurationLevel level, string name, out string value);
+        bool TryGet(GitConfigurationLevel level, GitConfigurationType type, string name, out string value);
```
mjcheetham, line 52: I think it might be better/less noisy to put this `GitConfigurationType type` at the end of every existing method's argument list (except `out` params), i.e..

```suggestion
        bool TryGet(GitConfigurationLevel level, string name, GitConfigurationType type, out string value);
```

..and then add appropriate extension methods to the `GitConfigurationExtensions` class that pass the default `GitConfigurationType.None` in as this extra argument.

This means that existing call-sites do not need to be updated (they'll end up resolving to these new extension methods with the old arg lists instead), and for call-sites that _do_ want to also specify a coerced type, it's just adding one more argument to the end of their call.

src/shared/Microsoft.Git.CredentialManager/GitConfiguration.cs
```diff
@@ -180,10 +194,11 @@ public void Enumerate(GitConfigurationLevel level, GitConfigurationEnumerationCa
             }
         }
 
-        public bool TryGet(GitConfigurationLevel level, string name, out string value)
+        public bool TryGet(GitConfigurationLevel level, GitConfigurationType type, string name, out string value)
         {
             string levelArg = GetLevelFilterArg(level);
-            using (Process git = _git.CreateProcess($"config --null {levelArg} {QuoteCmdArg(name)}"))
+            string typeArg = GetCanonicalizeTypeArg(type);
+            using (Process git = _git.CreateProcess($"config --null {levelArg} {typeArg} {QuoteCmdArg(name)}"))
```
mjcheetham, line 201: Which version of Git first supported the `--type=<type>` argument?

I see from the [documentation](https://git-scm.com/docs/git-config#Documentation/git-config.txt---bool) that there used to be (and still is, but deprecated) explicit options for each type instead.

```text
--bool
--int
--bool-or-int
--path
--expiry-date
    Historical options for selecting a type specifier. Prefer instead --type (see above).
```

We don't want to try and use this new argument on an old version of Git or else we'll get usage errors(!) This happened in the past when I added `--show-scope` which does not exist on the (old) version of Git available in some Ubuntu/Debian distributions.

If we can, try and use the oldest still supported arguments. Otherwise we'll need to introduce a `git version` call and check (that could be cached at least).

src/shared/TestInfrastructure/Objects/TestGitConfiguration.cs
```diff
@@ -51,6 +53,12 @@ public bool TryGet(GitConfigurationLevel level, string name, out string value)
                     if (values.Count == 1)
                     {
                         value = values[0];
+
+                        if (type == GitConfigurationType.Path) {
```
mjcheetham, line 57: nit: don't use K&R style

```suggestion
                        if (type == GitConfigurationType.Path)
                        {
```

src/shared/TestInfrastructure/Objects/TestGitConfiguration.cs
```diff
@@ -51,6 +53,12 @@ public bool TryGet(GitConfigurationLevel level, string name, out string value)
...
+                            // Create "fake" canonical path
+                            value = value.Replace("~", CanonicalPathPrefix);
+                        }
+
```
mjcheetham, line 61: nit: remove extra line

```suggestion
```

mjcheetham (changes requested):
Left a few nit comments, but also think it should be possible to reduce all the changes at the existing call-sites if we put the extra `type` arg at the end of the argument list, and add some more extension methods.

Also a question about Git version compatibility with the `--type=<type>` argument.

src/shared/Microsoft.Git.CredentialManager/GitConfiguration.cs
vdye: It looks like [2.18](https://git-scm.com/docs/git-config/2.18.0) is the first released version to support the `--type` parameter (although it only supports `--int`, `--bool`, and `--path` in [2.17 and older](https://git-scm.com/docs/git-config/2.17.0) - fine for the purposes of this change, since it's only using the path type anyway). 

Is there a minimum version of git that GCM Core explicitly supports? If not (or if it's older than 2.18), I'm not sure I like the idea of actively using a deprecated option, so an explicit version check would be a nice way to handle this (and helpful for using newer features in git in general, should they ever be needed). I think `Version` (or `GitVersion`?) would make sense as a property of the `Git` class, but I don't see any other properties there - is that an intentional design decision, or has there just not been any need for public properties there yet?

src/shared/Microsoft.Git.CredentialManager/GitConfiguration.cs
vdye: I had considered that (specifically, adding the argument at the end and having it use `GitConfigurationExtensions.None` by default), but I don't necessarily find it cleaner. There are only 5 non-test usages outside of `GitConfiguration` itself ([1](https://github.com/microsoft/Git-Credential-Manager-Core/pull/439/commits/7d644a4a29b3c15b90682aebc551067ba0def3ed#diff-62a94bca9b60857fb75038ef1cd82ecfc5dc87699a88a31cfadfb62e075261eaL61-R62), [2 & 3](https://github.com/microsoft/Git-Credential-Manager-Core/pull/439/commits/7d644a4a29b3c15b90682aebc551067ba0def3ed#diff-0be9b83a69df29220151264feb5c7fd3822f1a0ba4d9463885bd92bd772f6a60L97-R100), [4](https://github.com/microsoft/Git-Credential-Manager-Core/pull/439/commits/7d644a4a29b3c15b90682aebc551067ba0def3ed#diff-76e78a4449281f62e3fecf0bb1e92425e9bf4e6d031c6e26ffaa3ce9c12974a4L170-R170), [5](https://github.com/microsoft/Git-Credential-Manager-Core/pull/439/commits/7d644a4a29b3c15b90682aebc551067ba0def3ed#diff-76e78a4449281f62e3fecf0bb1e92425e9bf4e6d031c6e26ffaa3ce9c12974a4L221-R221)) of the non-extension `TryGet` & `GetAll` that needed updating, and updating them here is a one-time change. Conversely, an extension method for just those 5 cases is a persistent addition, one that as far as I can tell is just a way of hiding one of two optional parameters (as opposed to the current configuration, which I interpret as "the full method with an extension for hiding defaulting behavior").

One thing this made me realize, though, is that adding more and more options over time could get unwieldy (either with lots of extensions or long function signatures). A `GitConfigOptions` structure (initialized with default values for `GitConfigurationLevel` and `GitConfigurationType`) might be a clean way of handling that and future-proofing it for other options. What do you think?

src/shared/Microsoft.Git.CredentialManager/GitConfiguration.cs
dscho: Support for `--type=path` (and making it the preferred option) was introduced in git/git@fb0dc3bac135e9f6243bd6d293e8c9293c73b9cd, and the first Git version to include this patch was v2.18.0.

src/shared/Microsoft.Git.CredentialManager.Tests/GitConfigurationTests.cs
```diff
@@ -154,11 +155,82 @@ public void GitConfiguration_TryGet_Name_DoesNotExists_ReturnsFalse()
...
+            string homeDirectory = Environment.GetFolderPath(Environment.SpecialFolder.UserProfile);
+            string gitPath = GetGitPath();
+            var trace = new NullTrace();
+            var git = new GitProcess(trace, gitPath, repoPath);
+            IGitConfiguration config = git.GetConfiguration();
+
+            bool result = config.TryGet("example.path", true, out string value);
+            Assert.True(result);
+            Assert.NotNull(value);
+            Assert.Equal(Path.GetFullPath($"{homeDirectory}/test"), Path.GetFullPath(value));
```
dscho, line 178: Do we want to expand `value` to the full path here, or do we actually expect Git to do that for us already (i.e. use `value` here instead of `Path.GetFullPath(value)`)?

src/shared/Microsoft.Git.CredentialManager/GitConfiguration.cs
dscho: The `--path` option was introduced in git/git@1349484e341a3ec2ba02a86c8fbd97ea9dc8c756, the first Git version to support it was v1.7.0. I think that it is sane to require that minimum Git version ;-) (but v2.18.0 might be a bit too new, some Debian LTS versions are still based on v2.7.0 IIRC).

dscho (approved):

src/shared/Microsoft.Git.CredentialManager.Tests/GitConfigurationTests.cs
vdye: I added the `Path.GetFullPath(value)` because, on Windows, it will return a path formatted with forward slashes, whereas `Environment.GetFolderPath(Environment.SpecialFolder.UserProfile)` is formatted with backslashes. To your point, though, we should be testing `value` directly - I updated the implementation to explicitly replace the backslashes in `homeDirectory`, so the values should match.

src/shared/Microsoft.Git.CredentialManager/GitConfiguration.cs
mjcheetham: > There are only 5 non-test usages outside of GitConfiguration itself (1, 2 & 3, 4, 5) of the non-extension TryGet & GetAll that needed updating, and updating them here is a one-time change.

A fair enough argument! 😁 

One final suggestion then.. I read `GitConfigurationType.None` strangely.. I mean, we're not getting a "none" type, but rather we're not coalescing the raw data in to a specific type..

Perhaps `GitConfigurationType.String` or `GitConfigurationType.Raw` would be more suitable, IMO.

mjcheetham (commented):
One final comment about renaming `GitConfigurationType.None` to `.String` or `.Raw`.

src/shared/Microsoft.Git.CredentialManager/GitConfiguration.cs
mjcheetham: Right.. 2.18 is probably too new.

> Is there a minimum version of git that GCM Core explicitly supports?

Historically, no.. not really. On Windows it was "approximately the version that ship together in Git for Windows" but that doesn't hold anymore for Mac/Linux, nor always for Windows too (what with MinGit and VS etc).

A good rule of thumb would be looking at what version is shipped with the previous Ubuntu/Debian LTS.. we don't want to break existing users if possible, and/or what the minimum version of Git is required today. I'm not sure GCM uses _any_ options or commands that were not added a long time ago?

All currently called Git commands and args are:

```text
rev-parse --absolute-git-dir
remote -v show
config (--local|--global|--system) <name> <value>
config (--local|--global|--system) --add <name> <value>
config (--local|--global|--system) --unset <name>
config (--local|--global|--system) --unset-all <name>
config (--local|--global|--system) --replace-all <name> <value> <value-regex>
config --null (--local|--global|--system) <name>
config --null (--local|--global|--system) --list
config --null (--local|--global|--system) --get-all <name>
config --null (--local|--global|--system) --get-regex <name-regex>
```

Adding a version check would future proof us when (or rather _IF_) the old options go away, but I don't think the old options are really going to go away any time soon! @dscho ?

dscho: They won't be going away any time soon, no. But the `--path` option _is_ deprecated, so I would prefer us to use `--type=path` when possible.

src/shared/Microsoft.Git.CredentialManager/GitConfiguration.cs
vdye: Makes sense! Updated to `GitConfigurationType.Raw`.

src/shared/Microsoft.Git.CredentialManager/GitConfiguration.cs
vdye: The last commit in this series adds a `GitVersion` implementation & selects the right flags in `GitProcessConfiguration` based on the `GitProcess` version. Please let me know what you think of that approach (at the very least to make sure I followed repository conventions properly), or if you'd like me to move it into a separate pull request altogether.

src/shared/Microsoft.Git.CredentialManager/GitVersion.cs
```diff
@@ -0,0 +1,152 @@
...
+
+            foreach (string part in splitVersion)
+            {
+                if (Int32.TryParse(part, out int component))
+                {
+                    components.Add(component);
+                }
+                else
+                {
+                    // Exit at the first non-integer component
```
dscho, line 30: Makes sense, this would turn `2.33.0.vfs.2.3` into `[2, 33, 0]`.

TBH version comparing is (in my experience) somewhat of an ill-defined problem, and the more simplistic we can make it without breaking our use cases, the better.

As a point of reference, I would like to link [to the version comparison in Git for Windows' installer](https://github.com/git-for-windows/build-extra/blob/0ad275c0edab6a2e0bf36622411417948bca3dcc/installer/install.iss#L890-L953) and [the code in `git update-git-for-windows`](https://github.com/git-for-windows/build-extra/blob/0ad275c0edab6a2e0bf36622411417948bca3dcc/git-extra/git-update-git-for-windows#L11-L44). As you can see, they both do different things, and neither of them are complete.

I also had a brief hunch that there might be something in .NET, and there is: `System.Version`. Alas, it seems that [the only way to parse a string into an instance of that class](https://docs.microsoft.com/en-us/dotnet/api/system.version.-ctor?view=net-5.0#System_Version__ctor_System_String_) is very limited and would choke on any letters (such as `vfs`).
 
All this is to say: I think this implementation is nice and small, and good enough for our purposes. Thank you for implementing it!

src/shared/Microsoft.Git.CredentialManager/GitVersion.cs
```diff
@@ -0,0 +1,152 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.Git.CredentialManager
+{
+    public class GitVersion : IComparable, IComparable<GitVersion>
```
mjcheetham, line 7: Implementing `IComparable` on types can be hard to get right; looks like you've done a great job - nice! 😁 

Did you consider also implementing the `IEquatable` and `IEquatable<GitVersion>` interfaces?
The implementation is already there in your overridden `bool Equals(object)` method but adding the interface will mean should we ever put `GitVersion` in a dictionary or set, that collection can _know_ it can be equated.

This change does not block this PR by the way..

mjcheetham (approved):
Looks great! Thanks 😁

src/shared/Microsoft.Git.CredentialManager/GitConfiguration.cs
mjcheetham: Looks good to me!

mjcheetham:
Note I merged this without the Ubuntu 16.04 check passing because GitHub Actions has removed that particular image!

PR #446, opened 2021-09-10
Add CodeQL scanning

vtbassmatt:


vtbassmatt:
We seem to be ✅ clean from a CodeQL standpoint. It's the longest-running CI check at 6 minutes, though, so perhaps we don't want to put this in the PR path.

dscho (approved):
Looks good!

FWIW I really like that this is running on all PRs (as opposed to, say, a scheduled job) because that way it is easier to act on failed runs (because of the available context).

If anything, I would remove the commented-out lines, and the code comments that are talking about languages other than C#, but it is definitely not necessary to change that.

vtbassmatt:
> I would remove the commented-out lines, and the code comments that are talking about languages other than C#.

I will definitely clean this up before merging. Thanks!

mjcheetham (approved):

PR #461, opened 2021-09-21
Add support for using WSL Git for configuration

mjcheetham:
This PR enables GCM to understand values of `GIT_EXEC_PATH` that point to a Windows Subsystem for Linux (WSL) path, for example `\\wsl$\ubuntu\usr\libexec\git-core`.

This can use useful for scenarios where the user does not want to install Git for Windows or use it's configuration files, but rather use the WSL Git's view of the world.

Also add a piece of documentation that outlines how to correctly configure GCM and WSL.

### DIFF
--- a/docs/wsl.md
+++ b/docs/wsl.md
@@ -1,0 +1,96 @@
+# Windows Subsystem for Linux (WSL)
+
+Git Credential Manager Core can be used with the [Windows Subsystem for Linux
+(WSL)](https://aka.ms/wsl) to enable secure authentication of your remote Git
+repositories from inside of WSL.
+
+## Supported versions
+
+GCM Core supports being called from both WSL1 and WSL2 installations.
+
+**Note:** In order to use GCM with WSL, _without_ a Git for Windows installation
+you must be using GCM version 2.0.XXX or later.
+You must also be on Windows 10 Version 1903 and later. This is the first version
+of Windows that includes the required `wsl.exe` command-line tool that GCM uses
+to interoperate with Git in your WSL distributions.
+
+## Set up WSL & Git Credential Manager
+
+### Step 1
+
+[Install the latest Git for Windows ⬇️](https://github.com/git-for-windows/git/releases/latest)
+
+### Step 2
+
+_Inside your WSL installation_, run the following command to set GCM as the Git
+credential helper:
+
+```shell
+git config --global credential.helper /mnt/c/Program\ Files/Git/mingw64/libexec/git-core/git-credential-manager-core.exe
+```
+
+### Step 3 (Azure DevOps only)
+
+If you intend to use Azure DevOps you must _also_ set the following Git
+configuration _inside of your WSL installation_.
+
+```shell
+git config --global credential.https://dev.azure.com.useHttpPath true
+```
+
+## Using GCM & WSL without Git for Windows
+
+If you wish to use GCM Core inside of WSL _without installing Git for Windows_
+you must complete additional configuration so that GCM can callback to Git
+inside of your WSL installation.
+
+In **_Windows_** you need to update the `WSLENV` environment variable to include
+the value `GIT_EXEC_PATH/wp`. From an _Administrator_ Command Prompt run the
+following:
+
+```batch
+SETX WSLENV=%WSLENV%:GIT_EXEC_PATH/wp
+```
+
+..and then restart your WSL installation.
+
+## How it works
+
+GCM leverages the built-in interoperability between Windows and WSL, provided by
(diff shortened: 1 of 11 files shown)


src/shared/Microsoft.Git.CredentialManager/EnvironmentBase.cs
```diff
@@ -43,6 +43,18 @@ public interface IEnvironment
...
+        /// Create a process ready to start, with redirected streams.
+        /// </summary>
+        /// <param name="path">Absolute file path of executable or command to start.</param>
+        /// <param name="args">Command line arguments to pass to executable.</param>
+        /// <param name="useShellExecute">
+        /// True to resolve <paramref name="path"/> using the OS shell, false to use as an absolute file path.
+        /// </param>
+        /// <param name="workingDirectory">Working directory for the new process.</param>
+        /// <returns><see cref="Process"/> object ready to start.</returns>
+        Process CreateProcess(string path, string args, bool useShellExecute, string workingDirectory);
```
jeffhostetler, line 57: Do you ever need to pass a custom environment block?  That is, a custom set of environment variables for the child process to inherit.  Like a custom PATH or debugging/test flags or etc.

I'm not saying you actually need it right now, but rather just wondering if we might at some point.

jeffhostetler: Wait, I see you have that in a later class. I think.  I think it is hard to figure out the class hierarchy in a PR....

src/shared/Microsoft.Git.CredentialManager.Tests/WslUtilsTests.cs
```diff
@@ -0,0 +1,111 @@
...
+
+            Assert.NotNull(process);
+            Assert.Equal(expectedArgs, process.StartInfo.Arguments);
+            Assert.Equal(expectedFileName, process.StartInfo.FileName);
+            Assert.True(process.StartInfo.RedirectStandardInput);
+            Assert.True(process.StartInfo.RedirectStandardOutput);
+            Assert.True(process.StartInfo.RedirectStandardError);
+            Assert.False(process.StartInfo.UseShellExecute);
+        }
+
```
jeffhostetler, line 88: This might be overkill, but we might also test that `git --exec-path` returns the expected value.

docs/wsl.md
```diff
@@ -0,0 +1,96 @@
...
+installation, and not shared with others or the Windows host.
+
+## Can I install Git Credential Manager directly inside of WSL?
+
+Yes. Rather than install GCM as a Windows application (and have WSL Git invoke
+the Windows GCM), can you install GCM as a Linux application instead.
+
+To do this, simply follow the [GCM installation instructions for Linux](../README.md#linux-install-instructions).
+
+**Note:** In this scenario, because GCM is running as a Linux application
```
jeffhostetler, line 94: Does this also mean that GCM won't have a GUI available?

jeffhostetler (approved):

vtbassmatt (approved):
Reviewed the doc changes and left suggestions. But overall this is a great change!

src/shared/Microsoft.Git.CredentialManager.Tests/WslUtilsTests.cs
mjcheetham: We cannot do this easily as we need to be invoked from a WSL distro, with specific environment variables set etc. These tests don't actually start any processes, just check that the `Process` object is constructed correctly, and is ready to be `.Start()`ed.

src/shared/Microsoft.Git.CredentialManager/EnvironmentBase.cs
mjcheetham: > Do you ever need to pass a custom environment block?

No. We don't use this (at least at the moment). We just rely on default inheritance behaviour. Should we need this, then yes we can add this in the signature. I'd rather not do that now.

Aside: Windows <--> WSL environment inheritance works differently, and you have to be explicit in which envars are shared across the boundary. The `WSLENV` envar defines which envars are shared, and if there is any path-style transformations. See https://devblogs.microsoft.com/commandline/share-environment-vars-between-wsl-and-windows/

> I think it is hard to figure out the class hierarchy in a PR

The hierarchy is:

```text
IEnvironment <------- EnvironmentBase <--inherit-- PosixEnvironment
                             ^
                             |
                          inherit
                             |
                      WindowsEnvironment
 
```

docs/wsl.md
mjcheetham: The answer here is "it depends". We have GUI options for GCM+Linux. If the user has "gWSL" set up, then Linux GUI apps also work and appear as windows on the Windows host.

PR #464, opened 2021-09-22
Add DPAPI based credential store on Windows

### LINKED ISSUE #452: Create alternative Windows credential store using DPAPI
The Windows Credential Manager (wincredman) is currently the only supported credential store with GCM on Windows. There are several limitations with wincredman, namely:

1. Network logon sessions (such as over SSH) cannot persist credentials
2. There is a small maximum size of credential that can be stored (2560 bytes)

I propose opening up other credential stores (such as the plaintext store) for use on all platforms, as well as introducing a new DPAPI-based credential store on Windows, that can be used in these scenarios.

The DPAPI store would extend from the plaintext store and use the DPAPI `ProtectedData` APIs to encrypt the credential line in the plaintext file format.

mjcheetham:
Introduce a new credential store for Windows that uses DPAPI to encrypt the credentials on disk.

The default credential store on Windows will continue to be the Windows Credential Manager (wincredman), but by offering DPAPI we will allow the following limitations to be overcome:

- Network logon sessions (such as over SSH) cannot persist credentials
- There is a small maximum size of credential that can be stored (2560 bytes)

We only continue to use the wincredman solution as it offers a reasonable GUI for viewing and deleting stored credentials for the user. The plan would be to change the default to the DPAPI credential store if/when we even create our own easy to use GUI for managing GCM's stored credentials.

As part of this change it is now possible to also use the plaintext credential store on Windows and macOS, following the same model as we have for selecting the credential store on Linux. The only difference is that Windows and macOS have a default store, and on Linux you must select one.

Fixes #452


docs/credstores.md
```diff
@@ -1,48 +1,130 @@
...
+
+This credential store uses Windows DPAPI to encrypt credentials which are stored
+as files in your file system. The file structure is the same as the
+[plaintext files credential store](#plaintext-files) except the first line (the
+secret value) is protected by DPAPI.
+
+By default files are stored in `~/.gcm/dpapi_store` but this can be configured using
+the environment variable `GCM_DPAPI_STORE_PATH` environment variable.
+
+If the directory does not exist is will be created.
```
vtbassmatt, line 77: Typo: `s/is/it/`

Nit: `s/does not/doesn't/`

docs/credstores.md
```diff
@@ -140,15 +227,13 @@ Permissions on existing directories will not be modified.
...
 :warning: **WARNING** :warning:
 
 **This storage mechanism is NOT secure!**
 
-**Secrets and credentials are stored in plaintext files _without any security_!<br/>
-Git Credential Manager Core takes no liability for the safety of these
-credentials.**
+**Secrets and credentials are stored in plaintext files _without any security_!**
+
+**Git Credential Manager Core takes no liability for the safety of these credentials.**
```
vtbassmatt, line 236: This is an odd disclaimer. Consider replacing with "We strongly advise against using GCM Core in this configuration" or something like that.

vtbassmatt (approved):
Approving at a conceptual level + doc review. Nice!

docs/credstores.md
mjcheetham: I'll drop this line altogether since the line immediately below already says that.

vdye (approved):
Looks great! The only thing I'd like (if it was easier to integrate) would be tests for the DPAPI credential manager, but it looks like that's something that would be better suited to end-to-end tests anyway.

derrickstolee (approved):
I don't know enough about DPAPI to know if there are any gotchas with that integration, but the code looks well structured.

mjcheetham:
> I don't know enough about DPAPI to know if there are any gotchas with that integration [..]

The DPAPI interface as exposed in .NET (`ProtectedData`) is very simple.. just those `Protect` and `Unprotect` calls with optional entropy (set to `null` for us). This protects the data _for the current user_.

This is the same pattern that MSAL.NET and Visual Studio use for protecting their token caches.
https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/blob/835106e1148bc05405fb5fff3e235feb22b05a63/src/Microsoft.Identity.Client.Extensions.Msal/Accessors/DpApiEncryptedFileAccessor.cs#L38-L61

mjcheetham:
> The only thing I'd like (if it was easier to integrate) would be tests for the DPAPI credential manager [..]

That should be easy to add actually in the existing "unit" tests!

vdye (approved):

PR #468, opened 2021-09-27
supported platforms

vtbassmatt:


README.md
```diff
@@ -28,10 +28,15 @@ Two-factor authentication support for Bitbucket|&#10003;|&#10003;|&#10003;
 Windows Integrated Authentication (NTLM/Kerberos) support|&#10003;|_N/A_|_N/A_
 Basic HTTP authentication support|&#10003;|&#10003;|&#10003;
 Proxy support|&#10003;|&#10003;|&#10003;
+`amd64` support|&#10003;|&#10003;|&#10003;
+`x86` support|&#10003;|_N/A_|&#10003;
```
mjcheetham, line 32:
```suggestion
`x86` support|&#10003;|_N/A_|&#10007;
```

We don't have any `x86` builds for Linux, and not sure if we've tested `x86` on Linux.

Also technically we don't have any native `amd64` bits for Windows, but rely on WoW on 64-Bit Windows to run `x86` binaries.

ldennington:
I may be totally misunderstanding, but did https://github.com/GitCredentialManager/git-credential-manager/pull/468 address this? I thought we were planning to publish the matrix of supported distros, but looks like that PR focused on supported architectures (`x86`, `arm64`, etc.).

vtbassmatt:
@ldennington I wish I had taken better notes! What I _seem_ to recall is that we were going to back off of supporting particular distros (i.e. publishing our own packages) and instead make install-from-source the thing we supported. But maybe I'm mis-remembering?

PR #478, opened 2021-10-06
Add explicit GitHub device code authentication option

mjcheetham:
Implement an explicit OAuth device code authentication mode for the GitHub host provider. Previously the 'web browser' option combined both the interactive/browser based flow (when a UI was present), and a TTY-based device code flow.

This change allows users to select the device code flow even when they have a desktop/UI session present. This is useful to workaround possible problems with the browser loopback/redirect mechanism.

As well as introducing new `"browser"` and `"device"` values, we keep the existing `"oauth"` value for the [`GCM_GITHUB_AUTHMODES`](https://aka.ms/gcmcore-env#gcm_github_authmodes) environment variable and [`credential.gitHubAuthModes`](https://aka.ms/gcmcore-config#credentialgitHubAuthModes) Git config, and let it expand to both `browser` and `device` to allow users who may have set the older value before.

-|Avalonia<br/>(macOS)|Avalonia<br/>(Ubuntu)|WPF<br/>(Windows)
-|-|-|-
GUI Prompt|image|image|image
GUI Code Display|image|image|image
Terminal|image|image|image

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -185,7 +185,9 @@ If this option is not set, then the available authentication modes will be autom
 Value|Authentication Mode
 -|-
 _(unset)_|Automatically detect modes
-`oauth`|OAuth-based authentication
+`oauth`|Expands to: `browser, device`
+`browser`|OAuth authentication via a web browser _(requires a GUI)_
+`device`|OAuth authentication with a device code
 `basic`|Basic/PAT-based authentication
 
 #### Example
--- a/docs/environment.md
+++ b/docs/environment.md
@@ -324,7 +324,9 @@ If this option is not set, then the available authentication modes will be autom
 Value|Authentication Mode
 -|-
 _(unset)_|Automatically detect modes
-`oauth`|OAuth-based authentication
+`oauth`|Expands to: `browser, device`
+`browser`|OAuth authentication via a web browser _(requires a GUI)_
+`device`|OAuth authentication with a device code
 `basic`|Basic/PAT-based authentication
 
 ##### Windows


src/shared/GitHub.Tests/GitHubHostProviderTests.cs
```diff
@@ -121,7 +121,7 @@ public async Task GitHubHostProvider_GetSupportedAuthenticationModes(string uriS
         [Theory]
         [InlineData("https://example.com", null, "0.1", false, AuthenticationModes.Pat)]
         [InlineData("https://example.com", null, "0.1", true, AuthenticationModes.Basic | AuthenticationModes.Pat)]
-        [InlineData("https://example.com", null, "100.0", false, AuthenticationModes.Browser | AuthenticationModes.Pat)]
+        [InlineData("https://example.com", null, "100.0", false, AuthenticationModes.OAuth | AuthenticationModes.Pat)]
```
derrickstolee: This seems like an errant revert of a previous change.

src/shared/GitHub.Tests/GitHubHostProviderTests.cs
```diff
@@ -196,7 +196,7 @@ public async Task GitHubHostProvider_GenerateCredentialAsync_Browser_ReturnsCred
             ghAuthMock.Setup(x => x.GetAuthenticationAsync(expectedTargetUri, null, It.IsAny<AuthenticationModes>()))
                       .ReturnsAsync(new AuthenticationPromptResult(AuthenticationModes.Browser));
 
-            ghAuthMock.Setup(x => x.GetOAuthTokenAsync(expectedTargetUri, It.IsAny<IEnumerable<string>>()))
+            ghAuthMock.Setup(x => x.GetOAuthTokenViaBrowserAsync(expectedTargetUri, It.IsAny<IEnumerable<string>>()))
```
derrickstolee, line 199: In the message for this commit, you say 

```
This is useful to workaround possible
problems with the browser loopback/redirect mechanism.
```

Do you have an example? I'm guessing WSL has something to do with this.

src/shared/GitHub.Tests/GitHubHostProviderTests.cs
derrickstolee: Oh, I see. `OAuth` is the combination of two. Should we have lines for `Browser`, `Device`, _and_ `OAuth` on this test?

derrickstolee (approved):

PR #479, opened 2021-10-06
Added missing quotes in WSL instructions

nimatt:
One of the commands for WSL configuration was missing quotes which resulted in invalid configuration.

### DIFF
--- a/docs/wsl.md
+++ b/docs/wsl.md
@@ -22,7 +22,7 @@ _Inside your WSL installation_, run the following command to set GCM as the Git
 credential helper:
 
 ```shell
-git config --global credential.helper /mnt/c/Program\ Files/Git/mingw64/libexec/git-core/git-credential-manager-core.exe
+git config --global credential.helper "/mnt/c/Program\ Files/Git/mingw64/libexec/git-core/git-credential-manager-core.exe"
 ```
 
 If you intend to use Azure DevOps you must _also_ set the following Git


derrickstolee:
> One of the commands for WSL configuration was missing quotes which resulted in invalid configuration.

The space in "Program Files" is guarded by a slash (`\`), so I don't see how this was a problem before.

However, using quotes is more idiomatic.

vtbassmatt:
With the `"` I guess we don't still need the `\` before the space?

nimatt:
Actually both are required. When using only \ it does not get persisted in the git config

```
$ git config --global credential.helper /mnt/c/Program\ Files/Git/mingw64/libexec/git-core/git-credential-manager-core.exe
$ git config --global -l
credential.helper=/mnt/c/Program Files/Git/mingw64/libexec/git-core/git-credential-manager-core.exe
...
$ git fetch
/mnt/c/Program Files/Git/mingw64/libexec/git-core/git-credential-manager-core.exe get: 1: /mnt/c/Program: not found
...
```

Using only " will have the same result
```
$ git config --global credential.helper "/mnt/c/Program Files/Git/mingw64/libexec/git-core/git-credential-manager-core.exe"
$ git config --global -l
credential.helper=/mnt/c/Program Files/Git/mingw64/libexec/git-core/git-credential-manager-core.exe
...
$ git fetch
/mnt/c/Program Files/Git/mingw64/libexec/git-core/git-credential-manager-core.exe get: 1: /mnt/c/Program: not found
...
```

Only when using both we end up with the \ in the config and it starts working

```
$ git config --global credential.helper "/mnt/c/Program\ Files/Git/mingw64/libexec/git-core/git-credential-manager-core.exe"
$ git config --global -l
credential.helper=/mnt/c/Program\ Files/Git/mingw64/libexec/git-core/git-credential-manager-core.exe
...
```

PR #499, opened 2021-10-19
Add basic support for generating code coverage reports

mminns:
**Problem**

I'd like to refactor parts of the Bitbucket code, but I'm uncomfortable about the level of security provided by the automated testing to avoid regressions.

I'd like to expand the testing, so I'd like to know what is currently tested or not.

**Solution**

This adds a first pass at adding basic code coverage metrics and reporting. 
They have to be run manually, they are not currently wired up to CI etc, I'd suggest doing that in a follow up.

### DIFF
--- a/.code-coverage/coverlet.settings.xml
+++ b/.code-coverage/coverlet.settings.xml
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<RunSettings>
+  <DataCollectionRunSettings>
+    <DataCollectors>
+      <DataCollector friendlyName="XPlat code coverage">
+        <Configuration>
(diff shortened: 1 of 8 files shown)


dscho:
Could this also be integrated into the CI builds?

mminns:
Yes, I don't see why not. 

I'm not as familiar with the CI setup as I probably should be to understand where it would fit, where you would see the results etc.

I had assumed I'd do that as a follow up if the general idea was agreed.

mjcheetham (approved):
This looks great @mminns! Thanks for contributing this 😁

mminns:
👍 Thanks

stranik38 (commented):
- [ ]

PR #517, opened 2021-11-03
Set `credential.provider` after successful auto detection probe

### LINKED ISSUE #507: Perform host provider auto-detection only once per host
This issue is a follow up on https://github.com/microsoft/Git-Credential-Manager-Core/issues/492#issuecomment-947491034

Can the credential manager remember auto-detection results?
With the current implementation, we pay this auto-detection tax every time we `git fetch` even though the Windows' credentials manager has proper authentication information available.

In terms of implementation, maybe the credential manager could do the equivalent of
```
$ git config credential.example.com.provider <result of auto-detection>
```

Thanks

mjcheetham:
After we've detected the host provider from an auto-detection network probe, set the `credential.provider` setting so we can avoid the expensive operation in the future. This greatly improves the performance of subsequent credential requests.

If we fail to set this then warn the user and ask them to set the configuration manually.

Fixes #507

### DIFF
--- a/docs/autodetect.md
+++ b/docs/autodetect.md
@@ -31,6 +31,11 @@ hostname can be made. Only one HTTP `HEAD` call is made per credential request
 received by Git. To avoid this network call, please [explicitly configure](#explicit-configuration)
 the host provider for your self-hosted instance.
 
+After a successful detection of the host provider, GCM will automatically set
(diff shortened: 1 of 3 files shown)


dscho (approved):
Looks good to me!

ldennington (approved):
🚀

derrickstolee (approved):

PR #518, opened 2021-11-05
Issue/218

mminns:
### Problem

The current implementation of OAuth2 support for Bitbucket.org in GCM assumes any stored access_token is likely to have expired.

This is based on the _outdated_ fact that Bitbucket.org access_tokens expire after 1 hour and the assumption that users are running remote Git commands at a frequency less than once per 1 hr.

In more detail the current process relies on using a much longer living refresh_token to refresh the access_token before returning the new value to Git. This effectively means

- the access_token returned to Git is _never_ the stored version
- for Bitbucket.org/OAuth2 GCM will run an OAuth2 refresh flow for every GET request

This means expired access_tokens are _never_ returned to Git and therefore the Git process should _never_ fail due to an expired access_token.

This overhead in GCM, running the refresh flow, should be more efficient than allowing Git to GET expired credentials from GCM, try to use the, fail, request ERASE from GCM, then repeat with a fresh GCM to get fresh credentials when the user is running remote  Git commands at a frequency less than once per 1 hr.

But ...

If the user is running remote Git commands at a frequency of more than once per 1hr they are incurring the OAuth refresh flow overhead everytime when they already have a valid access_token.

And ...

Bitbucket.org has revised the lifetime of access_tokens, increasing it from 1hr to 2 hrs. https://community.developer.atlassian.com/t/oauth2-token-lifetime-expiration-clarification/22136

This changes the performance balance. its now more likely GCM is ignoring valid access_tokens and incurring the OAuth2 refresh flow overhead 

### Solution

Flip the assumption around. 

Now assume that stored access_tokens are possibly still valid, verify if they are still valid and if they are return them directly to Git.

GCM does not store lifespan metadata about access_tokens, so there is no simple check to determine if they have expired, as such it is necessary to do. simple REST API call to Bitbucket.org using the access_token for authentication.

If that works, the access_token is assumed to be valid and is returned to Git.

If that fails GCM will try to run the OAuth refresh flow and generate and store a new access_token and return it to Git.

If that fails GCM will delete the refresh_token and return nothing to Git.Git's process will fail and when Git next requests to GET the credentials a new interactive process will be run by GCM to generate new credentials.

There will still be an overhead due to the verification of stored access_tokens but this requires fewer HTTP calls than the full refresh flow so will be quicker.

This verification is only applied to Bitbucket.org credentials, not to Bitbucket DC credentials.

It is possible that if the user is carrying our remote Git commands less frequently than once every 2 hours that it is better to keep the old process, of always running the OAuth refresh flow, in this case the configuration options `GCM_BITBUCKET_ALWAYS_REFRESH_CREDENTIALS`/`credential.bitbucketAlwaysRefreshCredentials` can be used to enable that process. **Be aware** for Basic Auth this means the user has to re-enter credentials everytime.

### Testing

I'm running this manually at the moment.

I've refactored some code and made use of the new code coverage tools to significantly increase the levels of test coverage for Bitbucket code.

### DIFF
--- a/src/shared/Atlassian.Bitbucket.Tests/BitbucketAuthenticationTest.cs
+++ b/src/shared/Atlassian.Bitbucket.Tests/BitbucketAuthenticationTest.cs
@@ -1,0 +1,44 @@
+using System;
+using System.Threading.Tasks;
+using GitCredentialManager.Tests.Objects;
+using Xunit;
+
+namespace Atlassian.Bitbucket.Tests
+{
+    public class BitbucketAuthenticationTest
+    {
+        [Theory]
+        [InlineData("jsquire", "password")]
+        public async Task BitbucketAuthentication_GetBasicCredentialsAsync_SucceedsAfterUserInput(string username, string password)
+        {
+            var context = new TestCommandContext();
+            context.Terminal.Prompts["Username"] = username;
+            context.Terminal.SecretPrompts["Password"] = password;
+            System.Uri targetUri = null;
+
+            var bitbucketAuthentication = new BitbucketAuthentication(context);
+
+            var result = await bitbucketAuthentication.GetBasicCredentialsAsync(targetUri, username);
+
+            Assert.NotNull(result);
+            Assert.Equal(username, result.Account);
+            Assert.Equal(password, result.Password);
+        }
+
+        [Theory]
+        [InlineData("jsquire", "password")]
+        public async Task BitbucketAuthentication_ShowOAuthRequiredPromptAsync_SucceedsAfterUserInput(string username, string password)
+        {
+            var context = new TestCommandContext();
+            context.Terminal.Prompts["Press enter to continue..."] = " ";
+
+            var bitbucketAuthentication = new BitbucketAuthentication(context);
+
+            var result = await bitbucketAuthentication.ShowOAuthRequiredPromptAsync();
+
+            Assert.True(result);
+            Assert.Equal($"Your account has two-factor authentication enabled.{Environment.NewLine}" +
+                                           $"To continue you must complete authentication in your web browser.{Environment.NewLine}", context.Terminal.Messages[0].Item1);
+        }
+    }
+}
--- a/src/shared/Atlassian.Bitbucket.Tests/BitbucketAuthenticationTest.cs
+++ b/src/shared/Atlassian.Bitbucket.Tests/BitbucketAuthenticationTest.cs
@@ -25,9 +25,8 @@ namespace Atlassian.Bitbucket.Tests
             Assert.Equal(password, result.Password);
         }
 
-        [Theory]
-        [InlineData("jsquire", "password")]
-        public async Task BitbucketAuthentication_ShowOAuthRequiredPromptAsync_SucceedsAfterUserInput(string username, string password)
+        [Fact]
+        public async Task BitbucketAuthentication_ShowOAuthRequiredPromptAsync_SucceedsAfterUserInput()
         {
             var context = new TestCommandContext();
             context.Terminal.Prompts["Press enter to continue..."] = " ";
--- a/src/shared/Atlassian.Bitbucket.Tests/BitbucketHostProviderTest.cs
+++ b/src/shared/Atlassian.Bitbucket.Tests/BitbucketHostProviderTest.cs
@@ -56,7 +56,7 @@ namespace Atlassian.Bitbucket.Tests
         {
             InputArguments input = null;
             var provider = new BitbucketHostProvider(new TestCommandContext());
-            Assert.Equal(false, provider.IsSupported(input));
+            Assert.False(provider.IsSupported(input));
         }
 
         [Fact]
@@ -64,7 +64,7 @@ namespace Atlassian.Bitbucket.Tests
         {
             HttpResponseMessage httpResponseMessage = null;
             var provider = new BitbucketHostProvider(new TestCommandContext());
-            Assert.Equal(false, provider.IsSupported(httpResponseMessage));
+            Assert.False(provider.IsSupported(httpResponseMessage));
         }
 
         [Theory]
--- a/src/shared/Atlassian.Bitbucket.Tests/BitbucketHostProviderTest.cs
+++ b/src/shared/Atlassian.Bitbucket.Tests/BitbucketHostProviderTest.cs
@@ -361,11 +361,11 @@ namespace Atlassian.Bitbucket.Tests
 
             var provider = new BitbucketHostProvider(context, bitbucketAuthentication.Object, bitbucketApi.Object);
 
-            Assert.Equal(context.CredentialStore.Count, 0);
+            Assert.Equal(0, context.CredentialStore.Count);
 
             await provider.StoreCredentialAsync(input);
 
-            Assert.Equal(context.CredentialStore.Count, 1);
+            Assert.Equal(1, context.CredentialStore.Count);
         }
         
         [Theory]
@@ -380,11 +380,11 @@ namespace Atlassian.Bitbucket.Tests
 
             var provider = new BitbucketHostProvider(context, bitbucketAuthentication.Object, bitbucketApi.Object);
 
-            Assert.Equal(context.CredentialStore.Count, 1);
+            Assert.Equal(1, context.CredentialStore.Count);
 
             await provider.EraseCredentialAsync(input);
 
-            Assert.Equal(context.CredentialStore.Count, 0);
+            Assert.Equal(0, context.CredentialStore.Count);
         }
 
         #endregion
--- a/src/shared/Atlassian.Bitbucket.Tests/BitbucketHostProviderTest.cs
+++ b/src/shared/Atlassian.Bitbucket.Tests/BitbucketHostProviderTest.cs
@@ -12,22 +12,27 @@ namespace Atlassian.Bitbucket.Tests
 {
     public class BitbucketHostProviderTest
     {
+        #region Tests
+
         private const string MOCK_ACCESS_TOKEN = "at-0987654321";
         private const string MOCK_REFRESH_TOKEN = "rt-1234567809";
+        private const string BITBUCKET_DOT_ORG_HOST = "bitbucket.org";
+        private const string DC_SERVER_HOST = "example.com";
         private Mock<IBitbucketAuthentication> bitbucketAuthentication = new Mock<IBitbucketAuthentication>(MockBehavior.Strict);
         private Mock<IBitbucketRestApi> bitbucketApi = new Mock<IBitbucketRestApi>(MockBehavior.Strict);
 
         [Theory]
+        [InlineData("https", null, false)]
         // We report that we support unencrypted HTTP here so that we can fail and
         // show a helpful error message in the call to `GenerateCredentialAsync` instead.
-        [InlineData("http", "bitbucket.org", true)]
-        [InlineData("ssh", "bitbucket.org", false)]
-        [InlineData("https", "bitbucket.org", true)]
+        [InlineData("http", BITBUCKET_DOT_ORG_HOST, true)]
+        [InlineData("ssh", BITBUCKET_DOT_ORG_HOST, false)]
+        [InlineData("https", BITBUCKET_DOT_ORG_HOST, true)]
         [InlineData("https", "api.bitbucket.org", true)] // Currently does support sub domains.
 
         [InlineData("https", "bitbucket.ogg", false)] // No support of phony similar tld.
         [InlineData("https", "bitbucket.com", false)] // No support of wrong tld.
-        [InlineData("https", "example.com", false)] // No support of non bitbucket domains.
+        [InlineData("https", DC_SERVER_HOST, false)] // No support of non bitbucket domains.
 
         [InlineData("http", "bitbucket.my-company-server.com", false)]  // Currently no support for named on-premise instances
         [InlineData("https", "my-company-server.com", false)]
@@ -46,6 +51,22 @@ namespace Atlassian.Bitbucket.Tests
             Assert.Equal(expected, provider.IsSupported(input));
         }
 
+        [Fact]
+        public void BitbucketHostProvider_IsSupported_FailsForNullInput()
+        {
+            InputArguments input = null;
+            var provider = new BitbucketHostProvider(new TestCommandContext());
+            Assert.Equal(false, provider.IsSupported(input));
+        }
+
+        [Fact]
+        public void BitbucketHostProvider_IsSupported_FailsForNullHttpResponseMessage()
+        {
+            HttpResponseMessage httpResponseMessage = null;
+            var provider = new BitbucketHostProvider(new TestCommandContext());
+            Assert.Equal(false, provider.IsSupported(httpResponseMessage));
+        }
+
         [Theory]
         [InlineData("X-AREQUESTID", "123456789", true)] // only the specific header is acceptable
         [InlineData("X-REQUESTID", "123456789", false)]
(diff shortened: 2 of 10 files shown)


src/shared/Atlassian.Bitbucket.Tests/BitbucketOauth2ClientTest.cs
```diff
@@ -0,0 +1,144 @@
...
+        {
+            Assert.NotNull(result);
+            IEnumerable<char> access_token = null;
+            Assert.Equal(access_token, result.AccessToken);
+            IEnumerable<char> refresh_token = null;
+            Assert.Equal(refresh_token, result.RefreshToken);
+            IEnumerable<char> tokenType = null;
+            Assert.Equal(tokenType, result.TokenType);
+            Assert.Equal(null, result.Scopes);
+            //Assert.Equal(pkceCodeVerifier, result.ExpiresIn);
```
mjcheetham, line 84: Looks like this was left over from earlier :)

src/shared/Atlassian.Bitbucket.Tests/BitbucketHostProviderTest.cs
```diff
@@ -62,109 +83,133 @@ public void BitbucketHostProvider_IsSupported_HttpResponseMessage(string header,
...
             MockRemoteValidRefreshToken();
 
             var provider = new BitbucketHostProvider(context, bitbucketAuthentication.Object, bitbucketApi.Object);
 
-            if (userEntersCredentials.HasValue && !userEntersCredentials.Value)
-            {
-                Assert.ThrowsAsync<Exception>(async () => await provider.GetCredentialAsync(input));
-                return;
-            }
+            //if (userEntersCredentials.HasValue && !userEntersCredentials.Value)
```
mjcheetham, line 193: Remove commented out code please :)

src/shared/Atlassian.Bitbucket.Tests/BitbucketHostProviderTest.cs
```diff
@@ -174,30 +219,77 @@ public void BitbucketHostProvider_GetCredentialAsync_ForOAuth(string protocol, s
                 context.Environment.Variables.Add(BitbucketConstants.EnvironmentVariables.AuthenticationModes, preconfiguredAuthModes);
             }
 
-            MockUserEnteredBasicCredentials(bitbucketAuthentication, input, password);
-            MockValidRemoteAccount(bitbucketApi, input, password, true);
-            MockRemoteValidRefreshToken();
+            MockUserEntersValidBasicCredentials(bitbucketAuthentication, input, password);
+            MockRemoteBasicAuthAccountIsValidRequires2FA(bitbucketApi, input, password);
+            bitbucketAuthentication.Setup(m => m.ShowOAuthRequiredPromptAsync()).ReturnsAsync(true);
+            //MockRemoteValidRefreshToken();
```
mjcheetham, line 225: And here

src/shared/Atlassian.Bitbucket/BitbucketHostProvider.cs
```diff
@@ -83,16 +92,56 @@ public async Task<ICredential> GetCredentialAsync(InputArguments input)
...
+
+            var credentials = _context.CredentialStore.Get(credentialService, input.UserName);
+
+            if(credentials == null)
+            {
+                _context.Trace.WriteLine($"    Found none");
+                return null;
+            }
+
+            _context.Trace.WriteLine($"    Found credentials: {credentials.Account}/**********");
```
mjcheetham, line 119: If you wanted you could add the `credentials.Password` field here, and use `Trace.WriteLineSecrets` appropriately to mask the password/token.

src/shared/Atlassian.Bitbucket/BitbucketHostProvider.cs
```diff
@@ -101,42 +150,28 @@ public async Task<ICredential> GetCredentialAsync(InputArguments input)
...
 
-                    // Either we have an existing credential (user/pass OR some form of token [PAT or AT]),
-                    // or we have a freshly captured user/pass. Regardless, we must check if these credentials
-                    // pass and two-factor requirement on the account.
-                    _context.Trace.WriteLine("Checking if two-factor requirements for stored credentials...");
+                    var basicCredentials = await GetBasicCredentialsInteractive(input, authModes);
 
-                    bool requires2Fa = await RequiresTwoFactorAuthenticationAsync(credential, authModes);
-                    if (!requires2Fa)
+                    if(await ValidateCredentialsWork(input, basicCredentials, authModes))
```
mjcheetham, line 157: The prevailing style in GCM is a space between the `if` keyword and the `()`, i.e, `if (true)` and not `if(true)`.
Please update instances in this PR

src/shared/Atlassian.Bitbucket/BitbucketHostProvider.cs
```diff
@@ -145,6 +180,7 @@ public async Task<ICredential> GetCredentialAsync(InputArguments input)
             }
             else
             {
+                _context.Trace.WriteLine($"    Found refresh token: {refreshToken} ");
                 // TODO: should we try and compute if the AT has expired and use it?
```
mjcheetham, line 184: Can now drop this TODO comment, and the one below about "the AT is likely expired" since we're now actually checking and using the AT!

src/shared/Atlassian.Bitbucket/BitbucketHostProvider.cs
```diff
@@ -328,6 +411,59 @@ private async Task<bool> RequiresTwoFactorAuthenticationAsync(ICredential creden
...
+            }
+
+            var targetUri = input.GetRemoteUri();
+            _context.Trace.WriteLine($"Validate credentials ({credentials.Account}/**********) are fresh for {targetUri} ...");
+
+            if(!IsBitbucketOrg(targetUri))
+            {
+                // TODO Validate DC/Server credentials before returning them to Git
+                // Currently credentials for DC/Server are not checked by GCM.
+                // Instead the validation relies on Git to try and fail with the creneitals and then request GCM to erase them
```
mjcheetham, line 428: Typo "creneitals"

src/shared/Atlassian.Bitbucket/BitbucketHostProvider.cs
```diff
@@ -10,6 +10,8 @@ namespace Atlassian.Bitbucket
 {
     public class BitbucketHostProvider : IHostProvider
     {
+        public const bool USE_BEARER_TOKEN = true;
```
mjcheetham, line 13: Rather than introduce these constants, I think it'd be better to use `true` and `false` at the call sites, but use a named parameter. For example:

```csharp
_bitbucketApi.GetUserInformationAsync(null, accessToken, isBearerToken: true)
```

src/shared/Atlassian.Bitbucket/BitbucketHostProvider.cs
```diff
@@ -328,6 +411,59 @@ private async Task<bool> RequiresTwoFactorAuthenticationAsync(ICredential creden
             }
         }
 
+        private async Task<bool> ValidateCredentialsWork(InputArguments input, ICredential credentials, AuthenticationModes authModes) 
+        {
+            if(credentials == null)
+            {
+                return false;
+            }
+
```
mjcheetham, line 420: Could add a `TODO` here that ideally we'd also check if the credentials have expired based on some local metadata (once/if we get such metadata storage), and return `false` if they have. Just for future readers :)

mjcheetham (changes requested):
This looks great! Thanks :grin:
Just added a few style/typo nits.

mminns:
A few ;) 

I'll take a look.

src/shared/Atlassian.Bitbucket/BitbucketHostProvider.cs
```diff
@@ -83,16 +90,56 @@ public async Task<ICredential> GetCredentialAsync(InputArguments input)
...
+            if (credentials == null)
+            {
+                _context.Trace.WriteLine($"    Found none");
+                return null;
+            }
+
+            _context.Trace.WriteLineSecrets($"    Found credentials: {credentials.Account}/{{0}}", new object[] {credentials.Password});
+            
+            //check credentials are still valid
+            if (!await ValidateCredentialsWork(input, credentials, GetSupportedAuthenticationModes(targetUri))) {
```
mjcheetham, line 120:
```suggestion
            if (!await ValidateCredentialsWork(input, credentials, GetSupportedAuthenticationModes(targetUri)))
            {
```

src/shared/Atlassian.Bitbucket/BitbucketHostProvider.cs
```diff
@@ -83,16 +90,56 @@ public async Task<ICredential> GetCredentialAsync(InputArguments input)
...
+
+            var targetUri = input.GetRemoteUri();
+            string credentialService = GetServiceName(input);
+            _context.Trace.WriteLine($"Look for existing credentials under {credentialService} ...");
+
+            var credentials = _context.CredentialStore.Get(credentialService, input.UserName);
+
+            if (credentials == null)
+            {
+                _context.Trace.WriteLine($"    Found none");
```
mjcheetham, line 113:
```suggestion
                _context.Trace.WriteLine("Found no stored credential");
```

src/shared/Atlassian.Bitbucket/BitbucketHostProvider.cs
```diff
@@ -83,16 +90,56 @@ public async Task<ICredential> GetCredentialAsync(InputArguments input)
...
+                return null;
+            }
+
+            _context.Trace.WriteLineSecrets($"    Found credentials: {credentials.Account}/{{0}}", new object[] {credentials.Password});
```
mjcheetham, line 117:
```suggestion
            _context.Trace.WriteLineSecrets($"Found credentials: {credentials.Account}/{{0}}", new object[] {credentials.Password});
```

src/shared/Atlassian.Bitbucket/BitbucketHostProvider.cs
```diff
@@ -83,16 +90,56 @@ public async Task<ICredential> GetCredentialAsync(InputArguments input)
...
+            
+            //check credentials are still valid
```
mjcheetham, line 119:
```suggestion
            // Check credentials are still valid
```

src/shared/Atlassian.Bitbucket/BitbucketHostProvider.cs
```diff
@@ -83,16 +90,56 @@ public async Task<ICredential> GetCredentialAsync(InputArguments input)
...
-            string refreshTokenService = GetRefreshTokenServiceName(input);
+            var refreshTokenService = GetRefreshTokenServiceName(input);
 
             AuthenticationModes authModes = GetSupportedAuthenticationModes(targetUri);
 
             _context.Trace.WriteLine("Checking for refresh token...");
             ICredential refreshToken = SupportsOAuth(authModes) ? _context.CredentialStore.Get(refreshTokenService, input.UserName) : null;
             if (refreshToken is null)
             {
+                _context.Trace.WriteLine($"    Found none");
```
mjcheetham, line 142:
```suggestion
                _context.Trace.WriteLine("Found existing refresh token");
```

src/shared/Atlassian.Bitbucket/BitbucketHostProvider.cs
```diff
@@ -83,16 +90,56 @@ public async Task<ICredential> GetCredentialAsync(InputArguments input)
...
                 // There is no refresh token either because this is a non-2FA enabled account (where OAuth is not
                 // required), or because we previously erased the RT.
```
mjcheetham, line 145: GitHub doesn't let me leave a comment or suggestion on a "context" line... (grr)
Please can you remove this code (one line below) that is not used:

```csharp
// Check for the presence of a credential in the store
string credentialService = GetServiceName(input);
```

src/shared/Atlassian.Bitbucket/BitbucketHostProvider.cs
```diff
@@ -145,15 +178,14 @@ public async Task<ICredential> GetCredentialAsync(InputArguments input)
             }
             else
             {
-                // TODO: should we try and compute if the AT has expired and use it?
-                // This needs support from the credential store to record the expiry time!
+                _context.Trace.WriteLine($"    Found refresh token: {refreshToken} ");
```
mjcheetham, line 181:
```suggestion
                _context.Trace.WriteLine($"Found refresh token: {refreshToken} ");
```

src/shared/Atlassian.Bitbucket/BitbucketHostProvider.cs
```diff
@@ -145,15 +178,14 @@ public async Task<ICredential> GetCredentialAsync(InputArguments input)
...
 
                 // It's very likely that any access token expired between the last time we used/stored it.
                 // To ensure the AT is as 'fresh' as it can be, always first try to use the refresh token
                 // (which lives longer) to create a new AT (and possibly also a new RT).
```
mjcheetham, line 185: I don't think this comment is correct now, since if we're here then we've already found no such AT, or found an AT that doesn't work.

```suggestion
```

src/shared/Atlassian.Bitbucket.Tests/BitbucketAuthenticationTest.cs
```diff
@@ -0,0 +1,44 @@
...
+            var result = await bitbucketAuthentication.GetBasicCredentialsAsync(targetUri, username);
+
+            Assert.NotNull(result);
+            Assert.Equal(username, result.Account);
+            Assert.Equal(password, result.Password);
+        }
+
+        [Theory]
+        [InlineData("jsquire", "password")]
+        public async Task BitbucketAuthentication_ShowOAuthRequiredPromptAsync_SucceedsAfterUserInput(string username, string password)
```
mjcheetham, line 30: The inline data is not being used (and triggering a warning in the CI), so let's convert this `Theory` to a `Fact` and drop the unused params.

```suggestion
        [Fact]
        public async Task BitbucketAuthentication_ShowOAuthRequiredPromptAsync_SucceedsAfterUserInput()
```

src/shared/Atlassian.Bitbucket.Tests/BitbucketHostProviderTest.cs
```diff
@@ -46,6 +51,22 @@ public void BitbucketHostProvider_IsSupported(string protocol, string host, bool
             Assert.Equal(expected, provider.IsSupported(input));
         }
 
+        [Fact]
+        public void BitbucketHostProvider_IsSupported_FailsForNullInput()
+        {
+            InputArguments input = null;
+            var provider = new BitbucketHostProvider(new TestCommandContext());
+            Assert.Equal(false, provider.IsSupported(input));
```
mjcheetham, line 59: Xunit CI analysers are complaining about `Assert.Equal(false, X)`.. we should use `Assert.False(X)`:

```suggestion
            Assert.False(provider.IsSupported(input));
```

src/shared/Atlassian.Bitbucket.Tests/BitbucketHostProviderTest.cs
```diff
@@ -46,6 +51,22 @@ public void BitbucketHostProvider_IsSupported(string protocol, string host, bool
...
+        }
+
+        [Fact]
+        public void BitbucketHostProvider_IsSupported_FailsForNullHttpResponseMessage()
+        {
+            HttpResponseMessage httpResponseMessage = null;
+            var provider = new BitbucketHostProvider(new TestCommandContext());
+            Assert.Equal(false, provider.IsSupported(httpResponseMessage));
```
mjcheetham, line 67: Same here..

```suggestion
            Assert.False(provider.IsSupported(httpResponseMessage));
```

src/shared/Atlassian.Bitbucket.Tests/BitbucketHostProviderTest.cs
```diff
@@ -229,6 +315,81 @@ public void BitbucketHostProvider_GetSupportedAuthenticationModes(string uriStri
...
+        [InlineData("https", DC_SERVER_HOST, "jsquire")]
+        public async Task BitbucketHostProvider_StoreCredentialAsync(string protocol, string host, string username)
+        {
+            var input = MockInput(protocol, host, username);
+
+            var context = new TestCommandContext();
+
+            var provider = new BitbucketHostProvider(context, bitbucketAuthentication.Object, bitbucketApi.Object);
+
+            Assert.Equal(context.CredentialStore.Count, 0);
```
mjcheetham, line 364: The xunit `Assert.Equal` method's signature is `(expected, actual)`, so we should pass the `0` as the first arg (picked up by CI analysers).

```suggestion
            Assert.Equal(0, context.CredentialStore.Count);
```

src/shared/Atlassian.Bitbucket.Tests/BitbucketHostProviderTest.cs
```diff
@@ -229,6 +315,81 @@ public void BitbucketHostProvider_GetSupportedAuthenticationModes(string uriStri
...
+
+            await provider.StoreCredentialAsync(input);
+
+            Assert.Equal(context.CredentialStore.Count, 1);
```
mjcheetham, line 368: Same here..

```suggestion
            Assert.Equal(1, context.CredentialStore.Count);
```

src/shared/Atlassian.Bitbucket.Tests/BitbucketHostProviderTest.cs
```diff
@@ -229,6 +315,81 @@ public void BitbucketHostProvider_GetSupportedAuthenticationModes(string uriStri
...
+        {
+            var input = MockInput(protocol, host, username);
+
+            var context = new TestCommandContext();
+
+            MockStoredAccount(context, input, password);
+
+            var provider = new BitbucketHostProvider(context, bitbucketAuthentication.Object, bitbucketApi.Object);
+
+            Assert.Equal(context.CredentialStore.Count, 1);
```
mjcheetham, line 383:
```suggestion
            Assert.Equal(1, context.CredentialStore.Count);
```

src/shared/Atlassian.Bitbucket.Tests/BitbucketHostProviderTest.cs
```diff
@@ -229,6 +315,81 @@ public void BitbucketHostProvider_GetSupportedAuthenticationModes(string uriStri
...
+
+            await provider.EraseCredentialAsync(input);
+
+            Assert.Equal(context.CredentialStore.Count, 0);
```
mjcheetham, line 387:
```suggestion
            Assert.Equal(0, context.CredentialStore.Count);
```

src/shared/Atlassian.Bitbucket.Tests/BitbucketOauth2ClientTest.cs
```diff
@@ -0,0 +1,142 @@
...
+        private void VerifyOAuth2TokenResult(OAuth2TokenResult result)
+        {
+            Assert.NotNull(result);
+            IEnumerable<char> access_token = null;
+            Assert.Equal(access_token, result.AccessToken);
+            IEnumerable<char> refresh_token = null;
+            Assert.Equal(refresh_token, result.RefreshToken);
+            IEnumerable<char> tokenType = null;
+            Assert.Equal(tokenType, result.TokenType);
+            Assert.Equal(null, result.Scopes);
```
mjcheetham, line 82: Xunit prefers `Assert.Null(X)` over `Assert.Equal(null, X)`:

```suggestion
            Assert.Null(result.Scopes);
```

mjcheetham (changes requested):
Thanks again! Sorry for a few more minor changes.. mainly to make the style/xunit analysers happy and not spam us in CI builds.

mminns:
No problem.

For future reference do you use any particular code convention/formatting tool or configuration? Also where is the CI spamming?

mminns:
I believe I've addressed all the points, but let me know if I'm missing anything.

mjcheetham (approved):
Awesome! Thanks again @mminns

PR #520, opened 2021-11-08
Remove winget workflow

ldennington:
We have decided to remove GCM Core's winget publishing workflow for the
following reasons:

1. We want to encourage Windows users to install the bundled GCM Core that
comes with Git for Windows
2. GCM Core has 2 installers with the same arch and type (user and machine),
which wingetcreate doesn't support
3. Removing this option will decrease the complexity of our GCM Core's
release matrix
4. If users need to install GCM Core separately from Git for Windows, they
can download the appropriate installer from the releases page

Thus we are removing the workflow with this change.

Note: We are unable to deprecate the packages we have already released due
to lack of `winget` support for this scenario. Tracking here:

https://github.com/github/git-fundamentals/issues/756


derrickstolee (approved):
Love to see it.

image

PR #533, opened 2021-11-16
Fix release builds

mjcheetham:
Recent changes to the release build agent pool configuration means we have to update a few parts of our release build:

1. Use the new agent pool
2. Update NuGet packages and Azure DevOps tasks
3. Explicit authentication for NuGet feeds


dscho (approved):
Looks good!

PR #538, opened 2021-11-17
Show GitHub terminal prompts when no GUI session

### LINKED ISSUE #453: Allow terminal-based prompting on Windows over SSH
When using GCM over an SSH terminal connection (no GUI) the UI helpers are launched and therefore it is not possible to sign-in from the connecting terminal!

GCM should detect the presence (or lack thereof) of a GUI and show a UI helper or terminal prompt accordingly.

The Microsoft Authentication stack correctly responds to terminal/GUI availability, but the GitHub and Bitbucket stacks do not.

mjcheetham:
Show the terminal/TTY based prompts when there is no desktop/GUI sessions present for GitHub.

Currently the Bitbucket and Microsoft auth stacks correctly detect the lack of a desktop session and use terminal prompts - GitHub does not.

Without this change, SSH-ing in to a machine we'd still try and show a UI prompt where there is no way to interact with it!

Fixes #453

### DIFF
--- a/src/shared/GitHub/GitHubAuthentication.cs
+++ b/src/shared/GitHub/GitHubAuthentication.cs
@@ -79,7 +79,7 @@ namespace GitHub
                 throw new ArgumentException(@$"Must specify at least one {nameof(AuthenticationModes)}", nameof(modes));
             }
 
-            if (TryFindHelperExecutablePath(out string helperPath))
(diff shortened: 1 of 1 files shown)


dscho (approved):

ldennington (approved):
Good fix!

PR #544, opened 2021-11-18
workflows: add CodeQL

ldennington:
Adding CodeQL to new org/repo. In #543 we removed this workflow because it seemed to have broken during our transition from the microsoft org to the GitCredentialManager org. Re-adding with the assumption that starting from a clean slate will fix the issues we were seeing.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,40 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ main, release ]
+  pull_request:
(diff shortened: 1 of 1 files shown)


ldennington:
Note: The `analysis not found` result of the Code scanning results check is expected here. See [these docs](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/setting-up-code-scanning-for-a-repository#understanding-the-pull-request-checks) for details.

derrickstolee (approved):

PR #551, opened 2021-11-23
Rename entry executable to drop "-core" suffix

mjcheetham:
**UPDATE**

Rename the main entry executable from `git-credential-manager-core(.exe)` to simply `git-credential-manager(.exe)`, now that the older GCM4W has been removed from the Git for Windows project as an option (and the GCMC project has been renamed).

We do **not** rename the Debian package ID, nor the Homebrew Cask name yet as this will require more thought forr migration.

To help with migration somewhat, create symlinks and shim/copy-executables for the original executable name "git-credential-manager-core(.exe)" for consumers who have not updated to the new version.
    
We detect if the consumer is launching us via the "-core" symlink or executable shim by consulting the platform-native APIs to get the original "argv[0]". All of the .NET APIs sadly don't give us the real "argv[0]", so we need to use native APIs...
    
If we detect use of the old name, print a warning that the user should update their configuration, and a help link for more information.

Link to rendered "rename" doc: https://github.com/mjcheetham/git-credential-manager/blob/exec-rename/docs/rename.md


.github/workflows/release-homebrew.yaml
```diff
@@ -14,5 +14,5 @@ jobs:
         tap: microsoft/git
         name: git-credential-manager-core
         type: cask
-        releaseAsset: gcmcore-osx-(.*)\.pkg
+        releaseAsset: gcm-osx-(.*)\.pkg
```
dscho, line 17: This change belongs in the third commit, not the first, correct?

vtbassmatt:
Is there a deprecation plan for the shims? It seems like a lot of hassle for low rewards if we have to carry those indefinitely. But if we can document their disappearance in, say, a year...

vtbassmatt (approved):

.github/workflows/release-homebrew.yaml
mjcheetham: Yep! Rebase fail

ldennington:
Do we have a way of knowing when VS takes a new mingit version?

.azure-pipelines/templates/osx/pack.signed/step5-dist.yml
```diff
@@ -21,7 +21,7 @@ steps:
       downloadPath: '$(Build.StagingDirectory)/symbols'
 
   # Skip notarization until we can preserve the hardened runtime bit and sign the .NET runtime bits
-  # Tracked: https://github.com/microsoft/Git-Credential-Manager-Core/issues/108
+  # Tracked: https://github.com/GitCredentialManager/git-credential-manager/issues/108
```
ldennington, line 24: Nice. Sorry I missed this in my original pass.

src/linux/Packaging.Linux/shim.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+echo >&2 "warning: git-credential-manager-core was renamed to git-credential-manager"
+echo >&2 "warning: please update your credential.helper configuration"
```
ldennington, line 4: Do we want to include the command they should run to update?

src/windows/Installer.Windows/Setup.iss
```diff
@@ -106,8 +106,8 @@ Source: "{#PayloadDir}\Atlassian.Bitbucket.dll";                        DestDir:
 Source: "{#PayloadDir}\Atlassian.Bitbucket.UI.exe";                     DestDir: "{app}"; Flags: ignoreversion
 Source: "{#PayloadDir}\Atlassian.Bitbucket.UI.exe.config";              DestDir: "{app}"; Flags: ignoreversion
 Source: "{#PayloadDir}\Atlassian.Bitbucket.UI.Shared.dll";              DestDir: "{app}"; Flags: ignoreversion
-Source: "{#PayloadDir}\git-credential-manager-core.exe";                DestDir: "{app}"; Flags: ignoreversion
-Source: "{#PayloadDir}\git-credential-manager-core.exe.config";         DestDir: "{app}"; Flags: ignoreversion
+Source: "{#PayloadDir}\git-credential-manager.exe";                     DestDir: "{app}"; Flags: ignoreversion
+Source: "{#PayloadDir}\git-credential-manager.exe.config";              DestDir: "{app}"; Flags: ignoreversion
```
ldennington, line 110: Were the extra whitespaces after the semicolons intentional?

ldennington (approved):
Exciting times! 🚀

src/windows/Installer.Windows/Setup.iss
derrickstolee: It seems that we are using spaces, not tabs to align the `DestDir:` bits, so that would be necessary to keep the style.

.azure-pipelines/templates/osx/pack.signed/step5-dist.yml
mjcheetham: No problem!

src/linux/Packaging.Linux/shim.sh
mjcheetham: I asked about this, but I don't think we can be sure of the exact setup they have to be able to synthesise the right command. Did they update the system config? user config? `-c`? Envar?

mjcheetham:
> Is there a deprecation plan for the shims? It seems like a lot of hassle for low rewards if we have to carry those indefinitely. But if we can document their disappearance in, say, a year...

That's a good point. I know @dscho was thinking of removing the Windows shim script from Git for Windows "at some future date".

Would you put the "this is temporary" message in the warnings too?

Regarding the rewards; I think this is really most important for Git for Windows users, and those who have installed outside of Homebrew (tarballs, macOS pkg, or Debian deb). Homebrew updates will do an uninstall first of the old one, so keeping clean.

mjcheetham:
> Do we have a way of knowing when VS takes a new mingit version?

Major releases certainly, and sometimes minor releases. It requires a full test pass on their side of things and is an expensive thing to do. They usually also update when we advise them to, for security reasons for example.

ldennington (approved):
OMG IT'S HAPPENING! 🍾

src/shared/Core/Interop/Windows/Native/Shell32.cs
```diff
@@ -0,0 +1,26 @@
...
+        /// in a way that is similar to the standard C run-time argv and argc values.
+        /// </summary>
+        /// <param name="lpCmdLine">
+        /// Pointer to a null-terminated Unicode string that contains the full command line.
+        /// If this parameter is an empty string the function returns the path to the current executable file.
+        /// </param>
+        /// <param name="pNumArgs">
+        /// Pointer to an int that receives the number of array elements returned, similar to argc.
+        /// </param>
+        /// <returns>A pointer to an array of LPWSTR values, similar to argv.</returns>
```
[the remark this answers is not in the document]
mjcheetham: They came directly from the Windows API docs online 😅

src/shared/Core/ApplicationBase.cs
```diff
@@ -84,31 +83,18 @@ public Task<int> RunAsync(string[] args)
...
             {
-                return Path.ChangeExtension(candidatePath, null);
+                return argv0;
             }
 
-            return candidatePath;
-#endif
+            return Path.GetFullPath(
+                Path.Combine(Environment.CurrentDirectory, argv0)
+            );
```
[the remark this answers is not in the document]
mjcheetham: It's changes like these (avoiding .NET and calling platform native APIs) that makes me wonder where the balance of "help" vs "hinder" .NET is for us... 🤔

src/shared/Core/Interop/Windows/Native/Shell32.cs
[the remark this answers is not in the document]
ldennington: Even better 😁

docs/rename.md
```diff
@@ -0,0 +1,166 @@
...
+
+In November 2021, _"Git Credential Manager Core"_ was [renamed][rename-pr] to
+simply _"Git Credential Manager"_, dropping the "Core" moniker. We announced the
+new name in a [GitHub blog post][rename-blog], along with the new home for the
+project in it's own [organization][gcm-org].
+
+![Git Credential Manager Core renamed](img/gcmcore-rename.png)
+
+At the time, the actual exectuable name was not updated and continued to be
+`git-credential-manager-core`. As of [VERSION][rename-ver], the executable has
```
mjcheetham, line 11: The `VERSION` string is a placeholder for when we make the first release and know the version number of GCM.

docs/rename.md
```diff
@@ -0,0 +1,166 @@
...
+
+It is recommended to update your Git configuration to use the new executable
+name as soon as possible to prevent any issues in the future.
+
+## How to update
+
+### Git for Windows
+
+If you are using GCM bundled with Git for Windows (recommended), you should make
+sure you have updated to at least version WINGIT_VERSION.
```
mjcheetham, line 38: ..similarly, the `WINGET_VERISON` will be updated once we know the version number of the Git for Windows release that ships GCM with the new name.

docs/rename.md
```diff
@@ -0,0 +1,166 @@
...
+```console
+warning: git-credential-manager-core was renamed to git-credential-manager
+warning: see https://aka.ms/gcm/rename for more information
+```
+
+Since the executable was renamed in VERSION, GCM has also included symlinks
+using the old name in order to ensure no one's setups would immediately break.
+
+These links will remain until _two_ major Git versions are released after GCM
+VERSION, _**at which point the symlinks will no longer be included**_.
```
mjcheetham, line 28: I think giving two major Git versions is a good runway. Thoughts? @ldennington @dscho

(My "major" of course I mean `y` in `x.y.z`.. not `x`!)

mjcheetham:
I'm going to wait until @dscho is back for their opinion on this solution before merging. We want to make sure this will be OK with Git for Windows.

docs/rename.md
dscho: Sure!

src/shared/Microsoft.AzureRepos.Tests/AzureReposHostProviderTests.cs
```diff
@@ -570,6 +570,22 @@ public async Task AzureReposHostProvider_UnconfigureAsync_UseHttpPathSet_Removes
...
+            context.Git.Configuration.System[HelperKey] = new List<string> {"manager"};
+            context.Git.Configuration.System[AzDevUseHttpPathKey] = new List<string> {"true"};
+
+            await provider.UnconfigureAsync(ConfigurationTarget.System);
+
+            Assert.True(context.Git.Configuration.System.TryGetValue(AzDevUseHttpPathKey, out IList<string> actualValues));
+            Assert.Single(actualValues);
+            Assert.Equal("true", actualValues[0]);
+        }
+
```
dscho, line 588: I'm not quite sure whether this test was added intentionally? I might be misinterpreting, but nothing in that test speaks "drop the -core suffix" to me... Likewise with `AzureReposHostProvider_UnconfigureAsync_User_Windows_UseHttpPathSetAndManagerHelper_RemovesEntry` below.

src/linux/Packaging.Linux/build.sh
```diff
@@ -235,6 +235,12 @@ if [ ! -f "$LINK_TO/git-credential-manager" ]; then
         "$LINK_TO/git-credential-manager" || exit 1
 fi
 
+# Create legacy symlink with older name
+if [ ! -f "$LINK_TO/git-credential-manager-core" ]; then
+    ln -s -r "$INSTALL_TO/git-credential-manager" \
+        "$LINK_TO/git-credential-manager-core" || exit 1
+fi
+
```
dscho, line 243: Do we have any corresponding "uninstall" script that needs an equivalent change?

src/shared/Core/Interop/Windows/Native/Shell32.cs
```diff
@@ -0,0 +1,26 @@
...
+        [DllImport("Shell32.dll", CallingConvention = CallingConvention.StdCall, CharSet = CharSet.Unicode, SetLastError = true)]
+        public static extern IntPtr CommandLineToArgvW(IntPtr lpCmdLine, out int pNumArgs);
```
dscho, line 24: Do we need to import `LocalFree()`, too? That's [what I call](https://github.com/git-for-windows/MINGW-packages/blob/764f4bab64a06db7883188e12fe1bb458baab7ad/mingw-w64-git/git-wrapper.c#L361) in [the Git wrapper](https://github.com/git-for-windows/git/wiki/Git-wrapper) after [splitting the command line](https://github.com/git-for-windows/MINGW-packages/blob/764f4bab64a06db7883188e12fe1bb458baab7ad/mingw-w64-git/git-wrapper.c#L277).

src/linux/Packaging.Linux/build.sh
mjcheetham: We don't provide an uninstall script for Linux

src/shared/Microsoft.AzureRepos.Tests/AzureReposHostProviderTests.cs
mjcheetham: These tests are to ensure that, on Windows, when uninstalling GCM (as installed from a standalone installer), we do NOT delete the `useHttpPath` setting that was placed in the `--system` config file by Git for Windows.

The new tests check to see if `helper = manager`, whereas the existing tests check we don't remove the `useHttpPath` setting when `helper = manager-core`.

src/shared/Core/Interop/Windows/Native/Shell32.cs
mjcheetham: Added!

Moises3445:
> **UPDATE**
> 
> Rename the main entry executable from `git-credential-manager-core(.exe)` to simply `git-credential-manager(.exe)`, now that the older GCM4W has been removed from the Git for Windows project as an option (and the GCMC project has been renamed).
> 
> We do **not** rename the Debian package ID, nor the Homebrew Cask name yet as this will require more thought forr migration.
> 
> To help with migration somewhat, create symlinks and shim/copy-executables for the original executable name "git-credential-manager-core(.exe)" for consumers who have not updated to the new version.
> 
> We detect if the consumer is launching us via the "-core" symlink or executable shim by consul

Favourjacobmudiaga:
Hello

PR #560, opened 2021-12-04
Fix setx command

cjsimon:


dscho (approved):
According to https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/setx, this change is correct.

mjcheetham (approved):
Thanks for spotting this!

PR #575, opened 2021-12-16
Fix bug in Azure Repos binding `list` command

mjcheetham:
There has been a behaviour change in the System.CommandLine library. Previously string arguments that were missing would be null, but now they are the empty string.

Update downlevel use of the `organization` parameter in `BindCmd` to also expect an empty/whitespace string, as well as null.


derrickstolee (approved):
Seem like easy fixes. Thanks!

ldennington (approved):

PR #579, opened 2021-12-21
Allow shortcut to perform OAuth from first prompt with Bitbucket

### LINKED ISSUE #427: Bitbucket authentication workflow should support OAuth only flow
**Which version of GCM Core are you using?**

From a terminal, run `git-credential-manager-core --version` and paste the output.
2.0.498+7ad55fb809 (macOS)

* [ ] Azure DevOps
* [ ] GitHub
* [ ] GitHub Enterprise
* [x] Bitbucket
* [ ] Other - please describe

* [x] Yes

**Expected behavior**

On `git clone https://bitbucket.org/orgname/reponame`, the workflow should go directly to the browser 2FA or, have a button on the dialog to use 2FA instead. That way, accounts that need a "Login with Google" don't need to unnecessarily create a "app password", which is not required in the case of a 2FA anyways.

**Actual behavior**

On `git clone https://bitbucket.org/orgname/reponame`, the Atlassian Login dialog comes up. This requires that a username/password be entered before the 2FA redirect in the browser is opened. 

**Logs**

Set the environment variables `GCM_TRACE=1` and `GIT_TRACE=1` and re-run your Git command. Review and redact any private information and attach the log.

mjcheetham:
Today if the user has 2FA enabled on their Atlassian account they need to either:

1. Enter their username/password, then perform the OAuth dance in a browser (potentially needing to re-enter their user/pass again), or
2. Set the [`credential.bitbucketAuthModes`](https://aka.ms/gcm/config#credentialbitbucketauthmodes) setting to `oauth`.

This change adds a new option to the initial credential prompt "Sign in with OAuth". Selecting this option allows the user to avoid typing their username/password in the dialog, and GCM proceeds directly to performing the OAuth dance in the browser!

Before|After
-|-
image|image

Fixes #427

cc: @mminns

### DIFF
--- a/src/shared/Atlassian.Bitbucket/BitbucketHostProvider.cs
+++ b/src/shared/Atlassian.Bitbucket/BitbucketHostProvider.cs
@@ -74,50 +74,46 @@ namespace Atlassian.Bitbucket
 
         public async Task<ICredential> GetCredentialAsync(InputArguments input)
         {
-            ValidateTargetUri(input);
-
-            return await GetStoredCredentials(input) ?? await GetRefreshedCredentials(input);
-        }
-
-        private static void ValidateTargetUri(InputArguments input)
-        {
-            // Compute the remote URI
-            Uri targetUri = input.GetRemoteUri();
-
             // We should not allow unencrypted communication and should inform the user
             if (StringComparer.OrdinalIgnoreCase.Equals(input.Protocol, "http")
-                && IsBitbucketOrg(targetUri))
+                && input.TryGetHostAndPort(out string host, out _) && IsBitbucketOrg(host))
             {
                 throw new Exception("Unencrypted HTTP is not supported for Bitbucket.org. Ensure the repository remote URL is using HTTPS.");
             }
+
+            Uri remoteUri = input.GetRemoteUri();
+
+            AuthenticationModes authModes = GetSupportedAuthenticationModes(remoteUri);
+
+            return await GetStoredCredentials(remoteUri, input.UserName, authModes) ??
+                   await GetRefreshedCredentials(remoteUri, input.UserName, authModes);
         }
 
-        private async Task<ICredential> GetStoredCredentials(InputArguments input)
+        private async Task<ICredential> GetStoredCredentials(Uri remoteUri, string userName, AuthenticationModes authModes)
         {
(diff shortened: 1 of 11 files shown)


src/shared/Atlassian.Bitbucket/BitbucketHostProvider.cs
```diff
@@ -372,14 +371,14 @@ private async Task<string> ResolveBasicAuthUserNameAsync(string username, string
...
+            _context.Trace.WriteLineSecrets($"Check if 2FA is required for credentials ({credentials.Account}/{{0}})...", new object[] { credentials.Password });
 
-            if (!SupportsOAuth(authModes))
-            {
-                return false;
-            }
+            // if (!SupportsOAuth(authModes))
+            // {
+            //     return false;
+            // }
```
ldennington, line 381: Was this left in on purpose?

src/shared/Atlassian.Bitbucket/BitbucketHostProvider.cs
```diff
@@ -402,31 +401,30 @@ private async Task<bool> RequiresTwoFactorAuthenticationAsync(ICredential creden
...
+                // TODO: Validate DC/Server credentials before returning them to Git
                 // Currently credentials for DC/Server are not checked by GCM.
                 // Instead the validation relies on Git to try and fail with the credentials and then request GCM to erase them
                 _context.Trace.WriteLine("For DC/Server skip validating existing credentials");
                 return await Task.FromResult(true);
             }
 
             // Bitbucket supports both OAuth + Basic Auth unless there is explicit GCM configuration
-            // So the credentials could be for either scheme therefore need to potentiall test both posibilities.
+            // So the credentials could be for either scheme therefore need to potentially test both possibilities.
```
ldennington, line 427:
```suggestion
            // The credentials could be for either scheme therefore need to potentially test both possibilities.
```

ldennington (approved):
Looks good - thanks especially for the UTs, they helped with understanding this change 🌟.

derrickstolee (approved):
I found @ldennington's review to be at least as thorough as mine. No new comments on top of hers.

src/shared/Atlassian.Bitbucket/BitbucketHostProvider.cs
mjcheetham: oops!

PR #584, opened 2021-12-27
Update package Nerdbank.GitVersioning

hickford:
Fixes #583

PR #587, opened 2021-12-28
Skip helper when no choice necessary

hickford:
Usability improvement for GitHubAuthentication.

I also wrote some unit tests.

### DIFF
--- a/src/shared/GitHub/GitHubAuthentication.cs
+++ b/src/shared/GitHub/GitHubAuthentication.cs
@@ -80,15 +80,16 @@ namespace GitHub
             switch (modes)
             {
                 case AuthenticationModes.Browser:
-                    // no prompt or user interaction required
+                    // no user interaction required
                     return new AuthenticationPromptResult(AuthenticationModes.Browser);
                 case AuthenticationModes.Device:
-                    // no prompt or user interaction required
+                    // no user interaction required
                     return new AuthenticationPromptResult(AuthenticationModes.Device);
                 default:
                     break;
             }
 
+            ThrowIfUserInteractionDisabled();
             if (Context.SessionManager.IsDesktopSession && TryFindHelperExecutablePath(out string helperPath))
             {
                 var promptArgs = new StringBuilder("prompt");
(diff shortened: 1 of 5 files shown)


Deeprai32037i (approved):

src/shared/GitHub/GitHubHostProvider.cs
```diff
@@ -316,6 +316,11 @@ public static bool IsGitHubDotCom(string targetUrl)
 
         public static bool IsGitHubDotCom(Uri targetUri)
         {
+            if (targetUri is null)
+            {
+                throw new ArgumentNullException(nameof(targetUri));
+            }
```
mjcheetham, line 322: We have a set of arg validation helpers that would be better to use here:

```suggestion
            EnsureArgument.NotNull(targetUri, nameof(targetUri));
```

..or for a URI, might even be better to validate for an absolute URI:

```suggestion
            EnsureArgument.AbsoluteUri(targetUri, nameof(targetUri));
```

(this also checks for `null` too)

src/shared/GitHub/GitHubAuthentication.cs
```diff
@@ -79,6 +76,19 @@ public async Task<AuthenticationPromptResult> GetAuthenticationAsync(Uri targetU
...
+            {
+                case AuthenticationModes.Browser:
+                    // no user interaction required
+                    return new AuthenticationPromptResult(AuthenticationModes.Browser);
+                case AuthenticationModes.Device:
+                    // no user interaction required
+                    return new AuthenticationPromptResult(AuthenticationModes.Device);
+                default:
+                    break;
+            }
```
mjcheetham, line 89: Let's be consistent with the above `if (modes == .None)` style here:

```suggestion
            // If there is no mode choice to be made and no interaction required,
            // just return that result.
            if (modes == AuthenticationModes.Browser ||
                modes == AuthenticationModes.Device)
            {
                    return new AuthenticationPromptResult(modes);
            }
```

Directory.Build.props
```diff
@@ -23,7 +23,7 @@
   <ItemGroup>
     <!-- All projects should use Nerdbank.GitVersioning for consistent version numbers -->
     <PackageReference Include="Nerdbank.GitVersioning">
-      <Version>3.3.37</Version>
```
mjcheetham, line 26: Please rebase on the latest `main` branch; I've since merged in your PR (#584) to update this package.

mjcheetham (changes requested):
Looks good thanks! Just a few style changes and please rebase to pick-up the new Nerdbank.GitVersioning package commit now in `main`.

Directory.Build.props
hickford: Done.

mjcheetham (approved):
Great stuff! Thanks!

PR #595, opened 2022-01-01
use record types

hickford:
record types for easier debugging and unit testing

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/record

### DIFF
--- a/src/shared/Atlassian.Bitbucket/BitbucketRestApi.cs
+++ b/src/shared/Atlassian.Bitbucket/BitbucketRestApi.cs
@@ -12,7 +12,7 @@ namespace Atlassian.Bitbucket
         Task<RestApiResult<UserInfo>> GetUserInformationAsync(string userName, string password, bool isBearerToken);
     }
 
-    public class RestApiResult<T>
+    public record RestApiResult<T>
     {
         public RestApiResult(HttpStatusCode statusCode)
             : this(statusCode, default(T)) { }
--- a/src/shared/Core.Tests/Authentication/OAuth2ClientTests.cs
+++ b/src/shared/Core.Tests/Authentication/OAuth2ClientTests.cs
@@ -310,8 +310,7 @@ namespace GitCredentialManager.Tests.Authentication
             Uri tokenEndpoint = new Uri(baseUri, "/oauth/v2.0/access_token");
             Uri deviceEndpoint = new Uri(baseUri, "/oauth/v2.0/authorize_device");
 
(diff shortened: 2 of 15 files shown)


ldennington (commented):
A couple things:

1. Based on my (admittedly limited) understanding, records are best-suited for immutable data models. However, some of these classes contain mutable data. Is this the correct approach for such classes?
2. Can you add a few more details on how this makes debugging and testing easier?

hickford:
Records automatically implement ToString (useful for debugging) and value equality (useful for unit testing). Both these properties are helpful regardless of mutability.

Immutable data is easier to reason about, so made all the record types immutable.

mjcheetham:
Records are a C# 9 feature, which is not supported for .NET Framework (that we still target on Windows).

hickford:
I might be wrong (unable to test), but I think this *will* build targeting .NET Framework, assuming Visual Studio 2019 or later, as records are principally a language feature rather than a library feature.

What won't work: C# 9's `init` keyword https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/init as this requires .NET 5 library [IsExternalInit](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.isexternalinit) (I tried targeting .NET Standard 2.0 and got error "IsExternalInit not defined or imported").

mjcheetham:
I must say I'm a little hesitant to adopt _some_ of the "latest and greatest" C# features. Admittedly I'm not overtly familiar with `record [class]`s (and `record struct`s), and that drives some of it..

But, from my understanding so far, a `record [class]` provides value-type behaviour to a reference type, specifically equality. So in short you avoid passing copies but get free member-wise equality.

Of the types that you've converted from `class` to `record` I don't think we've doing any in-product tests for equality, so this is only of benefit to tests (as you call out).

A `record` also makes immutable data types easier to implement, especially when coupled with the `init` keyword on properties. Since we cannot use `init` for our Windows/.NET Framework target, I fear we'll fall in to the trap of half-adopting some language features, but not able to fully use them due to these "ah but on Windows this doesn't work" problems caught late on.

All this is to say I'm not wholly against `record`s, but I'm not sure if it's the right time to adopt them, at least until we can get to .NET 6 on Windows in a good state. Thoughts @ldennington ?

ldennington:
I agree that the benefit does not currently outweigh the cost given our ties to .NET Framework. Let's close this for now, with the understanding we may re-visit if things change in the future.

Issue #598, opened 2022-01-05
Fix GitHub Enterprise API URL for raw source code links


ejball:
When storing credentials for raw URLs of GitHub Enterprise instances (as Visual Studio does when [Fall back to Git Credential Manager authentication for all Source Link requests](https://docs.microsoft.com/en-us/visualstudio/debugger/general-debugging-options-dialog-box?view=vs-2022) is enabled), the [GitHub Enterprise API hostname](https://github.com/GitCredentialManager/git-credential-manager/blob/a461ba71b8d77ca7f5b0f1da5686fe7b3c26f9ec/src/shared/GitHub/GitHubRestApi.cs#L213-L215) does not necessarily match the raw URL hostname. For example, a raw URL of `https://raw.my.ghe.dev/...` should access the API at `https://my.ghe.dev/api/v3`. For our purposes, removing `raw.` from the hostname would fix the issue, but perhaps some sort of configuration is necessary in the general case.

ejball:
Source Link supports this via the [`ContentUrl` attribute of `SourceLinkGitHubHost`](https://github.com/dotnet/sourcelink/blob/1ca09f079b12727cd4b1e49278a5e7e893d0ea1b/docs/README.md#custom-content-urls) (GHE usage [mentioned here](https://github.com/dotnet/sourcelink/issues/43#issuecomment-406415061)).

mjcheetham:
This should have been fixed after https://github.com/GitCredentialManager/git-credential-manager/pull/599 was merged. This feature will go out in the next GCM release - watch this space!

PR #599, opened 2022-01-05
Fix GitHub Enterprise API URL for raw source code links.

ejball:
This is a simple fix of #598 for GitHub Enterprise instances that use a `raw.` hostname prefix for raw source code links. I've verified this fix locally by swapping out the `GitHub.dll` that is used by Visual Studio.


ldennington (approved):
This looks all right to me, although as you mentioned in the issue configuration may be necessary if similar issues crop up. My one request would be to add a comment specifying why this replacement is happening.

ejball:
Good idea; done.

ejball:
Can this be merged?

ldennington:
> Can this be merged?

Yep! Thanks for the ping.

PR #600, opened 2022-01-06
Add option to disable GUI prompts/prefer terminal prompts instead

### LINKED ISSUE #580: Preference to prefer command line interface over UI
Hi. I updated git-credential-manager and now when I `git push` I get a GitHub 'sign in' window (screenshot below). It works fine but it takes a couple of seconds to open on my computer. I preferred the old command line "Select an authentication method" interface (text copied below), because it's faster and I don't have to use the mouse (uncomfortable for me).

```
Select an authentication method for 'https://github.com/':
  1. Web browser (default)
  2. Personal access token
option (enter for default): 
```

Is it possible to set a preference to use the command line interface instead of the UI?

image

https://github.com/GitCredentialManager/git-credential-manager/releases/tag/v2.0.474

https://github.com/GitCredentialManager/git-credential-manager/pull/336

mjcheetham:
Add a new setting and environment variable that allows a user to disable the use of GUI prompts (in favour of terminal/text-based ones where available). The `GCM_GUI_PROMPT` envar mirrors the existing Git `GIT_TERMINAL_PROMPT` setting in its naming (being singular).

Note that we don't do anything with the MSAL prompts here, since there is already an existing setting for controlling the auth flow used (`GCM_MSAUTH_FLOW`). Also prompting is not always deterministic or avoidable with some MSAuth flows (i.e, Windows broker).

We also don't consider the user's web browser to be a "prompt" and continue to offer those authentication modes. This is only about input prompts GCM itself controls _where there is a terminal prompt available instead_.

Also update the FAQ document to include a note about the new settings, and also mention the "helper" variables that allow users to override the UI helpers for GitHub and Bitbucket.

Fixes #580

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -109,3 +109,17 @@ Follow the instructions in [our WSL guide](wsl.md) carefully. Especially note th
 ### Does GCM work with multiple users? If so, how?
 
 That's a fairly complicated question to answer, but in short, yes. See [our document on multiple users](multiple-users.md) for details.
+
+### How can I extend GUI prompts/integrate prompts with my application?
+
+You can replace the GUI prompts of the Bitbucket and GitHub host providers
+specifically by using the `credential.gitHubHelper`/`credential.bitbucketHelper`
+settings or `GCM_GITHUB_HELPER`/`GCM_BITBUCKET_HELPER` environment variables.
+
+Set these variables to the path of an external helper executable that responds
+to the requests as the bundled UI helpers do. See the current `--help` documents
+for the bundled UI helpers (`GitHub.UI`/`Atlassian.Bitbucket.UI`) for more
+information.
+
+You may also set these variables to the empty string `""` to force terminal/
+text-based prompts instead.
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -121,6 +121,10 @@ customize how GCM will prompt you (or not) for input. Please see the following:
 
 ### How can I extend GUI prompts/integrate prompts with my application?
 
+Application developers who use Git - think Visual Studio, GitKraken, etc. - may
+want to replace the GCM default UI with prompts styled to look like their
+application. This isn't complicated (though it is a bit of work).
+
 You can replace the GUI prompts of the Bitbucket and GitHub host providers
 specifically by using the `credential.gitHubHelper`/`credential.bitbucketHelper`
 settings or `GCM_GITHUB_HELPER`/`GCM_BITBUCKET_HELPER` environment variables.
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -110,6 +110,15 @@ Follow the instructions in [our WSL guide](wsl.md) carefully. Especially note th
 
 That's a fairly complicated question to answer, but in short, yes. See [our document on multiple users](multiple-users.md) for details.
 
+### How can I disable GUI dialogs and prompts?
+
+There are various environment variables and configuration options available to
+customize how GCM will prompt you (or not) for input. Please see the following:
+
+- [`GCM_INTERACTIVE`](environment.md#GCM_INTERACTIVE) / [`credential.interactive`](configuration.md#credentialinteractive)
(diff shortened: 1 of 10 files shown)


hickford:
Thanks for this! 

Consider defaulting to terminal for usability. Alice types `git push` in a terminal. If Alice hasn't anticipated having to authenticate, a prompt at the terminal is less surprising than an unexpected GUI window (to me, this feels like Clippy popping up out of nowhere to say "it looks like you're trying to authenticate"). The sequence "terminal -> browser -> terminal" is simpler and faster than "terminal -> GUI -> browser -> terminal".

Is this not typical? I suppose Bob might run git push via a GUI such as Git Extensions. In that case, I would prefer a GUI prompt to a terminal window, again to minimise context switching (Git Extensions GUI -> Git Credential Manager GUI -> browser -> Git Extensions GUI).

Can you tell whether git is running at an interactive terminal or launched by another app?

docs/faq.md
```diff
@@ -109,3 +109,26 @@ Follow the instructions in [our WSL guide](wsl.md) carefully. Especially note th
...
+There are various environment variables and configuration options available to
+customize how GCM will prompt you (or not) for input. Please see the following:
+
+- [`GCM_INTERACTIVE`](environment.md#GCM_INTERACTIVE) / [`credential.interactive`](configuration.md#credentialinteractive)
+- [`GCM_GUI_PROMPT`](environment.md#GCM_GUI_PROMPT) / [`credential.guiPrompt`](configuration.md#credentialguiprompt)
+- [`GIT_TERMINAL_PROMPT`](https://git-scm.com/docs/git#Documentation/git.txt-codeGITTERMINALPROMPTcode) (note this is a _Git setting_ that will affect Git as well as GCM)
+
+### How can I extend GUI prompts/integrate prompts with my application?
+
+You can replace the GUI prompts of the Bitbucket and GitHub host providers
```
vtbassmatt, line 124: My only quibble here is that not everyone reading this will be the target audience. I might start with a line like "Application developers who use Git --- think Visual Studio, GitKraken, etc. --- may want to replace the GCM default UI with prompts styled to look like their application. This isn't complicated (though it is a bit of work)."

vtbassmatt (approved):
This LGTM.

mjcheetham:
Hello!

> Consider defaulting to terminal for usability.

GCM and Git is used by a very wide range of people with various levels of comfortability with using a terminal. We must consider that those who are less comfortable using a terminal should want to use a GUI in the first instance. Those who are more "advanced" users are able to customise their setup using this new variable to disable GUI prompts.

> Is this not typical? I suppose Bob might run git push via a GUI such as Git Extensions. In that case, I would prefer a GUI prompt to a terminal window.

GCM is invoked by Git, and many.. many applications like editors/IDE (e.g. VSCode, Sourcetree). These often shell out to Git when you interact with their UIs. This is a very common scenario, and in fact the main driver for having GUI prompts - the editor/IDE does not expose the connected terminal, so there's no way for the user to respond.

> My preference is to minimise context switching [..] Can you tell whether git is running at an interactive terminal or launched by another app?

This would be my preference too, but sadly it's not really possible in practice to determine this. This is especially true on Windows where Cygwin/MSYS2 do 'interesting' things with pseudo terminals. `isatty(int)` is not enough to be able to determine the actual calling environment of GCM's parent Git process.

docs/faq.md
mjcheetham: That's a fair point. I'll add your suggestion.

hickford:
Thanks for the explanation!

src/shared/Core/Authentication/AuthenticationBase.cs
```diff
@@ -92,6 +90,16 @@ protected void ThrowIfUserInteractionDisabled()
             }
         }
 
+        protected void ThrowIfGuiPromptsDisabled()
```
ldennington, line 93: Are we using this anywhere? Or just adding it for potential future use?

ldennington (approved):
Great job! I especially love how you structured your commits - so clean and concise! ⭐

PR #612, opened 2022-01-18
Update `IsWslPath` to detect `\\wsl.localhost`-style paths

### LINKED ISSUE #611: GIT_EXEC_PATH being wsl.localhost rather than wsl$
Hi,

I'm trying to use the credential manager installed on Windows 11, within WSL2, without Git for Windows installed, following along as per the documentation (Configuring WSL without Git for Windows).

However, whenever I try using a git command from within WSL2, it looks like it's still looking for the Windows executable (git.exe), rather than the Linux executable:

```
Unhandled Exception: System.Exception: Failed to locate 'git.exe' executable on the path.
   at GitCredentialManager.EnvironmentExtensions.LocateExecutable(IEnvironment environment, String program)
   at GitCredentialManager.CommandContext.GetGitPath(IEnvironment environment, IFileSystem fileSystem, ITrace trace)
   at GitCredentialManager.CommandContext..ctor(String appPath)
   at GitCredentialManager.Program.Main(String[] args)
```

This is running version: **2.0.632.34631**.

I had a quick look at how the credential manager works out which executable to use, and noticed that it checks to see if the git exec path starts with `\\wsl$\`.

I'm unsure if I've misconfigured something here, but on my fresh install of Windows 11, the translated path on my machine ends up being `\\wsl.localhost\Ubuntu\usr\lib\git-core`, which I checked with the following command in cmd.exe launched via WSL2:

```
C:\Windows>echo %WSLENV%
WT_SESSION:WT_SESSION::WT_PROFILE_ID:GIT_EXEC_PATH/wp:WT_PROFILE_ID

C:\Windows>echo %GIT_EXEC_PATH%
\\wsl.localhost\Ubuntu\usr\lib\git-core
```

Which I assume doesn't meet the `IsWslPath` check, so it then continues to look for git.exe.

I'm happy to work on a PR to add the additional prefix to the `IsWslPath` check if that ends up being the solution, but otherwise if there's something I've missed in setting this up, would be happy with that solution as well.

mjcheetham:
Starting in [Windows build 21354](https://docs.microsoft.com/en-us/windows/wsl/release-notes#build-21354), the preferred WSL UNC file path is `\\wsl.localhost\<distro>\` and not the older administrative-share style `\\wsl$\<distro>\`.

The older "wsl$" style will continue to be supported, but won't be the default type of path we'll see from the system going forwards.

We now support detecting both styles as WSL paths.

Fixes #611

### DIFF
--- a/src/shared/Core.Tests/WslUtilsTests.cs
+++ b/src/shared/Core.Tests/WslUtilsTests.cs
@@ -6,7 +6,7 @@ namespace GitCredentialManager.Tests
 {
     public class WslUtilsTests
     {
-        [PlatformTheory(Platforms.Windows)]
+        [Theory]
         [InlineData(null, false)]
         [InlineData(@"", false)]
         [InlineData(@" ", false)]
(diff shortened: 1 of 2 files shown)


vdye (approved):
Looks good!

dscho (approved):

PR #614, opened 2022-01-20
Sign and ship GitLab binaries (and some minor fixes)

mjcheetham:
- Ensure the GitLab.dll host provider assembly is being code-signed on Windows and Mac, and that it it included in the Windows installers.
- Remove a few unused `using` statements
- Ensure the correct project configurations are being selected
- For custom GitLab instances allow all auth modes (since we cannot detect)
- Use constants where appropriate


src/shared/GitLab/GitLabHostProvider.cs
```diff
@@ -134,12 +134,12 @@ internal AuthenticationModes GetSupportedAuthenticationModes(Uri targetUri)
...
             if (GitLabConstants.IsGitLabDotCom(targetUri))
             {
-                return AuthenticationModes.All;
+                return GitLabConstants.DotComAuthenticationModes;
             }
 
             Context.Streams.Error.WriteLine($"Missing OAuth configuration for {targetUri.Host}, see https://github.com/GitCredentialManager/git-credential-manager/blob/main/docs/gitlab.md.");
             // Would like to query password_authentication_enabled_for_git, but can't unless logged in https://gitlab.com/gitlab-org/gitlab/-/issues/349463
-            return AuthenticationModes.Basic | AuthenticationModes.Pat;
+            return AuthenticationModes.All;
```
hickford, line 142: Better would be to include Browser if and only if GitLabDevClientId and GitLabDevClientSecret are populated,  otherwise OAuth will fail. I couldn't figure out how to check this and print the warning. 

Example config if it helps to test:

```
[credential "https://gitlab.freedesktop.org"]
        GitLabDevClientId = 6503d8c5a27187628440d44e0352833a2b49bce540c546c22a3378c8f5b74d45
        GitLabDevClientSecret = 2ae9343a034ff1baadaef1e7ce3197776b00746a02ddf0323bb34aca8bff6dc1
```

hickford (approved):

dscho (approved):
LGTM

src/shared/GitLab/GitLabHostProvider.cs
mjcheetham: That's a good point and idea. I've updated the PR to check for the presence of a configured client ID (that isn't gitlab.com's) and print a warning/link to the docs (and disallow browser auth).

derrickstolee (approved):

derrickstolee:
I dismissed the codescan alerts using the "Used in tests" reason. Hopefully those choices stick for next time.

PR #627, opened 2022-02-27
Pre-emptively store GitHub OAuth credential

hickford:
Pre-emptively store GitHub OAuth credential so don't have to authenticate again if push fails 

In general, if git push fails, the credential is not stored. This makes sense -- the credential might be bad. However in case of OAuth, we know the credential we generate to be correct, so can pre-emptively store it. This avoids having to authenticate again after a push fail, as in the example below, where push fails for a non-auth reason.

```
> git push upstream
Select an authentication method for 'https://github.com/':
  1. Web browser (default)
  2. Device code
  3. Personal access token
option (enter for default): 1
remote: Permission to GitCredentialManager/git-credential-manager.git denied to hickford.
fatal: unable to access 'https://github.com/GitCredentialManager/git-credential-manager/': The requested URL returned error: 403
> git push upstream
Select an authentication method for 'https://github.com/':
  1. Web browser (default)
  2. Device code
  3. Personal access token
option (enter for default):  
```

### DIFF
--- a/src/shared/GitHub/GitHubHostProvider.cs
+++ b/src/shared/GitHub/GitHubHostProvider.cs
@@ -150,10 +150,10 @@ namespace GitHub
                     return patCredential;
 
                 case AuthenticationModes.Browser:
-                    return await GenerateOAuthCredentialAsync(remoteUri, useBrowser: true);
+                    return await GenerateOAuthCredentialAsync(input, useBrowser: true);
 
                 case AuthenticationModes.Device:
-                    return await GenerateOAuthCredentialAsync(remoteUri, useBrowser: false);
+                    return await GenerateOAuthCredentialAsync(input, useBrowser: false);
 
                 case AuthenticationModes.Pat:
                     // The token returned by the user should be good to use directly as the password for Git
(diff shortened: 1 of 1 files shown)


mjcheetham:
The problem with the "store-on-`get`" approach is that we have to be very careful about our logic to determine if the token is good or bad for the given Git operation.

For the scenario you describe (get a valid token -> `push` fails for another reason -> try again and use same token), Git will tell us to `erase` the credential after the failure (but not _why_ the failure occurred). Do we erase this credential or not when asked to?

If we ignore erase requests when "we know better", the user may end up stuck with a bad credential that Git/the server keeps telling us is bad, but GCM thinks is OK. If Git could tell us why auth failed<sup>1</sup>, we might have more confidence in making a determination on if the credential is valid or not.

One example where I'm not sure how we could possibly know is with authZ for pushes that invoke a GitHub workflow. You need the `workflow` scope if the commits and tree's you're trying to push would [introduce or update a workflow file that is _not_ present in another branch in the same repository](https://docs.github.com/en/developers/apps/building-oauth-apps/scopes-for-oauth-apps#available-scopes). GCM doesn't get told about that 403, and neither do we know ahead of time in the `get` request if we're going to fail for the given operation. Invocations of credential helpers is also stateless.. `get` and `store` or `erase` are not connected in anyway to make the problem worse(!)

---
[1] note that Git doesn't invoke helpers for `403`s (only `401`s)

wendylove14700:
> git push upstream
Select an authentication method for 'https://github.com/':
  1. Web browser (default)
  2. Device code
  3. Personal access token
option (enter for default): 1
remote: Permission to GitCredentialManager/git-credential-manager.git denied to hickford.
fatal: unable to access 'https://github.com/GitCredentialManager/git-credential-manager/': The requested URL returned error: 403
> git push upstream
Select an authentication method for 'https://github.com/':
  1. Web browser (default)
  2. Device code
  3. Personal access token
option (enter for default):

Vladisvob (commented):
Duplicate of #

PR #629, opened 2022-03-01
Exclude Windows Server 2016 from WAM support

### LINKED ISSUE #487: Crash when using WAM on Windows Server 2016
**Which version of GCM Core are you using?**

2.0.498

* [x] Azure DevOps

* [x] Yes

---

* [x] https://dev.azure.com/`{org}`/...

* [x] Not applicable

---

**Expected behavior**

The Windows broker account picker appears and I can sign in with a new AAD account.

**Actual behavior**

An error is thrown:

```text
fatal: Could not get the account provider - account picker. See inner exception for details
fatal: Method not found: 'Void Windows.Security.Authentication.Web.Core.WebTokenRequest.put_CorrelationId(System.String)'.
```

**Logs**

Set the environment variables `GCM_TRACE=1` and `GIT_TRACE=1` and re-run your Git command. Review and redact any private information and attach the log.

[gcm.log](https://github.com/microsoft/Git-Credential-Manager-Core/files/7328438/gcm.log)

mjcheetham:
Windows Server 2016 does not support WAM, so we should not try to enable the broker unless we're on Server 2019 and later.

Also update the Windows (client; non-server) version checks to be tighter. Previously any major version of 10 or greater was considered "supported", but this is wrong. Windows 10 build 15063 was the first to support WAM.

Fixes #487

### DIFF
--- a/src/shared/Core/Authentication/MicrosoftAuthentication.cs
+++ b/src/shared/Core/Authentication/MicrosoftAuthentication.cs
@@ -55,7 +55,7 @@ namespace GitCredentialManager.Authentication
             IsBrokerInitialized = true;
 
             // Broker is only supported on Windows 10 and later
-            if (!PlatformUtils.IsWindows10OrGreater())
+            if (!PlatformUtils.IsWindowsBrokerSupported())
             {
                 return;
             }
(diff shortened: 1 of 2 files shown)


jeffhostetler (approved):
Not sure I understand enough of the issues or APIs to have an an 
informed opinion, but LGTM.

derrickstolee (approved):

mjcheetham:
cc @bgavrilms

bgavrilMS:
Not sure you need to go to all this trouble, we fixed this in MSAL itself.

mjcheetham:
> Not sure you need to go to all this trouble, we fixed this in MSAL itself.

The problem is that `IsBrokerAvailable()` is an instance method on `PublicClientApplicationBuilder`, which means we need to load MSAL and create a `PCAB` (with client ID etc) in `Main()` for every GCM invocation. We don't always know the client ID at this stage, and the extra cost required isn't worth it IMO.

PR #630, opened 2022-03-02
Add `install-from-source.sh` and surrounding infrastructure

ldennington:
Add a helper script for Linux users to install GCM from source. This includes:

1. Updates to `build.sh` to support a new `--install-from-source` parameter.
2. Addition of the script itself.
3. Addition of a new GitHub actions workflow to validate the script on supported distributions.
4. Addition of instructions to obtain/run the script in `README.md`. __Note:__ These instructions _will not_ work currently, as they point to the future location of the script in `main`.

### DIFF
--- a/.github/workflows/validate-install-from-source.yml
+++ b/.github/workflows/validate-install-from-source.yml
@@ -1,0 +1,33 @@
+name: validate-install-from-source
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  docker:
+    name: ${{matrix.vector.image}}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        vector:
+        - image: ubuntu
+        - image: debian
+        - image: linuxmintd/mint20-amd64
+        - image: fedora
+        - image: centos
+        - image: redhat/ubi8
+        - image: alpine
+    container: ${{matrix.vector.image}}
+    steps:
+      - uses: actions/checkout@v1
+      - run: |
+          if [ ${{matrix.vector.image}} == "centos" ]; then
+            sed -i 's/mirrorlist/#mirrorlist/g' /etc/yum.repos.d/CentOS-Linux-*
+            sed -i 's|#baseurl=http://mirror.centos.org|baseurl=http://vault.centos.org|g' /etc/yum.repos.d/CentOS-Linux-*
+          fi
+
+          sh "${GITHUB_WORKSPACE}/src/linux/Packaging.Linux/install-from-source.sh" -y
+          git-credential-manager-core --help || exit 1
--- a/README.md
+++ b/README.md
@@ -89,6 +89,32 @@ sudo /usr/local/share/gcm-core/uninstall.sh
 <a name="linux-install-instructions"></a>
 ### Linux
 
+#### Experimental: install from source helper script
+
+If you would like to help dogfood our new install from source helper script,
+run the following:
+
+1. To ensure `curl` is installed:
+
+```shell
+curl --version
+```
+
+If `curl` is not installed, please use your distribution's package manager
+to install it.
+
+0. To download and run the script:
+
+```shell
+curl -LO https://raw.githubusercontent.com/GitCredentialManager/git-credential-manager/main/src/linux/Packaging.Linux/install-from-source.sh &&
+sh ./install-from-source.sh &&
+git-credential-manager-core configure
+```
+
+__Note:__ You will be prompted to enter your credentials so that the script
+can download GCM's dependencies using your distribution's package
+manager.
+
 #### Ubuntu/Debian distributions
 
 Download the latest [.deb package](https://github.com/GitCredentialManager/git-credential-manager/releases/latest), and run the following:
--- a/src/linux/Packaging.Linux/Packaging.Linux.csproj
+++ b/src/linux/Packaging.Linux/Packaging.Linux.csproj
@@ -7,6 +7,10 @@
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <InstallFromSource>false</InstallFromSource>
+  </PropertyGroup>
+
   <ItemGroup>
     <None Include="**/*" />
   </ItemGroup>
@@ -19,8 +23,8 @@
   <Import Project="Sdk.targets" Sdk="Microsoft.NET.Sdk" />
 
   <Target Name="CoreCompile" DependsOnTargets="GetBuildVersion" Condition="'$(OSPlatform)'=='linux'">
-    <Message Text="$(MSBuildProjectDirectory)\build.sh --configuration='$(Configuration)' --version='$(BuildVersion)'" Importance="High" />
-    <Exec Command="$(MSBuildProjectDirectory)\build.sh --configuration='$(Configuration)' --version='$(BuildVersion)'" />
+    <Message Text="$(MSBuildProjectDirectory)\build.sh --install-from-source=$(InstallFromSource) --configuration='$(Configuration)' --version='$(BuildVersion)'" Importance="High" />
+    <Exec Command="$(MSBuildProjectDirectory)\build.sh --install-from-source=$(InstallFromSource) --configuration='$(Configuration)' --version='$(BuildVersion)'" />
   </Target>
 
   <Target Name="CoreClean">
--- a/src/linux/Packaging.Linux/build.sh
+++ b/src/linux/Packaging.Linux/build.sh
@@ -19,7 +19,6 @@ make_absolute () {
 # Building
 #####################################################################
 echo "Building Packaging.Linux..."
-
 # Parse script arguments
 for i in "$@"
 do
@@ -30,6 +29,10 @@ case "$i" in
     ;;
     --version=*)
     VERSION="${i#*=}"
+    shift # past argument=value
+    ;;
+    --install-from-source=*)
+    INSTALL_FROM_SOURCE=${i#*=}
     shift # past argument=value
     ;;
     *)
@@ -59,22 +62,28 @@ if [ -z "$VERSION" ]; then
     die "--version was not set"
 fi
 
-ARCH="`dpkg-architecture -q DEB_HOST_ARCH`"
-if test -z "$ARCH"; then
-  die "Could not determine host architecture!"
+if [ $INSTALL_FROM_SOURCE = false ]; then
+    ARCH="`dpkg-architecture -q DEB_HOST_ARCH`"
+    if test -z "$ARCH"; then
+    die "Could not determine host architecture!"
+    fi
 fi
 
 # Outputs
 PAYLOAD="$PROJ_OUT/payload/$CONFIGURATION"
 SYMBOLOUT="$PROJ_OUT/payload.sym/$CONFIGURATION"
 
-TAROUT="$PROJ_OUT/tar/$CONFIGURATION"
-TARBALL="$TAROUT/gcmcore-linux_$ARCH.$VERSION.tar.gz"
-SYMTARBALL="$TAROUT/symbols-linux_$ARCH.$VERSION.tar.gz"
+if [ $INSTALL_FROM_SOURCE = false ]; then
+    TAROUT="$PROJ_OUT/tar/$CONFIGURATION"
+    TARBALL="$TAROUT/gcmcore-linux_$ARCH.$VERSION.tar.gz"
+    SYMTARBALL="$TAROUT/symbols-linux_$ARCH.$VERSION.tar.gz"
 
-DEBOUT="$PROJ_OUT/deb/$CONFIGURATION"
-DEBROOT="$DEBOUT/root"
-DEBPKG="$DEBOUT/gcmcore-linux_$ARCH.$VERSION.deb"
+    DEBOUT="$PROJ_OUT/deb/$CONFIGURATION"
+    DEBROOT="$DEBOUT/root"
+    DEBPKG="$DEBOUT/gcmcore-linux_$ARCH.$VERSION.deb"
+else
+    INSTALL_LOCATION="/usr/local"
+fi
 
 # Cleanup payload directory
 if [ -d "$PAYLOAD" ]; then
@@ -89,7 +98,13 @@ if [ -d "$SYMBOLOUT" ]; then
 fi
 
 # Ensure directories exists
-mkdir -p "$PAYLOAD" "$SYMBOLOUT" "$DEBROOT"
+mkdir -p "$PAYLOAD" "$SYMBOLOUT"
+
+if [ $INSTALL_FROM_SOURCE = false ]; then
+    mkdir -p "$DEBROOT"
+else
+    mkdir -p "$INSTALL_LOCATION"
+fi
 
 # Publish core application executables
 echo "Publishing core application..."
@@ -135,40 +150,44 @@ mv "$PAYLOAD"/*.pdb "$SYMBOLOUT" || exit 1
 echo "Build complete."
 
 #####################################################################
-# PACKING
+# PACKING AND INSTALLING
 #####################################################################
-echo "Packing Packaging.Linux..."
-# Cleanup any old archive files
-if [ -e "$TAROUT" ]; then
-    echo "Deleteing old archive '$TAROUT'..."
-    rm "$TAROUT"
-fi
-
-# Ensure the parent directory for the archive exists
-mkdir -p "$TAROUT" || exit 1
-
 # Set full read, write, execute permissions for owner and just read and execute permissions for group and other
 echo "Setting file permissions..."
 /bin/chmod -R 755 "$PAYLOAD" || exit 1
 
-# Build binaries tarball
-echo "Building binaries tarball..."
-pushd "$PAYLOAD"
-tar -czvf "$TARBALL" * || exit 1
-popd
+if [ $INSTALL_FROM_SOURCE = false ]; then
+    echo "Packing Packaging.Linux..."
+    # Cleanup any old archive files
+    if [ -e "$TAROUT" ]; then
+        echo "Deleteing old archive '$TAROUT'..."
+        rm "$TAROUT"
+    fi
 
-# Build symbols tarball
-echo "Building symbols tarball..."
-pushd "$SYMBOLOUT"
-tar -czvf "$SYMTARBALL" * || exit 1
-popd
+    # Ensure the parent directory for the archive exists
+    mkdir -p "$TAROUT" || exit 1
 
-# Build .deb
-INSTALL_TO="$DEBROOT/usr/local/share/gcm-core/"
-LINK_TO="$DEBROOT/usr/local/bin/"
-mkdir -p "$DEBROOT/DEBIAN" "$INSTALL_TO" "$LINK_TO" || exit 1
+    # Build binaries tarball
+    echo "Building binaries tarball..."
+    pushd "$PAYLOAD"
+    tar -czvf "$TARBALL" * || exit 1
+    popd
+
+    # Build symbols tarball
+    echo "Building symbols tarball..."
+    pushd "$SYMBOLOUT"
+    tar -czvf "$SYMTARBALL" * || exit 1
+    popd
+
+    # Build .deb
+    INSTALL_TO="$DEBROOT/usr/local/share/gcm-core/"
+    LINK_TO="$DEBROOT/usr/bin/"
+    mkdir -p "$DEBROOT/DEBIAN" "$INSTALL_TO" "$LINK_TO" || exit 1
 
 # make the debian control file
+# this is purposefully not indented, see
+# https://stackoverflow.com/questions/9349616/bash-eof-in-if-statement
+# for details
 cat >"$DEBROOT/DEBIAN/control" <<EOF
 Package: gcmcore
 Version: $VERSION
@@ -183,13 +202,25 @@ Description: Cross Platform Git Credential Manager command line utility.
  For more information see https://aka.ms/gcm
 EOF
 
+    dpkg-deb --build "$DEBROOT" "$DEBPKG" || exit 1
+else
+    echo "Installing..."
+
+    # Install directories
+    INSTALL_TO="$INSTALL_LOCATION/share/gcm-core/"
+    LINK_TO="$INSTALL_LOCATION/bin/"
+    MESSAGE="Install complete."
+fi
+
+mkdir -p "$INSTALL_TO" "$LINK_TO"
+
 # Copy all binaries and shared libraries to target installation location
 cp -R "$PAYLOAD"/* "$INSTALL_TO" || exit 1
 
 # Create symlink
-ln -s -r "$INSTALL_TO/git-credential-manager-core" \
-    "$LINK_TO/git-credential-manager-core" || exit 1
+if [ ! -f "$LINK_TO/git-credential-manager-core" ]; then
+    ln -s -r "$INSTALL_TO/git-credential-manager-core" \
+        "$LINK_TO/git-credential-manager-core" || exit 1
+fi
 
-dpkg-deb --build "$DEBROOT" "$DEBPKG" || exit 1
-
-echo "Pack complete."
+echo $MESSAGE
--- a/src/linux/Packaging.Linux/install-from-source.sh
+++ b/src/linux/Packaging.Linux/install-from-source.sh
@@ -1,0 +1,183 @@
+#!/bin/sh
+
+# halt execution immediately on failure
+# note there are some scenarios in which this will not exit;
+# see https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html
+# for additional details
+set -e
+
+is_ci=
+for i in "$@"; do
+    case "$i" in
+        -y)
+        is_ci=true
+        shift # past argument=value
+        ;;
+    esac
+done
+
+# in non-ci scenarios, advertise what we will be doing and
+# give user the option to exit
+if [ -z $is_ci ]; then
+    echo "This script will download, compile, and install Git Credential Manager to:
+
+    /usr/local/bin
+
+Git Credential Manager is licensed under the MIT License: https://aka.ms/gcm/license"
+
+    while true; do
+        read -p "Do you want to continue? [Y/n] " yn
+        case $yn in
+            [Yy]*|"")
+                break
+            ;;
+            [Nn]*)
+                exit
+            ;;
+            *) 
+                echo "Please answer yes or no."
+            ;;
+        esac
+    done
+fi
+
+install_shared_packages() {
+    pkg_manager=$1
+    install_verb=$2
+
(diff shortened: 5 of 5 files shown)


src/linux/Packaging.Linux/Packaging.Linux.csproj
```diff
@@ -18,9 +22,14 @@
   <!-- Implicit SDK targets import (so we can override the default targets below) -->
   <Import Project="Sdk.targets" Sdk="Microsoft.NET.Sdk" />
 
-  <Target Name="CoreCompile" DependsOnTargets="GetBuildVersion" Condition="'$(OSPlatform)'=='linux'">
-    <Message Text="$(MSBuildProjectDirectory)\build.sh --configuration='$(Configuration)' --version='$(BuildVersion)'" Importance="High" />
-    <Exec Command="$(MSBuildProjectDirectory)\build.sh --configuration='$(Configuration)' --version='$(BuildVersion)'" />
+  <Target Name="CoreCompile" DependsOnTargets="GetBuildVersion" Condition="'$(OSPlatform)'=='linux' And '$(InstallFromSource)'=='false'">
+    <Message Text="$(MSBuildProjectDirectory)\build.sh --install-from-source=false --configuration='$(Configuration)' --version='$(BuildVersion)'" Importance="High" />
```
dscho, line 26: Should we maybe do `--install-from-source=$(InstallFromSource)` instead of having two separate `<Target>`s?

src/linux/Packaging.Linux/build.sh
```diff
@@ -135,61 +151,74 @@ mv "$PAYLOAD"/*.pdb "$SYMBOLOUT" || exit 1
...
-    "$LINK_TO/git-credential-manager-core" || exit 1
-
-dpkg-deb --build "$DEBROOT" "$DEBPKG" || exit 1
+if [ ! -f "$LINK_TO/git-credential-manager-core" ]; then
+    ln -s -r "$INSTALL_TO/git-credential-manager-core" \
+        "$LINK_TO/git-credential-manager-core" || exit 1
+fi
 
-echo "Pack complete."
+echo "Install complete."
```
dscho, line 224: I guess this should still say `Pack complete` if `"$INSTALL_FROM_SOURCE" != true`? The way I would do it: set a variable `message` in the `if`/`else` above, to `"Install complete."` and `"Pack complete."`, respectively.

src/linux/Packaging.Linux/install-from-source.sh
```diff
@@ -0,0 +1,101 @@
...
+        if [ $pkg_manager = apk ]; then
+            $SUDO_CMD $pkg_manager $install_verb $package
+        else
+            $SUDO_CMD $pkg_manager $install_verb $package -y
+        fi
+    done
+}
+
+download_dotnet_script() {
+    curl -o dotnet-install.sh -L https://dot.net/v1/dotnet-install.sh
```
dscho, line 32: This `curl` call could fail, but the script would not stop in that case. How about calling `set -e` at the beginning of this script, so that the script execution stops as soon as a command failed?

src/linux/Packaging.Linux/install-from-source.sh
```diff
@@ -0,0 +1,101 @@
...
+    chmod +x ./dotnet-install.sh
+}
+
+distribution=$(sed -n '/^ID=/p' /etc/os-release | cut -c 4- | tr -d '"')
```
dscho, line 36: A slightly more robust alternative would be:

```sh
eval "$(sed -n 's/^ID=/distribution=/p' /etc/os-release)"
```

This would not assume that the value is enclosed in double-quotes (the `cut -c 4-` assumes that the character after `ID=` is uninteresting 😄).

src/linux/Packaging.Linux/install-from-source.sh
```diff
@@ -0,0 +1,101 @@
...
+case $distribution in
+    debian | ubuntu | linuxmint)
+        # add dotnet package repository/signing key
+        $SUDO_CMD apt update && $SUDO_CMD apt install wget -y
+        eval wget https://packages.microsoft.com/config/ubuntu/20.04/packages-microsoft-prod.deb -O packages-microsoft-prod.deb
```
dscho, line 41: No need to `eval` this command.

Also, you used `curl` above, maybe we require only one of `curl` and `wget`? In other words, I would suggest to replace this call by:

```sh
curl -LO https://packages.microsoft.com/config/ubuntu/20.04/packages-microsoft-prod.deb
```

Also, do we have to use the `VERSION_ID` from `/etc/os-release` to avoid hard-coding `20.04` in this URL?

src/linux/Packaging.Linux/install-from-source.sh
```diff
@@ -0,0 +1,101 @@
...
+
+        install_shared_packages apt install
+
+        # install dotnet packages
+        $SUDO_CMD apt install dotnet-sdk-5.0 dpkg-dev -y
+    ;;
+    fedora | centos | rhel)
+        if [ $distribution = centos ]; then
+            sed -i 's/mirrorlist/#mirrorlist/g' /etc/yum.repos.d/CentOS-Linux-*
+            sed -i 's|#baseurl=http://mirror.centos.org|baseurl=http://vault.centos.org|g' /etc/yum.repos.d/CentOS-Linux-*
```
dscho, line 58: I am not sure that CentOS users would like their `/etc/yum.repos.d/` being edited as part of installing GCM. Can we avoid that?

dscho (commented):
Looks good! I have a couple of suggestions which I hope you will find helpful.

src/linux/Packaging.Linux/install-from-source.sh
```diff
@@ -0,0 +1,101 @@
...
+    do
+        if [ $pkg_manager = apk ]; then
+            $SUDO_CMD $pkg_manager $install_verb $package
+        else
+            $SUDO_CMD $pkg_manager $install_verb $package -y
+        fi
+    done
+}
+
+download_dotnet_script() {
```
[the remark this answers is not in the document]
mjcheetham: Would also need to verify the minimum version of the .NET SDK (5.0 right now) was installed too.

```shell
% dotnet --list--sdks
5.0.403 [/usr/local/share/dotnet/sdk]
6.0.100 [/usr/local/share/dotnet/sdk]
```

src/linux/Packaging.Linux/install-from-source.sh
mjcheetham: ..or a `|| exit 1` or `die(..)` even.

src/linux/Packaging.Linux/install-from-source.sh
mjcheetham: > Also, you used curl above, maybe we require only one of curl and wget?

+1 using `curl` over `wget` (or visa versa).

> Also, do we have to use the VERSION_ID from /etc/os-release to avoid hard-coding 20.04 in this URL?

Does the .NET team publish runtimes for all the different distros we want/test with?

.github/workflows/validate-install-from-source.yml
```diff
@@ -0,0 +1,28 @@
...
+        - image: linuxmintd/mint20-amd64
+        - image: fedora
+        - image: centos
+        - image: redhat/ubi8
+        - image: alpine
+    container: ${{matrix.vector.image}}
+    steps:
+      - uses: actions/checkout@v1
+      # Run install from source script, passing an empty string to
+      # `--sudo-cmd` since containers run as root.
```
mjcheetham, line 27: I'm wondering if we can avoid this `--sudo-cmd=""` here by updating the script to do something like:

```shell
if [ -z "$SUDO_CMD" &&  ]; then
    if [ `id -u` != 0 ]; then
        SUDO_CMD=sudo
    fi
fi
```

If we're already effectively `root`, there's no need for `sudo`.

src/linux/Packaging.Linux/install-from-source.sh
```diff
@@ -0,0 +1,101 @@
...
+    ;;
+    *)
+        echo "ERROR: Unsupported Linux distribution"
+        exit
+    ;;
+esac
+
+# clone and build source
+if [ ! -d git-credential-manager ]; then
+    git clone https://github.com/ldennington/git-credential-manager.git || exit
```
mjcheetham, line 94:
```suggestion
    git clone https://github.com/GitCredentialManager/git-credential-manager.git || exit
```

:wink:

src/linux/Packaging.Linux/install-from-source.sh
```diff
@@ -0,0 +1,101 @@
...
+fi
+cd git-credential-manager
+$SUDO_CMD dotnet build ./src/linux/Packaging.Linux/Packaging.Linux.csproj -c Release -p:InstallFromSource=true
+export PATH=$PATH:$HOME/usr/local/bin
```
mjcheetham, line 98: What if the user already has `/usr/local/bin` in their `$PATH`?

src/linux/Packaging.Linux/install-from-source.sh
```diff
@@ -0,0 +1,101 @@
...
+
+# configure gcm
+git-credential-manager-core configure
```
mjcheetham, line 101: I guess this begs the question to if we should update the Debian package to actually run `configure` as a post-install action? Today we don't. I think the install-from-source solution should leave things in the same state as `dpkg -i`.

src/linux/Packaging.Linux/install-from-source.sh
mjcheetham, line 94: Not all distributions have `git` installed by default (and they may have an older version too ofc). Might need to guard against this, and/or update dependency instructions.

src/linux/Packaging.Linux/install-from-source.sh
```diff
@@ -0,0 +1,101 @@
...
+        export DOTNET_ROOT=$(pwd)/.dotnet
+        export PATH=$PATH:$DOTNET_ROOT
+    ;;
+    *)
+        echo "ERROR: Unsupported Linux distribution"
+        exit
+    ;;
+esac
+
+# clone and build source
```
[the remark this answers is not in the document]
mjcheetham: +1 to the point about avoiding a clone if we can.

¯\\\_(ツ)\_/¯ to the script snippet @dscho included :laughing:

src/linux/Packaging.Linux/install-from-source.sh
```diff
@@ -0,0 +1,101 @@
+#!/bin/sh
```
mjcheetham, line 1: Do we want the script print a message explaining what it does, and prompt:

```shell
This script will download, compile, and install Git Credential Manager to:

  ~/usr/local/bin

Git Credential Manager is licensed under the MIT License: https://aka.ms/gcm/license

Do you want to continue? [Y/n]
```

Might hinder automation; so adding a `--yes` option to skip would be required here too.

Could also do all our pre-checks here for missing/required dependencies:

```shell
This script will download, compile, and install Git Credential Manager to:

  ~/usr/local/bin

Git Credential Manager is licensed under the MIT License: https://aka.ms/gcm/license

ERROR: There are missing dependencies required to continue:
 * git
 * curl
Please install these and try again.
```

mjcheetham (commented):
Looks great so far! I had a few ideas for avoiding `--sudo-cmd` and some questions about modifying the `PATH` and Git config.

src/linux/Packaging.Linux/install-from-source.sh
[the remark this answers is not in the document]
dscho: Wat? You think it might be incorrect? 🤣

src/linux/Packaging.Linux/Packaging.Linux.csproj
ldennington: Ooh great suggestion!

src/linux/Packaging.Linux/install-from-source.sh
ldennington: Mmm this is a good point. Perhaps we should avoid using the debian package altogether and use `dotnet-install.sh` for Debian-based distros as well.

src/linux/Packaging.Linux/install-from-source.sh
ldennington: I can try. I've admittedly had a fair amount of trouble with CentOS, but you're right - there's probably a better way.

.github/workflows/validate-install-from-source.yml
ldennington: Lovely! Thank you.

src/linux/Packaging.Linux/install-from-source.sh
```diff
@@ -0,0 +1,101 @@
...
+# Parse script argument that allows users to customize sudo command
+for i in "$@"
+do
+    case "$i" in
+        --sudo-cmd=*)
+        SUDO_CMD="${i#*=}"
+        shift # past argument=value
+        ;;
+    esac
+done
```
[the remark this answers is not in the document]
ldennington: In one of my conversations with @dscho he talked about users potentially wanting to run as `gksudo`, for example, which is why I structured it this way. That being said, I haven't validated anything except the use of `sudo` so perhaps that's more of a future feature. Happy to remove for now.

src/linux/Packaging.Linux/install-from-source.sh
ldennington: OOPS.

src/linux/Packaging.Linux/install-from-source.sh
ldennington: Sure, happy to remove.

src/linux/Packaging.Linux/install-from-source.sh
ldennington: `git` is one of the packages we install in the `install_shared_packages` method, which is called for each distribution. Are you thinking of adding an additional installation mechanism?

src/linux/Packaging.Linux/install-from-source.sh
ldennington: I'm a little confused - we do attempt to install `git` and `curl` for all distributions when we call the `install_shared_packages()` method. Should we not be doing this and instead warn users as you indicated?

src/linux/Packaging.Linux/install-from-source.sh
ldennington: There was nothing in the mirrorlist when I ran this on GH agents. I've updated this to only execute in CI scenarios in my next iteration. Just cleaning things up and will push shortly.

.github/workflows/validate-install-from-source.yml
```diff
@@ -0,0 +1,27 @@
+name: validate-install-from-source
+
+on:
+  push:
+    branches:
+      - main
+      - ifs-test
```
derrickstolee, line 7: (Reminder to drop this line when you are ready at the end.)

.github/workflows/validate-install-from-source.yml
```diff
@@ -0,0 +1,27 @@
...
+        - image: debian
+        - image: linuxmintd/mint20-amd64
+        - image: fedora
+        - image: centos
+        - image: redhat/ubi8
+        - image: alpine
+    container: ${{matrix.vector.image}}
+    steps:
+      - uses: actions/checkout@v1
+      - run: sh "${GITHUB_WORKSPACE}/src/linux/Packaging.Linux/install-from-source.sh" -y
```
derrickstolee, line 27: Do we want to validate that GCM is on the path after this script passes?

src/linux/Packaging.Linux/install-from-source.sh
```diff
@@ -0,0 +1,176 @@
...
+
+# in non-ci scenarios, advertise what we will be doing and
+# give user the option to exit
+if [ ! "$is_ci" ]; then
+    echo "This script will download, compile, and install Git Credential Manager to:
+
+    ~/usr/local/bin
```
derrickstolee, line 24: I was initially confused by this `~`, but you _do_ add `~/usr/local/bin` into the `PATH`, so :+1:.

src/linux/Packaging.Linux/install-from-source.sh
ldennington: Did some investigation here - using `dotnet-install.sh` for Debian-based distros is not a great idea. Apparently there are different package dependencies for different versions, and I don't want us to be in the business of maintaining that. So, I stuck to using the Debian package with the distribution/version in the repository URL.

This meant I had to split out `linuxmint` though, as there's no feed for it. I used the packages corresponding to the latest version here. Given that I don't _think_ many of our users are running linuxmint I feel like this is ok, but this gut feeling could be incorrect. Let me know if changes are recommended to support earlier versions.

README.md
```diff
@@ -89,6 +89,32 @@ sudo /usr/local/share/gcm-core/uninstall.sh
...
+If `curl` is not installed, please use your distribution's package manager
+to install it.
+
+0. To download and run the script:
+
+```shell
+curl -o install-from-source.sh -L https://raw.githubusercontent.com/GitCredentialManager/git-credential-manager/main/src/linux/Packaging.Linux/install-from-source.sh &&
+sh ./install-from-source.sh
+git-credential-manager-core configure
+```
```
derrickstolee, line 112: I tried these instructions on my Ubuntu machine, but with this first command:

```
curl -o install-from-source.sh -L https://raw.githubusercontent.com/ldennington/git-credential-manager/ifs-test/src/linux/Packaging.Linux/install-from-source.sh
```

I got these errors at the end:

```
0 upgraded, 0 newly installed, 0 to remove and 32 not upgraded.
Hit:1 http://us.archive.ubuntu.com/ubuntu impish InRelease
Hit:2 http://us.archive.ubuntu.com/ubuntu impish-updates InRelease                                         
Get:3 https://packages.microsoft.com/ubuntu/21.10/prod impish InRelease [10.5 kB]                          
Hit:4 http://us.archive.ubuntu.com/ubuntu impish-backports InRelease             
Hit:5 http://security.ubuntu.com/ubuntu impish-security InRelease
Get:6 https://packages.microsoft.com/ubuntu/21.10/prod impish/main arm64 Packages [567 B]
Get:7 https://packages.microsoft.com/ubuntu/21.10/prod impish/main amd64 Packages [5,298 B]
Get:8 https://packages.microsoft.com/ubuntu/21.10/prod impish/main armhf Packages [567 B]
Fetched 16.9 kB in 0s (34.7 kB/s)
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
32 packages can be upgraded. Run 'apt list --upgradable' to see them.
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
E: Unable to locate package dotnet-sdk-5.0
E: Couldn't find any package by glob 'dotnet-sdk-5.0'
```

I earlier had this output related to the Microsoft feeds:

```
(Reading database ... 195994 files and directories currently installed.)
Preparing to unpack packages-microsoft-prod.deb ...
Unpacking packages-microsoft-prod (1.0-ubuntu21.10.1) over (1.0-ubuntu21.04.1) ...
Setting up packages-microsoft-prod (1.0-ubuntu21.10.1) ...
Installing new version of config file /etc/apt/sources.list.d/microsoft-prod.list ...
```

```ShellSession
$ cat /etc/os-release
PRETTY_NAME="Ubuntu 21.10"
```

src/linux/Packaging.Linux/install-from-source.sh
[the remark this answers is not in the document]
ldennington: Seemed to work fine for me. I made a few minor tweaks to keep formatting consistent with the rest of the script, but other than that I'm good with it.

README.md
derrickstolee: I updated the script locally to output the URL used for the Microsoft feed: `https://packages.microsoft.com/config/ubuntu/21.10/packages-microsoft-prod.deb` This seems correct to me.

derrickstolee (commented):
I'm really happy with the organization here, but I'm getting a failure on my Ubuntu machine that seems surprising, since it works on the Actions Ubuntu runner.

README.md
ldennington: Mmm this is very odd. Thanks for pointing it out, will investigate.

ldennington: Super weird - latest test run still green for ubuntu:

https://github.com/ldennington/git-credential-manager/runs/5454412044?check_suite_focus=true

derrickstolee: yeah, and it's not that an error is being squelched. I see the package being installed correctly.

ldennington: Set up a 21.04 VM locally and it's still running without issue. I also see the following:

```shell
sudo apt install dotnet-sdk-5.0
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
dotnet-sdk-5.0 is already the newest version (5.0.405-1).
0 upgraded, 0 newly installed, 0 to remove and 238 not upgraded.
```

I'm wondering if something went wrong with an `apt update`? Perhaps try running that and then re-running the script?

derrickstolee: Following up, I found [these instructions](https://docs.microsoft.com/en-us/dotnet/core/install/linux-ubuntu#2110-) which use a different download than you suggested, but more importantly has these steps:

```
sudo apt-get update; \
  sudo apt-get install -y apt-transport-https && \
  sudo apt-get update && \
  sudo apt-get install -y dotnet-sdk-5.0
```

I needed these extra steps to make things work (`apt-transport-https`) as an earlier step, before the dotnet install.

Perhaps this will make it work:

```
        # install dotnet packages and dependencies
        $sudo_cmd apt update
        $sudo_cmd apt install apt-transport-https -y
        $sudo_cmd apt update
        $sudo_cmd apt install dotnet-sdk-5.0 dpkg-dev -y
```

(Of course, now that I have the install progressed at this point, my machine becomes less useful as a test bed.)

derrickstolee: That was also my only issue. The rest of the install succeeded.

src/linux/Packaging.Linux/install-from-source.sh
```diff
@@ -0,0 +1,176 @@
...
+        $sudo_cmd dpkg -i packages-microsoft-prod.deb
+        rm packages-microsoft-prod.deb
+
+        # proactively install tzdata to prevent prompts
+        export DEBIAN_FRONTEND=noninteractive
+        $sudo_cmd apt install -y --no-install-recommends tzdata
+
+        # install dotnet packages and dependencies
+        $sudo_cmd apt update
+        $sudo_cmd apt install dotnet-sdk-5.0 dpkg-dev apt-transport-https -y
```
derrickstolee, line 127:
```suggestion
        # install dotnet packages and dependencies
        $sudo_cmd apt update
        $sudo_cmd apt install apt-transport-https -y
        $sudo_cmd apt update
        $sudo_cmd apt install dotnet-sdk-5.0 dpkg-dev -y
```

(Suggested change based on [official Microsoft docs](https://docs.microsoft.com/en-us/dotnet/core/install/linux-ubuntu#2110-))

README.md
ldennington: This was a great find! Thank you so much, Stolee. #1 dogfooder!! 🐶

ldennington:
Link to the latest validation run with all distros passing!

https://github.com/ldennington/git-credential-manager/actions/runs/1953486987

derrickstolee (approved):

src/linux/Packaging.Linux/install-from-source.sh
dscho: Maybe this is a bug in the CI agents (or the Docker container?) that should be handled at a separate level, e.g. in the GitHub workflow?

src/linux/Packaging.Linux/build.sh
```diff
@@ -59,22 +62,28 @@ if [ -z "$VERSION" ]; then
     die "--version was not set"
 fi
 
-ARCH="`dpkg-architecture -q DEB_HOST_ARCH`"
-if test -z "$ARCH"; then
-  die "Could not determine host architecture!"
+if ! "$INSTALL_FROM_SOURCE"; then
```
dscho, line 65: I _think_ `if ! <string>` is a Bash-ism. It would probably be better to use `if [ -z "$INSTALL_FROM_SOURCE" ]` here, as well as in similar other instances

dscho (approved):
I have just one concern about a potential Bash-ism, otherwise looks good to me!

src/linux/Packaging.Linux/build.sh
ldennington: I unfortunately was not able to use `if [ -z "$INSTALL_FROM_SOURCE" ]`, since we're passing in true or false (which we need to use for our csproj config to work correctly). So, I instead switched to `if [ "$INSTALL_FROM_SOURCE" = false ]` which I believe is ok for shell scripts, but please do correct me if I've misunderstood. I was able to update `if ! $is_ci;` to `if [ -z $is_ci ];` in `install-from-source.sh` though, which is nice.

dscho: Nice!

dscho (approved):

src/linux/Packaging.Linux/install-from-source.sh
mjcheetham: Ah right, you're correct! I had missed that when I wrote this comment - sorry! 🤦‍♂️ 

> Should we not be doing this and instead warn users as you indicated?

I wonder if _always running_ `<pkgmgr> <installcmd> curl git` is a good idea for those who are using custom-built Git's, or other such setups?
We are already doing a check for `dotnet` and the appropriate SDKs/versions before attempting to install, why not also `git` and `curl`?

Even just a simple `which -s git` and `which -s curl` could be enough not to stomp over someones setup? Personally if a script stomped on my setup I'd not be happy 😁

src/linux/Packaging.Linux/install-from-source.sh
```diff
@@ -0,0 +1,173 @@
...
+
+        # proactively install tzdata to prevent prompts
+        export DEBIAN_FRONTEND=noninteractive
+        $sudo_cmd apt install -y --no-install-recommends tzdata
+
+        # install dotnet packages and dependencies
+        $sudo_cmd apt update
+        $sudo_cmd apt install apt-transport-https -y
+        $sudo_cmd apt update
+        $sudo_cmd apt install dotnet-sdk-5.0 dpkg-dev -y
```
mjcheetham, line 129: We always install `dotnet` for Debian or Ubuntu users, but not for other distros (which call `ensure_dotnet_installed()`). Is there a reason not to also test if we already have the deps for Debian?

mjcheetham (commented):
A few questions about detecting dependencies and avoiding installing them.

src/linux/Packaging.Linux/install-from-source.sh
ldennington: Great catch!

src/linux/Packaging.Linux/install-from-source.sh
ldennington: I added a version of this in the latest iteration. Note that I did not use the `-s` flag because that resulted in the following during my testing:

```shell
ldennington@ubuntu:~$ which -s git
Illegal option -s
Usage: /usr/bin/which [-a] args
```

PR #635, opened 2022-03-22
Update to .NET 6

buchwasa:
This PR updates GCM to .NET 6, which is beneficial as it is an LTS and has support for Apple Silicon (which I plan on adding support in another PR)

This PR has been tested running through the provided "test" configuration in VS Code on macOS. Additionally, I did build, install and run to further test it out.

### DIFF
--- a/src/shared/Core/Authentication/MicrosoftAuthentication.cs
+++ b/src/shared/Core/Authentication/MicrosoftAuthentication.cs
@@ -478,7 +478,7 @@ namespace GitCredentialManager.Authentication
 
             return defaultValue;
 #else
-            // OS broker requires .NET Framework right now until we migrate to .NET 5.0 (net6.0-windows10.x.y.z)
+            // OS broker requires .NET Framework right now until we migrate to .NET 5.0 (net5.0-windows10.x.y.z)
             return false;
 #endif
         }
--- a/src/shared/Core/Authentication/MicrosoftAuthentication.cs
+++ b/src/shared/Core/Authentication/MicrosoftAuthentication.cs
@@ -478,7 +478,7 @@ namespace GitCredentialManager.Authentication
 
             return defaultValue;
 #else
-            // OS broker requires .NET Framework right now until we migrate to .NET 5.0 (net5.0-windows10.x.y.z)
+            // OS broker requires .NET Framework right now until we migrate to .NET 5.0 (net6.0-windows10.x.y.z)
             return false;
 #endif
         }
--- a/.azure-pipelines/templates/linux/compile.yml
+++ b/.azure-pipelines/templates/linux/compile.yml
@@ -1,9 +1,9 @@
 steps:
   - task: UseDotNet@2
-    displayName: Use .NET SDK 5.0.203
+    displayName: Use .NET SDK 6.0.201
     inputs:
       packageType: sdk
-      version: 5.0.203
+      version: 6.0.201
 
   - task: DotNetCoreCLI@2
     displayName: Compile common code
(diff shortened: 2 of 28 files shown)


Lutherp77:
Luther Phillips

Lutherp77:
Luther Phillips

Lutherp77:
Luther Phillips

Lutherp77:
Luther Phillips

Lutherp77:
Luther Phillips

src/shared/Core/Authentication/MicrosoftAuthentication.cs
```diff
@@ -478,7 +478,7 @@ public static bool CanUseBroker(ICommandContext context)
 
             return defaultValue;
 #else
-            // OS broker requires .NET Framework right now until we migrate to .NET 5.0 (net6.0-windows10.x.y.z)
+            // OS broker requires .NET Framework right now until we migrate to .NET 5.0 (net5.0-windows10.x.y.z)
```
ldennington, line 481: What is the reason to _not_ update the comment?

ldennington (approved):
Thanks so much for your contribution @buchwasa ⭐ ! I took a look at the code and just left one super nitpicky comment. Looks like macOS CI failed on the first run though - I've queued another run in the hopes the issue was intermittent, but it will need further investigation if not.

Finally, I spoke to @mjcheetham about #636. This is something we definitely want, but are going to hold on merging for a bit while we wait for this change to settle. We also have some followup comments, but I'll drop them on that PR for you to address.

src/shared/Core/Authentication/MicrosoftAuthentication.cs
buchwasa: I don't really have good reasoning, when I noticed the change, I just kind of thought that it would be a good idea to leave that comment alone. I can revert that if you'd like.

derrickstolee: It's more of a dig to clean up this `#if`, so I'm happy to leave it. (Do we still have some need to support compiling on .NET Framework?)

The change makes the comment more consistent, so I'd say feel free to leave it as you have it.

ldennington: Yes, we have to compile on .NET Framework on Windows (see the note in [this section](https://github.com/GitCredentialManager/git-credential-manager/blob/main/docs/architecture.md#overview) of `Architecture.md`).

mjcheetham (approved):
This looks great @buchwasa, thanks!

Just one minor request would be to fix up the newly introduced build warning we'd now get from the .NET 6 SDK:

```
NETSDK1179: One of '--self-contained' or '--no-self-contained' options are required when '--runtime' is used.
```

We need to explicitly specify `--self-contained` for each `dotnet publish` step we're doing in the _src/osx/Installer.Mac/layout.sh_ script, for example:

```diff
 dotnet publish "$GCM_SRC" \
 	--no-restore \
 	-m:1 \
 	--configuration="$CONFIGURATION" \
 	--framework="$FRAMEWORK" \
 	--runtime="$RUNTIME" \
+ 	--self-contained \
 	--output="$(make_absolute "$PAYLOAD")" || exit 1
```

This is what we're already doing in the Linux _src/linux/Packaging.Linux/build.sh_ script.

buchwasa:
Thanks for pointing that out @mjcheetham! I saw it a bit ago and I guess it slipped my mind to get that fixed, sorry about that.

04161999 (commented):
Display Name inputs

Version 6.0.201

PR #638, opened 2022-03-24
Remove build number from the release artifacts

### LINKED ISSUE #577: Use full version in git tags, or remove the build number from the artifacts to ensure they match
This would simplify a lot of scripts that automates the installation of GCM Core, especially on Linux. This is because, for any other project, it's common to use something like the following:

```bash
latest_version=$(basename "$(curl -fsSL -o /dev/null -w "%{url_effective}" https://github.com/docker/buildx/releases/latest)")

curl -fsSL "https://github.com/docker/buildx/releases/download/${latest_version}/buildx-${latest_version}.darwin-amd64"
```

But the problem of doing something similar with GCM Core is that the git tag does not match the version in the artifact name, and the build number is not guessable. I'm solving this issue by relying on a external service: https://github.com/Contextualist/glare

```bash
latest_version=$(basename "$(curl -fsSL -o /dev/null -w "%{url_effective}" https://github.com/GitCredentialManager/git-credential-manager/releases/latest)")

curl -fsSL "https://glare.vercel.app/GitCredentialManager/git-credential-manager@${latest_version}/gcmcore-linux_amd64.+.tar.gz$"
```

But it would be super cool if I could remove glare from the GCM Core installation flow.

This is especially a problem, because [there is no way to use a APT repository and thus leveraging the version resolution to APT itself.](https://github.com/GitCredentialManager/git-credential-manager/issues/447)

felipecrs:
Fixes #577

### DIFF
--- a/.azure-pipelines/templates/osx/pack.signed/step3-pack.yml
+++ b/.azure-pipelines/templates/osx/pack.signed/step3-pack.yml
@@ -18,14 +18,14 @@ steps:
   - script: nbgv cloud --common-vars
     displayName: Set version variables
 
-  - script: src/osx/Installer.Mac/pack.sh --payload='$(Build.StagingDirectory)/payload' --version='$(GitBuildVersion)' --output='$(Build.StagingDirectory)/components/com.microsoft.gitcredentialmanager.component.pkg'
+  - script: src/osx/Installer.Mac/pack.sh --payload='$(Build.StagingDirectory)/payload' --version='$(GitBuildVersionSimple)' --output='$(Build.StagingDirectory)/components/com.microsoft.gitcredentialmanager.component.pkg'
     displayName: Create component package
 
-  - script: src/osx/Installer.Mac/dist.sh --package-path='$(Build.StagingDirectory)/components' --version='$(GitBuildVersion)' --output='$(Build.StagingDirectory)/pkg/gcmcore-osx-$(GitBuildVersion).pkg' || exit 1
+  - script: src/osx/Installer.Mac/dist.sh --package-path='$(Build.StagingDirectory)/components' --version='$(GitBuildVersionSimple)' --output='$(Build.StagingDirectory)/pkg/gcmcore-osx-$(GitBuildVersionSimple).pkg' || exit 1
     displayName: Create product archive
 
   - task: PublishPipelineArtifact@0
     displayName: Upload unsigned package
     inputs:
       artifactName: 'tmp.macinstaller_unsigned'
-      targetPath: '$(Build.StagingDirectory)/pkg/gcmcore-osx-$(GitBuildVersion).pkg'
+      targetPath: '$(Build.StagingDirectory)/pkg/gcmcore-osx-$(GitBuildVersionSimple).pkg'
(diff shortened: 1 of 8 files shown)


felipecrs:
I wonder if this is all we need. :)

felipecrs:
I think this should do the trick. :)

ldennington (commented):
Thanks so much for doing this @felipecrs ⭐ ! I have just a little bit of feedback:

1. On reviewing the CI results, it looks like there's a little syntactical issue in `Setup.iss` that needs to be fixed up.
2. Would you mind re-organizing your commits a bit? I'm thinking if you put the Inno changes first, it'll be easier for reviewers to understand where `GitBuildVersionSimple` comes from. Additionally, if you could add some more details to the `Use another approach` commit (i.e. what is the new approach you're using?) that would also be helpful. (As a heads up, you can do this pretty easily with [`git rebase -i`](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History).
3. Were you able to validate this change? I know doing this with release-related stuff can be tricky, so let us know if you need assistance on that front. Or if you were able to verify that packages produced now contain the simplified version, then I'm fine with moving forward after the above suggestions have been implemented.

felipecrs:
> 2\. Would you mind re-organizing your commits a bit? I'm thinking if you put the Inno changes first, 
it'll be easier for reviewers to understand where `GitBuildVersionSimple` comes from. Additionally, if you could add some more details to the `Use another approach` commit (i.e. what is the new approach you're using?) that would also be helpful. (As a heads up, you can do this pretty easily with [`git rebase -i`](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History).

Sure, I can do it. I always expect my PRs to be squashed during merge, so I don't worry much about the intermediate commit messages, that's why.

Working on it.

felipecrs:
@ldennington thanks a lot for the review. I pushed a commit to fix the InnoSetup syntax error, but the PR itself isn't ready yet. I have to address the `rebase -i` stuff and also check how I can test this locally.

For now, please don't spend any time reviewing it. I'll let you know when it's ready to review again.

felipecrs:
@ldennington actually `GitBuildVersionSimple` comes from _Nerdbank.GitVersioning_:

https://github.com/dotnet/Nerdbank.GitVersioning/blob/master/doc/cloudbuild.md#set-special-build-variables-for-use-in-subsequent-build-steps

And not from the Inno setup. This means that the order of commits is correct.

I also corrected the commit messages. Please let me know if it's ok now. If not, I can also squash them all in a single one if you prefer.

Now, about testing, I was unfortunately not able to test the changes by myself. I would appreciate help in this front (perhaps the CI loops are enough?)

ldennington:
I think triggering a test release to see the updates to our artifacts would be a good idea. However, this is problematic for two reasons:

1. Our release pipeline is in a private Azure DevOps account, so you cannot see it.
2. While I have permissions to at least see the pipeline, I do not have permissions to trigger it.

Perhaps @mjcheetham will be kind enough to help us out by triggering a release against the `remove-build-number` branch? (Note: I've already pushed it to the main repository).

ldennington (approved):

felipecrs:
@ldennington thank a lot for your support in this PR! :)

PR #648, opened 2022-04-04
Fix install from source Debian compile bug

ldennington:
Fixes #644

A Linux user reported an oddly small Debian package associated with the 2.0.692 release. This was caused by the refactoring of `build.sh` in #630. The fix is fairly straightforward - we have to ensure our binaries are copied to the expected location before attempting to create the package. This change implements that fix, and also fixes a component of the symlink path that was also mistakenly dropped in #630.

Here is the set of artifacts from before the fix was implemented, with the Debian package at an unlikely size of 940 B:

https://dev.azure.com/mseng/AzureDevOps/_build/results?buildId=17061102&view=artifacts&pathAsName=false&type=publishedArtifacts

Here is the set of artifacts from a test build I ran after the fix was implemented, with a more expected package size of 107 MB:

https://dev.azure.com/mseng/AzureDevOps/_build/results?buildId=17076015&view=artifacts&pathAsName=false&type=publishedArtifacts

PR #650, opened 2022-04-05
fixup! fixup! Add install from source script

ldennington:
While #648 led to the creation of a Debian package of the expected size, I realized during testing of this package that there was still a problem. After installing the package locally and running `git-credential-manager-core configure`, I saw the following error:

`git-credential-manager-core: command not found`

This is due to the fact that we need to create the symlink between `usr/local/bin/git-credential-manager-core` and `usr/local/share/git-credential-manager-core` _before_ we create the package. After applying this fix and building the package locally, the `git-credential-manager-core configure` command completed successfully.


derrickstolee (approved):

dscho (approved):
Good catch!

PR #607, opened 2022-01-12
Issue/573 OAuth2 support for Bitbucket DC

mminns:
(Bloody Hell it works.)

Bitbucket DC 7.20 now supports OAuth 2.0 connections from external applications.
https://confluence.atlassian.com/bitbucketserver/bitbucket-data-center-and-server-7-20-release-notes-1101934428.html

This means OAuth 2.0 access_tokens can be used as authentication between Git and Bitbucket DC.

The PR refactors the Bitbucket module provide a more generic framework for supporting OAuth 2.0 and providing implementations for both Bitbucket Cloud and Bitbucket DC as they are unfortunately not exactly the same.

Because each Bitbucket DC installation is a unique instance the OAuth 2.0 ClientId and ClientSecret are unique to each installation and therefore GCM must be configured to support each Bitbucket DC instance.

Documentation is updated to describe how to configure and use the feature.

Tests are extended to cover Bitbucket DC.

I've been dogfooding it for the last month.

PR #237, opened 2020-11-26
Add first support for Linux arm64

dennisameling:
Currently, trying to install Git Credential Manager Core on Linux arm64 (e.g. WSL2 on the Surface Pro X or Raspberry Pi 4) results in the following error:

```shell
dennis@DESKTOP-8HTP3NV:~/repos$ sudo dpkg -i gmcore.deb
dpkg: error processing archive gmcore.deb (--install):
 package architecture (amd64) does not match system (arm64)
Errors were encountered while processing:
 gmcore.deb
```
 
This PR adds support for Linux arm64 builds, both when building natively on arm64 or while cross-compiling from an x86 host (in which case you can now run `dotnet build -c LinuxDebug -r linux-arm64`).

Can confirm it works on my arm64 Surface Pro X through WSL2 Ubuntu :tada::

```shell
dennis@DESKTOP-8HTP3NV:~/repos/gmcoretest$ sudo dpkg -i gcmcore-linux_arm64.2.0.301.22928.deb
[sudo] password for dennis:
Selecting previously unselected package gcmcore.
(Reading database ... 40440 files and directories currently installed.)
Preparing to unpack gcmcore-linux_arm64.2.0.301.22928.deb ...
Unpacking gcmcore (2.0.301.22928) ...
Setting up gcmcore (2.0.301.22928) ...
```
```shell
dennis@DESKTOP-8HTP3NV:~/repos/gmcoretest$ git-credential-manager-core --version
Git Credential Manager version 2.0.301-beta+90596ecdf4 (Linux, .NET Core 3.1.10)
```

Things that still need to happen AFAIK:
- Add support for signing multiple `.deb` files at once in `.github/run_esrp_signing.py`, it currently only looks at the first `.deb` file it finds. How does that sound?
- Running tests on arm64. This is possible e.g. with Travis (https://docs.travis-ci.com/user/multi-cpu-architectures/#multi-cpu-availaibility); please let me know if only running tests on amd64 is enough for now or whether you'd like me to set up a Travis config for running tests on arm64 as well.

PR #668, opened 2022-04-15
Lint markdown

wolf99:
Making this change with multiple commits to make it easier to revert any parts that people don't like 🙂 
I have some questions before making later commits.
In a final commit, I have introduced a GitHub workflow that adds a markdown lint step so that the linting can be easily maintained.

This PR will probably cause merge conflicts with  #237, #551, and #607

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -67,12 +67,12 @@ members of the project's leadership.
 
 ## Attribution
 
-This Code of Conduct is adapted from the [Contributor Covenant][homepage], version 1.4,
-available at [Contributor Covenant Code of Conduct][coc].
+This Code of Conduct is adapted from the [Contributor Covenant][cc-homepage], version 1.4,
+available at [Contributor Covenant Code of Conduct][cc-coc].
 
 For answers to common questions about this code of conduct, see the
-[Contributor Covenant FAQ][faq]
+[Contributor Covenant FAQ][cc-faq]
 
-[homepage]: https://www.contributor-covenant.org
-[coc]: https://www.contributor-covenant.org/version/1/4/code-of-conduct.html
-[faq]: https://www.contributor-covenant.org/faq
+[cc-homepage]: https://www.contributor-covenant.org
+[cc-coc]: https://www.contributor-covenant.org/version/1/4/code-of-conduct.html
+[cc-faq]: https://www.contributor-covenant.org/faq
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -68,9 +68,11 @@ members of the project's leadership.
 ## Attribution
 
 This Code of Conduct is adapted from the [Contributor Covenant][homepage], version 1.4,
-available at https://www.contributor-covenant.org/version/1/4/code-of-conduct.html
+available at [Contributor Covenant Code of Conduct][coc].
+
+For answers to common questions about this code of conduct, see the
+[Contributor Covenant FAQ][faq]
 
 [homepage]: https://www.contributor-covenant.org
-
-For answers to common questions about this code of conduct, see
-https://www.contributor-covenant.org/faq
+[coc]: https://www.contributor-covenant.org/version/1/4/code-of-conduct.html
+[faq]: https://www.contributor-covenant.org/faq
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,6 +1,6 @@
-## Contributing
+# Contributing
 
-[issue]: https://github.com/GitCredentialManager/git-credential-manager/issues 
+[issue]: https://github.com/GitCredentialManager/git-credential-manager/issues
 [fork]: https://github.com/GitCredentialManager/git-credential-manager/fork
 [pr]: https://github.com/GitCredentialManager/git-credential-manager/compare
 [code-of-conduct]: CODE_OF_CONDUCT.md
@@ -13,24 +13,24 @@ Please note that this project is released with a [Contributor Code of Conduct][c
 
 ## Start with an issue
 
-0. Open an [issue][issue] to discuss the change you want to see.
+1. Open an [issue][issue] to discuss the change you want to see.
 This helps us coordinate and reduce duplication.
-0. Once we've had some discussion, you're ready to code!
+2. Once we've had some discussion, you're ready to code!
 
 ## Submitting a pull request
 
-0. [Fork][fork] and clone the repository
-0. Configure and install the dependencies: `dotnet restore`
-0. Make sure the tests pass on your machine: `dotnet test`
-0. Create a new branch: `git switch -c my-branch-name`
-0. Make your change, add tests, and make sure the tests still pass
-0. For UI updates, test your changes by executing a `dotnet run` in applicable UI-related project directories:
+1. [Fork][fork] and clone the repository
+2. Configure and install the dependencies: `dotnet restore`
+3. Make sure the tests pass on your machine: `dotnet test`
+4. Create a new branch: `git switch -c my-branch-name`
+5. Make your change, add tests, and make sure the tests still pass
+6. For UI updates, test your changes by executing a `dotnet run` in applicable UI-related project directories:
     - `Atlassian.Bitbucket.UI.Avalonia`
     - `GitHub.UI.Avalonia`
     - `Atlassian.Bitbucket.UI.Windows`
     - `GitHub.UI.Windows`
-0. Push to your fork and [submit a pull request][pr]
-0. Pat your self on the back and wait for your pull request to be reviewed and merged.
+7. Push to your fork and [submit a pull request][pr]
+8. Pat your self on the back and wait for your pull request to be reviewed and merged.
 
 Here are a few things you can do that will increase the likelihood of your pull request being accepted:
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,22 +15,22 @@ Please note that this project is released with a [Contributor Code of Conduct][c
 
 1. Open an [issue][issue] to discuss the change you want to see.
 This helps us coordinate and reduce duplication.
-2. Once we've had some discussion, you're ready to code!
+1. Once we've had some discussion, you're ready to code!
 
 ## Submitting a pull request
 
 1. [Fork][fork] and clone the repository
-2. Configure and install the dependencies: `dotnet restore`
-3. Make sure the tests pass on your machine: `dotnet test`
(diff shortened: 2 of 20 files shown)


README.md
```diff
@@ -85,8 +85,6 @@ sudo /usr/local/share/gcm-core/uninstall.sh
 
 ---
 
-<!-- this explicit anchor should stay stable so that external docs can link here -->
-<a name="linux-install-instructions"></a>
```
wolf99, line 89: Is this actually linked from external docs? If so, is there a commitment not to break this specific link for some reason

Docs (anywhere) can link to headings in markdown using the `url#heading` syntax, as in `docs/wsl.md` in this commit.

wolf99, line 89: Is this actually linked from external docs?

If so, is there a commitment not to break this specific link for some reason?

Docs (anywhere) can link to headings in markdown using the `url#heading` syntax, as in `docs/wsl.md` in this commit. There isn't usually any need to add additional, explicit anchors for markdown headings.

wolf99:
I have a couple more commits to do, so moving this to draft until they're ready.

wolf99:
both `docs/environment.md` and `docs/configuration.md` contain many multiple headings with the same content.

The first contains many headings of:

- "Example"
- "Windows"
- "macOS/Linux"

While the latter contains many headings of "Example"

There are some options to resolve this:

1. Restructure these documents  so as to avoid this heading pattern
2. Add configuration to markdown lint CI to ignore these forever

Which is preferred?

README.md
vtbassmatt: It's linked from at least the [GitHub docs](https://docs.github.com/en/get-started/getting-started-with-git/caching-your-github-credentials-in-git#git-credential-manager) on the topic (you may have to click over to the Linux tab) and possibly other vendors as well.

wolf99: Thanks! I'll experiment with a few options for solving this in a lint-friendly way then.

Are these docs rendered only on GitHub or also I  other locations that I should test the results against?

vtbassmatt: If you mean the docs in this repo, no – we don't render them anyplace else.

CONTRIBUTING.md
```diff
@@ -13,24 +13,24 @@ Please note that this project is released with a [Contributor Code of Conduct][c
 
 ## Start with an issue
 
-0. Open an [issue][issue] to discuss the change you want to see.
+1. Open an [issue][issue] to discuss the change you want to see.
 This helps us coordinate and reduce duplication.
-0. Once we've had some discussion, you're ready to code!
+2. Once we've had some discussion, you're ready to code!
```
ldennington, line 18: Are we able to add an exception instead of making this change? We tend to use the `1`/`0` formatting on purpose, as it makes it easier to add new items in lists (i.e. we don't have to go change all the other numbers when we put something new in the middle).

CODE_OF_CONDUCT.md
```diff
@@ -68,9 +68,11 @@ members of the project's leadership.
...
-available at https://www.contributor-covenant.org/version/1/4/code-of-conduct.html
+available at [Contributor Covenant Code of Conduct][coc].
 
-[homepage]: https://www.contributor-covenant.org
+For answers to common questions about this code of conduct, see the
+[Contributor Covenant FAQ][faq]
 
-For answers to common questions about this code of conduct, see
-https://www.contributor-covenant.org/faq
+[homepage]: https://www.contributor-covenant.org
```
ldennington, line 76: Nit: maybe `cc-homepage` or `cc_homepage`?

docs/bitbucket-development.md
```diff
@@ -65,18 +65,18 @@ The Access and Refresh Tokens will be stored against the username and the userna
 
 # On-Premise Bitbucket
 
-On-premise Bitbucket, more correctly known as Bitbucket Server or Bitbucket DC, has a number of differences compared to the cloud instance of Bitbucket, https://bitbucket.org.
+On-premise Bitbucket, more correctly known as Bitbucket Server or Bitbucket DC, has a number of differences compared to the cloud instance of Bitbucket, [bitbucket.org](https://bitbucket.org).
```
ldennington, line 68: Should we have short names for links in this document similar to those added in `CODE_OF_CONDUCT.md`?

docs/gitlab.md
```diff
@@ -41,6 +41,6 @@ git config --global credential.gitlabauthmodes browser
 
 Improved support requires changes in GitLab. Please vote for these issues if they affect you:
 
-1. No support for OAuth device authorization (necessary for machines without web browser) https://gitlab.com/gitlab-org/gitlab/-/issues/332682
-2. Only domains with prefix `gitlab.` are recognised as GitLab remotes https://gitlab.com/gitlab-org/gitlab/-/issues/349464
-3. Username/password authentication is suggested even if disabled on server https://gitlab.com/gitlab-org/gitlab/-/issues/349463
+1. No support for OAuth device authorization (necessary for machines without web browser): [GitLab issue 332682](https://gitlab.com/gitlab-org/gitlab/-/issues/332682)
```
ldennington, line 44: Should we have short names for the links in this document similar to those added in `CODE_OF_CONDUCT.md`?

ldennington (approved):
Thanks for working to make our docs better! ✏️

ldennington:
> Two more general issues remain:
> 
> 1. Both [`docs/environment.md`](https://github.com/GitCredentialManager/git-credential-manager/blob/main/docs/environment.md) and [`docs/configuration.md`](https://github.com/GitCredentialManager/git-credential-manager/blob/main/docs/configuration.md) contain many multiple headings with the same content.
>    The first contains many headings of:
>    
>    * "Example"
>    * "Windows"
>    * "macOS/Linux"
>    
>    While the latter contains many headings of "Example"
>    There are some options to resolve this:
>    
>    1. Restructure these documents  so as to avoid this heading pattern
>    2. Add configuration to markdown lint CI to ignore these forever (this option is currently implemented in the workflow introduced in this PR)
>    
>    Which is preferred?
> 2. There appears to not be a uniform approach to line length. It looks like some files break on a given character column length (something < 80), while others break at the end of sentence.
>    My own opinion is that the latter is better for readability and more in the spirit of markdown itself, but as with most things in programming it is more important that any approach be uniform and I am happy to implement any approach that is preferred.
>    The linting workflow introduced in this PR is currently configured to ignore line length.
>    Is there a preference?



> Two more general issues remain:
> 
> 1. Both [`docs/environment.md`](https://github.com/GitCredentialManager/git-credential-manager/blob/main/docs/environment.md) and [`docs/configuration.md`](https://github.com/GitCredentialManager/git-credential-manager/blob/main/docs/configuration.md) contain many multiple headings with the same content.
>    The first contains many headings of:
>    
>    * "Example"
>    * "Windows"
>    * "macOS/Linux"
>    
>    While the latter contains many headings of "Example"
>    There are some options to resolve this:
>    
>    1. Restructure these documents  so as to avoid this heading pattern
>    2. Add configuration to markdown lint CI to ignore these forever (this option is currently implemented in the workflow introduced in this PR)
I think it is fine to ignore in this case. I don't think having multiple of the same heading in one file is a huge deal.
>    
>    Which is preferred?
> 2. There appears to not be a uniform approach to line length. It looks like some files break on a given character column length (something < 80), while others break at the end of sentence.
>    My own opinion is that the latter is better for readability and more in the spirit of markdown itself, but as with most things in programming it is more important that any approach be uniform and I am happy to implement any approach that is preferred.
>    The linting workflow introduced in this PR is currently configured to ignore line length.
>    Is there a preference?
My preference would be to break on a character column link. If we decide to do this, I'm good with doing < 80.

README.md
wolf99: Turns out GitHub Flavored Markdown does not support custom heading links in any form.
So I've ended up simply adding an "ignore"  for the specific linting rule for this anchor (in 8cf643a)

CONTRIBUTING.md
wolf99: I've always felt this reasoning was overrated; if someone is making changes to a document anyway, then it's not hard to change a handful of list numbers as well (for relatively short lists). It's can also get a little confusing when working with sub-lists and sub-sub lists, when they are all the same number.

The benefit of proper numbering is that the markdown provides the same information in raw form as in rendered form.

Nevertheless, I have updated the docs to use the form of repeated `1.`, as this form is accepted by the linter (repeated `0.` also works but means that lists start at 0 rather than 1), in commit 1b24d82

CODE_OF_CONDUCT.md
wolf99: Good call!
Added in 7f3f110

docs/bitbucket-development.md
wolf99: Yes! 🙂 
I will work on doing this for all markdown files.

At the same time I will replace `aka.ms` links that refer to other docs within this repo with relative links, so as to:

1. Remove dependency on MS short link service
2. Allow links to work offline.

Please let me know if this latter part is not wanted.

wolf99:
Thank you for the review @ldennington !

There are two things remaining in progress

1. Line lengths
2. Reference style links

I will continue to work on these, but perhaps this can be merged as it currently is, with those as separate PRs?

The reason being it is already a large change and I am aware of other Rs that also have doc changes, I wouldn't like this PR to block these changes or vice-versa. The line-length one especially may be a lot of change across many files.

wolf99:
Rebased and fixed merge conflicts

CONTRIBUTING.md
ldennington: Thank you!

docs/bitbucket-development.md
ldennington: I am ok with the latter part, but I think @mjcheetham may want to weigh in on this.

ldennington:
> Thank you for the review @ldennington !
> 
> There are two things remaining in progress
> 
> 1. Line lengths
> 2. Reference style links
> 
> I will continue to work on these, but perhaps this can be merged as it currently is, with those as separate PRs?
> 
> The reason being it is already a large change and I am aware of other PRs that also have doc changes, I wouldn't like this PR to block these changes cause too many merge conflicts to those or vice-versa. Both of the above changes will be a lot of change across many files.

I'm fine with this!

> Thank you for the review @ldennington !
> 
> There are two things remaining in progress
> 
> 1. Line lengths
> 2. Reference style links
> 
> I will continue to work on these, but perhaps this can be merged as it currently is, with those as separate PRs?
> 
> The reason being it is already a large change and I am aware of other PRs that also have doc changes, I wouldn't like this PR to block these changes cause too many merge conflicts to those or vice-versa. Both of the above changes will be a lot of change across many files.

Sounds great! I just did another review, and it all looks good. I will proceed with the merge.

PR #674, opened 2022-04-19
Add dependabot

wolf99:
#668 will add usage of a GitHub action (markdownlint) to the repo workflows. This PR adds dependabot so as to keep this an any other action up to date.

I am not familiar enough with nuget and dot net projects to know if it is worth adding nuget package-ecosystem (or others).

If someone in the know could comment here I will be happy to adjust as needed.
Info on dependabot usage can be seen here: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/about-dependabot-version-updates

PR #678, opened 2022-04-27
Add dependabot

Luciferic666:
I am not familiar enough with nuget and dot net projects to know if it is worth adding nuget package-ecosystem (or others).

If someone in the know could comment here I will be happy to adjust as needed.
Info on dependabot usage can be seen here: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/about-dependabot-version-updates


dscho:
This was a bogus version of #674.

wolf99:
Why would someone even do this?! SMH

PR #593, opened 2021-12-31
Add link to GitHub application settings

hickford:

PR #679, opened 2022-04-28
Add GitHub.com OAuth app revoke instructions to FAQs

mjcheetham:
_This PR is based on #593 to address remaining feedback comments._

Add a link to the application registration for the GitHub.com OAuth app registration, and include an entry in the FAQs about how to revoke consent.


derrickstolee (approved):
Thanks for this! It's good to have this reference.

Issue #606, opened 2022-01-09
Make it easier to build for ARM


ankitbko:
As of now code changes are required to build for ARM64 as detailed in #237 . The `build.sh` has also changed since the PR was created and now has pinned runtime variable. https://github.com/GitCredentialManager/git-credential-manager/blob/859900272fae3d3b89e39cf85ffdcf28003f5afc/src/linux/Packaging.Linux/build.sh#L53 

Ask is to parameterize as much as possible so even though it is not officially supported, users can build a version for ARM without having to make all these code changes.

Surya-06:
@mjcheetham I would like to work on this issue. Please let me know if this is alright and any pointers to get started.

ldennington:
@Surya-06 thank you for volunteering to work on this issue 🌟! You are more than welcome to start whenever you have spare time to tackle it. As for pointers, I would take a look at the [`build.sh` script](https://github.com/GitCredentialManager/git-credential-manager/blob/main/src/linux/Packaging.Linux/build.sh) and the [dotnet documentation on Linux runtimes](https://docs.microsoft.com/en-us/dotnet/core/rid-catalog#linux-rids).

Surya-06:
thanks @ldennington , I'll try to find some time and close this.

Surya-06:
@ldennington I took a fresh pull today and somehow when I just run `dotnet -c build LinuxRelease` or `dotnet -c build LinuxDebug` it builds just fine (earlier I got multiple errors when restoring the solution files and build would fail). I also faced issues with dotnet 5 vs 6 etc. Can you please confirm if what I'm seeing is right? I can't remember setting any persistent variable which could've affected subsequent builds etc. and noticed that there was a change which upgraded the dotnet version. PS: I'm building on a Rpi 4B.

ldennington:
@Surya-06 - I'm a bit confused about what the ask is. Are you still having problems building? If so, could you paste in the error text you see?

Surya-06:
[...]

PR #685, opened 2022-05-03
Pass through $PATH to find dotnet

easwarh:
This commit makes it easier to build on Linux from source. Related to #606, but
applies to x64 builds as well.

Signed-off-by: Easwar Hariharan <easwar.hariharan@microsoft.com>

### DIFF
--- a/src/linux/Packaging.Linux/build.sh
+++ b/src/linux/Packaging.Linux/build.sh
@@ -106,9 +106,13 @@ else
     mkdir -p "$INSTALL_LOCATION"
 fi
 
+if [ -z "$DOTNET_ROOT" ]; then
+    DOTNET_ROOT="$(dirname $(which dotnet))"
+fi
+
 # Publish core application executables
 echo "Publishing core application..."
-dotnet publish "$GCM_SRC" \
+$DOTNET_ROOT/dotnet publish "$GCM_SRC" \
 	--configuration="$CONFIGURATION" \
 	--framework="$FRAMEWORK" \
 	--runtime="$RUNTIME" \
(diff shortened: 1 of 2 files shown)


mjcheetham (changes requested):
Hello 👋!

Thanks for the contribution. I'm in favour of making it easier to build GCM on more platforms, but I have one concern about this change as it stands.

The `$DOTNET_ROOT` variable is set by the `dotnet` app host executable, which is typically the caller of these scripts (via MSBuild). However in the cases that this script is called directly (not via `dotnet`) then the variable is not set, and the commands being run in this script will be `/dotnet publish ...` which will not resolve correctly.

If we are to accept this then there should be a check that if the `$DOTNET_ROOT` variable is not set, then the script should fall back to invoking `dotnet` and let the shell resolve the executable location.. 
..OR add something like this to explicitly resolve:

```shell
if [ -z "$DOTNET_ROOT" ]; then
    DOTNET_ROOT="$(dirname $(which dotnet))"
fi
```

easwarh:
Hi @mjcheetham, that's a great call out, thanks for the feedback! I'll fix this up, would you rather have a fixup commit (commit --fixup), a new commit, or an amendment to the existing commit (commit --amend)?

ldennington:
@easwarh - I know you directed your question at Matthew, but FWIW my usual workflow in GCM is to amend my existing commits.

easwarh:
> @easwarh - I know you directed your question at Matthew, but FWIW my usual workflow in GCM is to amend my existing commits.

Thanks for the headsup. I'll push a merged commit.

mjcheetham (approved):
Looks good to me!

src/linux/Packaging.Linux/build.sh
```diff
@@ -106,9 +106,13 @@ else
...
 fi
 
+if [ -z "$DOTNET_ROOT" ]; then
+    DOTNET_ROOT="$(dirname $(which dotnet))"
+fi
+
 # Publish core application executables
 echo "Publishing core application..."
-dotnet publish "$GCM_SRC" \
+$DOTNET_ROOT/dotnet publish "$GCM_SRC" \
```
Osama2535, line 115: Oatef4313@gmail.com

PR #699, opened 2022-05-13
Gitlab hostname only support.

### LINKED ISSUE #698: GitLabHostProvider ignoring gitlab hosts if addressed by hostname only.

2.0.696+4365b917da for windows.

* [x] GitLab

* [x] Yes

**SetUp / Infrastructure**

We have an internal on premise gitlab instance that is reachable from within our companies network.
The host is configured to set its hostname only in the clone url. The `domain.tld` are omitted.
Therefore cloning any repository from that instance will result in an origin name of `hostname` e.g. `abcgitlab`.
The server is reachable internally either by `abcgitlab` as well as `abcgitlab.abc.lan`.
The gitlab instance was configured to allow GCM OAuth usage as described in [GitLab support](https://github.com/GitCredentialManager/git-credential-manager/blob/main/docs/gitlab.md).
Local credential configurations exist for both `https://abcgitlab` as well as `https://abcgitlab.abc.lan` containing the same settings.

**Expected behavior**

I was expecting to be forwarded to the web browser to grant access to the application GCM.

**Actual behavior**

GCM is falling back to generic authentication instead of configured OAuth for gitlab on premise.
GCM (or git) also modifies my configuration file and sets my `provider` value from `gitlab` back to `generic`.

**Logs**

Where do I find them? Windows system (for work), but could try to reproduce on private linux machines.
Would need to setup described infrastructure from above first though.

**Personal Analysis**

I assume the reason is due to [GitLabHostProvider.cs:65](https://github.com/GitCredentialManager/git-credential-manager/blob/46d1b3cc31a385b44f071a37a888c563c83adb59/src/shared/GitLab/GitLabHostProvider.cs#L65).
Changing the remote url of the git repository to have at least three parts works as expected but is kind of cumbersome as the gitlab server is providing clone urls without `domain.tld`.

For tests I compiled my own adaption of the GitLab-Hostprovider and changed mentioned checks to require just one element as well as comparing to the name of our host.
This resulted in `GitLabDevClientId` and `GitLabDevClientSecret` being searched and read from the unscoped `credential` section only.
I find this kind of confusing as the one thing should have nothing to do with the other.
Any `GitLabDevClient*` values in the then appropriate `https://abcgitlab` section (instead of the previously used `https://abcgitlab.abc.lan` section) simply were ignored.
Therefore I am wondering whether the fully qualified domain restriction is there due to a side effect or whether it is there for a (good) reason.

Me personally would prefer to be able to address our gitlab instance with its host name only omitting `domain.tld`.
Also I'd be willing to contribute an implementation for this. But for this, I'd need to understand what the purpose of certain decissions made were, so I can follow the overall idea of the plugin/provider.

Bene81:
Closes #698 .

This merge request offers support for addressing git hosts by hostname only.

The overall idea of this implementation is: if a host uri does not contain any dots it has to be a host in the local network.
If thats the case the resulting credential scope is returned.

Otherwise the old scope determination algorithm applies.
UnitTest has been adapted. I am open for discussions or improvement suggestions.

### DIFF
--- a/src/shared/Core/UriExtensions.cs
+++ b/src/shared/Core/UriExtensions.cs
@@ -102,6 +102,16 @@ namespace GitCredentialManager
                 }
             }
 
+            // check whether the host specification contains of hostname only
+            // this usually means, the git hosting instance is part of your local
+            // network.
+            if (!(string.IsNullOrWhiteSpace(host) || string.IsNullOrEmpty(host)) &&
+                !host.Contains("."))
+            {
+                yield return $"{schemeAndDelim}{host}";
+                yield break;
+            }
+
             // Unfold the host by sub-domain, left-to-right
             while (!string.IsNullOrWhiteSpace(host))
             {
--- a/src/shared/Core.Tests/UriExtensionsTests.cs
+++ b/src/shared/Core.Tests/UriExtensionsTests.cs
@@ -27,7 +27,7 @@ namespace GitCredentialManager.Tests
         }
 
         [Theory]
-        [InlineData("http://com")]
+        [InlineData("http://hostname", "http://hostname")]
         [InlineData("http://example.com",
             "http://example.com")]
         [InlineData("http://foo.example.com",


src/shared/Core/UriExtensions.cs
```diff
@@ -102,6 +102,16 @@ public static IEnumerable<string> GetGitConfigurationScopes(this Uri uri)
...
             }
 
+            // check whether the host specification contains of hostname only
+            // this usually means, the git hosting instance is part of your local
+            // network.
+            if (!(string.IsNullOrWhiteSpace(host) || string.IsNullOrEmpty(host)) &&
+                !host.Contains("."))
+            {
+                yield return $"{schemeAndDelim}{host}";
+                yield break;
```
ldennington, line 112: Can you clarify why we're using `yield break`? Something like:

```suggestion
                // If we have reached this point, there are no more subdomains to unfold, so exit early.
                yield break;
```

src/shared/Core/UriExtensions.cs
```diff
@@ -102,6 +102,16 @@ public static IEnumerable<string> GetGitConfigurationScopes(this Uri uri)
                 }
             }
 
+            // check whether the host specification contains of hostname only
+            // this usually means, the git hosting instance is part of your local
+            // network.
+            if (!(string.IsNullOrWhiteSpace(host) || string.IsNullOrEmpty(host)) &&
```
ldennington, line 108:
```suggestion
            if (!string.IsNullOrWhiteSpace(host) &&
```

src/shared/Core/UriExtensions.cs
```diff
@@ -102,6 +102,16 @@ public static IEnumerable<string> GetGitConfigurationScopes(this Uri uri)
                 }
             }
 
+            // check whether the host specification contains of hostname only
+            // this usually means, the git hosting instance is part of your local
+            // network.
```
ldennington, line 107:
```suggestion
            // Check whether the URL only contains hostname.
            // This usually means the host is on your local network.
```

ldennington (changes requested):
Thank you for opening this PR! 🎉  I've added a few suggestions for cleanup, but looks good overall.

Bene81:
Thanks for your suggestions. I have reviewed and applied them.
Just for my personal understanding: why is it guarenteed that the URL hostname/domain name is either `null` or `whitespace` but never `string.Empty`?

mjcheetham:
> why is it guarenteed that the URL hostname/domain name is either `null` or `whitespace` but never `string.Empty`?

The `string.IsNullOrWhiteSpace` method also checks for the empty string, and is therefore a superset of the tests used in `string.IsNullOrEmpty`. [C# documentation](https://docs.microsoft.com/en-us/dotnet/api/system.string.isnullorwhitespace?view=net-6.0)

Value|`string.IsNullOrEmpty`|`string.IsNullOrWhiteSpace`
-|-|-
`"foo"`|`false`|`false`
`null`|`true`|`true`
`""`<br/>`string.Empty`|`true`|`true`
`"   "`|`false`|`true`

Bene81:
Thanks for hinting. Sorry, should have figured that out myself.

ldennington (approved):

PR #700, opened 2022-05-14
Add broken link checking Action

wolf99:
Also adjusts the entire workflow to only run if a PR or push contains changes to a markdown file.

This will require the addition of an access token in the repository, if one does not already exist. The token is used to avoid rate-limiting by GitHub when the Lychee Action uses the GitHub API. Further info in the comments in the change.


wolf99:
Lots of URLs that are in code blocks are failing. My understanding is that Lychee should automatically ignore code blocks, so I have raised an issue on the Lychee Action: https://github.com/lycheeverse/lychee/issues/618 .

Moving this to draft until this is resolved.

wolf99:
Oh man, godaddy and globalsign websites suck.

wolf99:
@ldennington the link checker fails this GitLab link (https://github.com/GitCredentialManager/git-credential-manager/blob/main/docs/faq.md?plain=1#L158) because it is only accessible to a logged in user. I could try to tell the checker to ignore it, but there would then be no warning if GitLab were to change the URL.

Options:

1. Ignore it and take the risk. This is the same as the current situation where there is no link checker
2. Remove the URL and rely on the prose description only. This suffers the same problem really as the prose would need updating if GitLab changed where these settings are.

As the risk is essentially the same either way, I suggest option 1. I would like to get your thought on it, as maybe there are other options that I am not aware of?

wolf99:
I have raised https://github.com/lycheeverse/lychee-action/issues/136 to understand how to mitigate the issue with the `file:///...` link.

ldennington:
> As the risk is essentially the same either way, I suggest option 1. I would like to get your thought on it, as maybe there are other options that I am not aware of?

I agree that we should go with option 1. I did a quick search, and I'm not seeing any instructions for revoking OAuth Application access in GitLab's docs, which means we'll need to stick with these DIY instructions. Although the link checker thinks this is problematic, the docs are ultimately intended for our users, and I think they will find the link helpful 🙂.

wolf99:
Rebased and now ignores the specific GitLab URL as agreed, so this is good to go.
Apologies for the delay, I've been moving across the country 🙂 .

mjcheetham (approved):

mjcheetham:
Thanks for this! Broken links are a pain which should be avoided now :-)

PR #708, opened 2022-05-18
Fix install from source validation

ldennington:
#705 exposed a previously-obfuscated issue in our `validate-install-from-source` pipeline. We were not cloning the repo with its full history, which broke `Nerdbank.GitVersioning` (see failure [here](https://github.com/GitCredentialManager/git-credential-manager/actions/runs/2346353840)). This change fixes the issue (check out [this successful test](https://github.com/ldennington/git-credential-manager/actions/runs/2346516061) run in my fork).

### DIFF
--- a/.github/workflows/validate-install-from-source.yml
+++ b/.github/workflows/validate-install-from-source.yml
@@ -23,6 +23,8 @@ jobs:
     container: ${{matrix.vector.image}}
     steps:
       - uses: actions/checkout@v3
+        with:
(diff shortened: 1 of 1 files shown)


payephyoeaung87 (commented):
Thank you I'm responsible for my shortcomings Gihub I apologize to the organization. Again, thank you very much.

payephyoeaung87:
I can only say thank you for the review because of my lack of qualifications.

PR #717, opened 2022-05-31
Manually scan `$PATH` on POSIX systems 

### LINKED ISSUE #671: which isn't always located in /usr/bin
In PosixEnvironment.TryLocateExecutable, the program attemtps to execute `/usr/bin/which` which isn't always located there.

In my system running NixOS, it is actually located in `/nix/store/z8rqsaqqgnjxld3kf180k0h1p7zn1wv3-system-path/bin/which` (however this can change) and `/usr/bin` only contains `env`.

The hardcoded path:
https://github.com/GitCredentialManager/git-credential-manager/blob/2ab9429da68614c30f3d67782115700ddcf0040c/src/shared/Core/Interop/Posix/PosixEnvironment.cs#L34-L36

mjcheetham:
Because the `which` command is not always located in `/usr/bin` on all POSIX systems, we cannot directly call `/usr/bin/which` to do program location for us. Oh the irony..

We might be able to use `env` to locate `which`, but again the `env` utility isn't always found in the same place, so we're stuck.

Instead we now manually scan the `$PATH` variable directories looking for the program we need to find, in the same way we do on Windows (for different reasons).

This should also be slightly faster as you'd hope .NET's `String.Split` method and `lstat` calls are faster than `fork`/`exec`-ing another executable?

Fixes #671

### DIFF
--- a/src/shared/Core.Tests/EnvironmentTests.cs
+++ b/src/shared/Core.Tests/EnvironmentTests.cs
@@ -1,4 +1,6 @@
+using System;
 using System.Collections.Generic;
+using GitCredentialManager.Interop.Posix;
 using GitCredentialManager.Interop.Windows;
 using GitCredentialManager.Tests.Objects;
 using Xunit;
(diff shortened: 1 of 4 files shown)


src/shared/Core/EnvironmentBase.cs
```diff
@@ -104,6 +103,38 @@ public virtual Process CreateProcess(string path, string args, bool useShellExec
 
             return new Process { StartInfo = psi };
         }
+
+        public bool TryLocateExecutable(string program, out string path)
```
ldennington, line 107: Awesome! This method is super easy to understand, and it's great we were able to consolidate the UNIX and Windows logic.

src/shared/Core.Tests/EnvironmentTests.cs
```diff
@@ -67,5 +68,66 @@ public void WindowsEnvironment_TryLocateExecutable_Windows_ExistsMultiple_Return
...
+
+            Assert.True(actualResult);
+            Assert.Equal(expectedPath, actualPath);
+        }
+
+        [PlatformFact(Platforms.Posix)]
+        public void PosixEnvironment_TryLocateExecutable_ExistsMultiple_ReturnTrueAndFirstPath()
+        {
+            string pathVar = "/home/john.doe/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin";
+            string execName = "foo";
```
ldennington, line 113: Nit: consider making these instance variables or constants to avoid repetition.

src/shared/Core/EnvironmentBase.cs
```diff
@@ -104,6 +103,38 @@ public virtual Process CreateProcess(string path, string args, bool useShellExec
...
+            // find it! Oh the irony..
+            // We could also try using "env" to then locate "which", but the same problem exists in
+            // that "env" isn't always in a standard location.
+            //
+            // On Windows we should avoid using the equivalent utility "where.exe" because this will
+            // include the current working directory in the search, and we don't want this.
+            //
+            // The upshot of the above means we cannot use either of "which" or "where.exe" and must
+            // instead manually scan the PATH variable looking for the program.
+            // At least both Windows and UNIX use the same name for the $PATH or %PATH% variable!
```
ldennington, line 120: Nit: we seem to be using POSIX in some places (the commit message) and UNIX in others (this comment). Is it ok to use them interchangeably in this way?

ldennington (approved):
This is awesome! 🚀 Added a few nitpicky comments, but overall great work!

src/shared/Core/EnvironmentBase.cs
mjcheetham: Yeah, I've not been very strict in the past about POSIX vs UNIX. POSIX is a subset of UNIX-like behaviour/implementation.

Perhaps this should probably be "Unix" and use class names like `UnixEnvironment` etc because, especially in this case, POSIX doesn't _strictly_ define the file system path formats beyond [this](https://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap03.html#tag_03_266).

VMS paths can be pretty different from UNIX paths. I know @dscho has more opinions on UNIX vs POSIX.

In GCM there are various mapped C Standard Library APIs that are under `src/Core/Interop/Posix/Native` because they're in the POSIX standard, but then there are implementation-specific-isms that live in `Interop/Linux/Native` and `Interop/MacOS/Native`.

mjcheetham:
@dscho I'd like your opinion on this change.

We're moving from `/usr/bin/which` to manually scanning `$PATH` (split on `:` and search for `<path>/<program>` left to right until we find a file).

My original assumption here was that `/usr/bin/env` was also not standardised (like the location of `which` is not), but seems like it may be?

In either case, do you think it's better to stick with this change (scanning `$PATH` manually) or change to using `/usr/bin/env` to help locate things by doing `/usr/bin/env which <program>`?

dscho:
> We're moving from `/usr/bin/which` to manually scanning `$PATH` (split on `:` and search for `<path>/<program>` left to right until we find a file).

Makes sense to me. It's not only better-controlled code, it's also substantially quicker than spawning a program.

> My original assumption here was that `/usr/bin/env` was also not standardised (like the location of `which` is not), but seems like it may be?

The location of `/usr/bin/env` is pretty much guaranteed by the billions of scripts out there that have `#!/usr/bin/env <blablabla>` as first line, where `<blablabla>` is anything from `python` to `perl` to `ruby`.

If anybody would install `env` anywhere else, all those scripts would fail to work.

> In either case, do you think it's better to stick with this change (scanning `$PATH` manually) or change to using `/usr/bin/env` to help locate things by doing `/usr/bin/env which <program>`? (Bearing in mind we're having to manually scan `%PATH%` on Windows anyway.)

Honestly, I'd rather scan `$PATH` manually. That gives us much better control (for example, we can choose _not_ to respect `PATHEXT` and decide that we're only interested in `.exe` files).

dscho:
> > In either case, do you think it's better to stick with this change (scanning `$PATH` manually) or change to using `/usr/bin/env` to help locate things by doing `/usr/bin/env which <program>`? (Bearing in mind we're having to manually scan `%PATH%` on Windows anyway.)
> 
> Honestly, I'd rather scan `$PATH` manually. That gives us much better control (for example, we can choose _not_ to respect `PATHEXT` and decide that we're only interested in `.exe` files).

Obviously this comment does not _quite_ apply here because this PR is about POSIX systems, but I still like the manual approach better.

src/shared/Core.Tests/EnvironmentTests.cs
mjcheetham: Done!

PR #719, opened 2022-05-31
Add `read_repository` GitLab OAuth Scope

### LINKED ISSUE #669: [GitLab] Token provided by GCM for gitlab.com does not work for Source Link request.
GCM for GitLab uses the `repository_write` scope for its tokens. This provides read/write access for Git over HTTP operations but does not provide repository read operations via the GitLab API. This means that GCM does not provide tokens that can be used to authenticate Source Link requests.

This isn't a bug, per se, as GCM is using the correct scope for pure git.exe operations. This is its intended purpose. It is, however, a missed opportunity. Visual Studio leverages GCM to authenticate Source Link requests, and expanding the scope used could allow Visual Studio to use Source Link for private gitlab.com repos, which has been a gap for several releases. 

A more in depth write up can be found here: https://github.com/dotnet/sourcelink/issues/281#issuecomment-1100426378

I am hoping somebody from GCM or GitLab can speak to this. My testing of GCM token's against GitLab API requests makes me **believe** that the scope is the issue, but I'm not 100% sure.

Thanks!

ldennington:
# Overview

Add `read_repository` to supported OAuth Scopes for GitLab. Because `read_repository` grants permissions to read files on private projects using the Repository Files API, it will enable Visual Studio to use Source Link to download files from private GitLab repositories 🎉. The Git Credential Manager GitLab OAuth Application has also been updated with the `read_repository` scope to work with this change.

# 🚨 Important Notes 🚨 

In order to use the new `read_repository` scope, users will need to do the following with existing OAuth apps for their instances:

1. Update them to have the `read_repository` scope.
2. Require users to revoke the application by doing the following:
     1. Navigate to https://gitlab.com/-/profile/applications
     2. Scroll to **Authorized applications**
     3. Click the "Revoke" button next to the name of their instance's app (**Git Credential Manager** is used here for demonstration purposes)
         Screen Shot 2022-05-31 at 2 13 30 PM
     5. Wait for a notification stating **The application was revoked access.**
        Screen Shot 2022-05-31 at 2 15 49 PM
     6. Re-authorize the application with the new scope (GCM should handle this flow for them)

Additionally, although `read_repository` works for the [API required by source link](https://docs.gitlab.com/ee/api/repository_files.html#get-raw-file-from-repository), our testing did uncover that there are certain APIs ([listing a repository tree](https://docs.gitlab.com/ee/api/repositories.html#list-repository-tree), for example) that require additional scopes (in this case, `read_api`). These scopes may also need to be added in the future.

# Testing

I did the following to test these changes:

1. Create a new test OAuth application with `write_repository` and `read_repository` scopes.
2. Update `GitLabDevClientId` and `GitLabDevClientSecret` in my global Git config with the test Application ID and Secret.
3. Remove existing GitLab access/refresh tokens from Keychain.
5. Update [this line](https://github.com/GitCredentialManager/git-credential-manager/blob/f5bd57908b17e7772dde33d84c3a2634fb9d4768/src/shared/GitLab/GitLabHostProvider.cs#L15) to contain `read_repository` locally and build and install GCM with this new scope.
6. Clone a repo (to force re-authentication and add new access/refresh tokens to Keychain).
7. Copy the new access token from Keychain.
8. Attempt this CURL request.
```console
curl "https://gitlab.com/api/v4/projects/36617582/repository/files/README.md/raw?access_token=<COPIED TOKEN>"
```
9. Verified contents of `README.md` were printed to my terminal.

Fixes #669


ldennington:
@Bene81 as FYI.

mjcheetham (approved):
We may want to add the instructions on how to revoke access to the FAQs, or a link from an FAQ to a separate "how to revoke access on GitLab" markdown doc.

hickford:
> 1. Update the OAuth app to have the read_repository scope.
> 2. Require users to revoke the application 

Is step 2 necessary? What happens if the user doesn't revoke?

PR #721, opened 2022-06-02
Fix failing GitHub Action Script

ldennington:
While [testing the `arm64` updates to our release workflow](https://github.com/ldennington/git-credential-manager/runs/6700234325?check_suite_focus=true), I noticed that the GitHub Actions script to publish the release had mysteriously started failing. It turns out that #715 was the culprit - as of `v5`, `actions/github-script` updated GitHub API method calls to contain the word `rest`. [Here](https://github.com/ldennington/git-credential-manager/runs/6702218352?check_suite_focus=true) is an example of a successful test run with these changes.


dscho (approved):
Good catch!

PR #731, opened 2022-06-09
Ignore Homebrew Git shim

### LINKED ISSUE #676: when brew upgrade: "fatal: Failed to unset Git configuration entry 'credential.https://dev.azure.com.useHttpPath' [1]"
Full output:

```
=> Upgrading 1 outdated package:
microsoft/git/git-credential-manager-core 2.0.632.34631 -> 2.0.696
==> Upgrading git-credential-manager-core
==> Downloading https://github.com/GitCredentialManager/git-credential-manager/releases/do
==> Downloading from https://objects.githubusercontent.com/github-production-release-asset
######################################################################## 100.0%
==> Running uninstall script /usr/local/share/gcm-core/uninstall.sh
Password:
Unconfiguring credential helper...
Unconfiguring component 'Git Credential Manager'...
Unconfiguring component 'Azure Repos provider'...
fatal: Failed to unset Git configuration entry 'credential.https://dev.azure.com.useHttpPath' [1]
git: This shim is internal and must be run via brew.
```

Seemed to upgrade successfully, but just wanted to share this FYI.

ldennington:
Users running `brew uninstall` or `brew upgrade` currently see the following message (indicating that the operation wasn't fully successful):

```
Unconfiguring component 'Azure Repos provider'...
fatal: Failed to unset Git configuration entry 'credential.https://dev.azure.com.useHttpPath' [1]
git: This shim is internal and must be run via brew.
```

This is because GCM is finding the Homebrew Git shim on the PATH and calling it instead of the user's Git installation. Specifying that we should ignore the Homebrew shim when attempting to locate the Git executable on the PATH resolves the issue.

Fixes #676.

### DIFF
--- a/src/osx/Installer.Mac/uninstall.sh
+++ b/src/osx/Installer.Mac/uninstall.sh
@@ -12,7 +12,7 @@ fi
 
 # Unconfigure (as the current user)
 echo "Unconfiguring credential helper..."
-sudo -u `/usr/bin/logname` "$GCMBIN" unconfigure
+sudo -u `/usr/bin/logname` -E "$GCMBIN" unconfigure
 
 # Remove symlink
 if [ -L /usr/local/bin/git-credential-manager-core ]
--- a/src/shared/Core.Tests/EnvironmentTests.cs
+++ b/src/shared/Core.Tests/EnvironmentTests.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using GitCredentialManager.Interop.Posix;
 using GitCredentialManager.Interop.Windows;
 using GitCredentialManager.Tests.Objects;
@@ -119,6 +120,32 @@ namespace GitCredentialManager.Tests
             var env = new PosixEnvironment(fs, envars);
 
             bool actualResult = env.TryLocateExecutable(PosixExecName, out string actualPath);
+
+            Assert.True(actualResult);
+            Assert.Equal(expectedPath, actualPath);
+        }
+
+        [PlatformFact(Platforms.MacOS)]
+        public void MacOSEnvironment_TryLocateExecutable_Paths_Are_Ignored()
+        {
+            List<string> pathsToIgnore = new List<string>()
+            {
+                "/home/john.doe/bin/foo"
+            };
+            string expectedPath = "/usr/local/bin/foo";
+
+            var fs = new TestFileSystem
+            {
+                Files = new Dictionary<string, byte[]>
+                {
+                    [pathsToIgnore.FirstOrDefault()] = Array.Empty<byte>(),
+                    [expectedPath] = Array.Empty<byte>(),
+                }
+            };
+            var envars = new Dictionary<string, string> {["PATH"] = PosixPathVar};
+            var env = new PosixEnvironment(fs, envars);
+
+            bool actualResult = env.TryLocateExecutable(PosixExecName, pathsToIgnore, out string actualPath);
 
             Assert.True(actualResult);
             Assert.Equal(expectedPath, actualPath);
--- a/src/shared/Core/EnvironmentBase.cs
+++ b/src/shared/Core/EnvironmentBase.cs
@@ -104,7 +104,12 @@ namespace GitCredentialManager
             return new Process { StartInfo = psi };
         }
 
-        public bool TryLocateExecutable(string program, out string path)
+        public virtual bool TryLocateExecutable(string program, out string path)
+        {
+            return TryLocateExecutable(program, null, out path);
+        }
+
+        internal virtual bool TryLocateExecutable(string program, ICollection<string> pathsToIgnore, out string path)
         {
             // On UNIX-like systems we would normally use the "which" utility to locate a program,
             // but since distributions don't always place "which" in a consistent location we cannot
@@ -124,7 +129,8 @@ namespace GitCredentialManager
                 foreach (var basePath in paths)
                 {
                     string candidatePath = Path.Combine(basePath, program);
-                    if (FileSystem.FileExists(candidatePath))
+                    if (FileSystem.FileExists(candidatePath) && (pathsToIgnore is null ||
+                        !pathsToIgnore.Contains(candidatePath, StringComparer.OrdinalIgnoreCase)))
                     {
                         path = candidatePath;
                         return true;
--- a/src/shared/Core/Interop/MacOS/MacOSEnvironment.cs
+++ b/src/shared/Core/Interop/MacOS/MacOSEnvironment.cs
@@ -1,0 +1,38 @@
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Threading;
+using GitCredentialManager.Interop.Posix;
+
+namespace GitCredentialManager.Interop.MacOS
+{
+    public class MacOSEnvironment : PosixEnvironment
+    {
+        private ICollection<string> _pathsToIgnore;
+
+        public MacOSEnvironment(IFileSystem fileSystem)
+            : base(fileSystem) { }
+
+        internal MacOSEnvironment(IFileSystem fileSystem, IReadOnlyDictionary<string, string> variables)
+            : base(fileSystem)
+        {
+            EnsureArgument.NotNull(variables, nameof(variables));
+            Variables = variables;
+        }
+
+        public override bool TryLocateExecutable(string program, out string path)
+        {
+            if (_pathsToIgnore is null)
+            {
+                _pathsToIgnore = new List<string>();
+                if (Variables.TryGetValue("HOMEBREW_PREFIX", out string homebrewPrefix))
+                {
+                    string homebrewGit = Path.Combine(homebrewPrefix, "Homebrew/Library/Homebrew/shims/shared/git");
+                    _pathsToIgnore.Add(homebrewGit);
+                }
+            }
+            return TryLocateExecutable(program, _pathsToIgnore, out path);
+        }
+    }
+}
--- a/src/shared/Core/CommandContext.cs
+++ b/src/shared/Core/CommandContext.cs
@@ -107,7 +107,7 @@ namespace GitCredentialManager
                 FileSystem        = new MacOSFileSystem();
                 SessionManager    = new MacOSSessionManager();
                 SystemPrompts     = new MacOSSystemPrompts();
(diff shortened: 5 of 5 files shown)


src/shared/Core/EnvironmentBase.cs
```diff
@@ -104,7 +104,12 @@ public virtual Process CreateProcess(string path, string args, bool useShellExec
             return new Process { StartInfo = psi };
         }
 
-        public bool TryLocateExecutable(string program, out string path)
+        public virtual bool TryLocateExecutable(string program, out string path)
+        {
+            return TryLocateExecutable(program, new List<string>(), out path);
+        }
+
+        public bool TryLocateExecutable(string program, List<string> pathsToIgnore, out string path)
```
mjcheetham, line 112: Make this method `protected virtual` and accept an `ICollection<string>` instead

```suggestion
        protected virtual bool TryLocateExecutable(string program, ICollection<string> pathsToIgnore, out string path)
```

src/shared/Core/EnvironmentBase.cs
```diff
@@ -104,7 +104,12 @@ public virtual Process CreateProcess(string path, string args, bool useShellExec
             return new Process { StartInfo = psi };
         }
 
-        public bool TryLocateExecutable(string program, out string path)
+        public virtual bool TryLocateExecutable(string program, out string path)
+        {
+            return TryLocateExecutable(program, new List<string>(), out path);
```
mjcheetham, line 109: Pass `null` here instead of creating a new heap object

```suggestion
            return TryLocateExecutable(program, null, out path);
```

src/shared/Core/EnvironmentBase.cs
```diff
@@ -124,7 +129,8 @@ public bool TryLocateExecutable(string program, out string path)
                 foreach (var basePath in paths)
                 {
                     string candidatePath = Path.Combine(basePath, program);
-                    if (FileSystem.FileExists(candidatePath))
+                    if (FileSystem.FileExists(candidatePath) &&
+                        !pathsToIgnore.Contains(candidatePath, StringComparer.OrdinalIgnoreCase))
```
mjcheetham, line 133: Update to also permit possible `null` arg

```suggestion
                    if (FileSystem.FileExists(candidatePath) && (pathsToIgnore is null ||
                        !pathsToIgnore.Contains(candidatePath, StringComparer.OrdinalIgnoreCase)))
```

src/shared/Core/Interop/MacOS/MacOSEnvironment.cs
```diff
@@ -0,0 +1,34 @@
...
+        {
+            string homebrewPrefix = Environment.GetEnvironmentVariable("HOMEBREW_PREFIX");
+            List<string> pathsToIgnore = new List<string>();
+
+            if (homebrewPrefix != null)
+            {
+                pathsToIgnore.Add($"{homebrewPrefix}/Homebrew/Library/Homebrew/shims/shared/git");
+            }
+
+            return TryLocateExecutable(program, pathsToIgnore, out path);
```
mjcheetham, line 31: Could we possibly cache this ignore list to avoid needing to build it on each invocation? Also we should probably use both the `Variables` member to search for variables and `Path.Combine` to combine the paths.

Unit tests initialise the envars field directly but `Environment.GetEnvironmentVariable()` will look at the _real_ environment.

```suggestion
            if (_pathsToIgnore is null)
            {
                _pathsToIgnore = new List<string>();

                if (Variables.TryGetValue("HOMEBREW_PREFIX", out string homebrewPrefix))
                {
                    string homebrewGit = Path.Combine(homebrewPrefix, "Homebrew/Library/Homebrew/shims/shared/git");
                    _pathsToIgnore.Add(homebrewGit);
                }
            }

            return TryLocateExecutable(program, _pathsToIgnore, out path);
```


..where `_pathsToIgnore` is a private `ICollection<string>` field.

src/shared/Core/Interop/MacOS/MacOSEnvironment.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        public MacOSEnvironment(IFileSystem fileSystem)
+            : this(fileSystem, GetCurrentVariables()) { }
+
+        internal MacOSEnvironment(IFileSystem fileSystem, IReadOnlyDictionary<string, string> variables)
+            : base(fileSystem)
+        {
+            EnsureArgument.NotNull(variables, nameof(variables));
+            Variables = variables;
+        }
```
mjcheetham, line 19: We don't need to replicate the constructor of `PosixEnvironment` here – just call through to the corresponding `base` constructor.

```suggestion
        public MacOSEnvironment(IFileSystem fileSystem)
            : base(fileSystem) { }

        internal MacOSEnvironment(IFileSystem fileSystem, IReadOnlyDictionary<string, string> variables)
            : base(fileSystem, variables) { }
```

src/shared/Core/Interop/Posix/PosixEnvironment.cs
```diff
@@ -34,7 +34,7 @@ protected override string[] SplitPathVariable(string value)
 
         #endregion
 
-        private static IReadOnlyDictionary<string, string> GetCurrentVariables()
+        protected static IReadOnlyDictionary<string, string> GetCurrentVariables()
```
mjcheetham, line 37: We don't need to make this `protected` if we don't need now to call `GetCurrentVariables()` in derived types 😉

src/shared/Core.Tests/EnvironmentTests.cs
```diff
@@ -123,5 +124,31 @@ public void PosixEnvironment_TryLocateExecutable_ExistsMultiple_ReturnTrueAndFir
...
+
+            var fs = new TestFileSystem
+            {
+                Files = new Dictionary<string, byte[]>
+                {
+                    [pathsToIgnore.FirstOrDefault()] = Array.Empty<byte>(),
+                    [expectedPath] = Array.Empty<byte>(),
+                }
+            };
+            var envars = new Dictionary<string, string> {["PATH"] = PosixPathVar};
```
mjcheetham, line 145: Here `PosixPathVar` is `/home/john.doe/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin` which doesn't contain the `/home/john.doe/bin/ignore/` directory anyway, so this test will always return `true` since the existing code will not see the `ignore/` dir as existing.

Instead create two empty files at `/home/john.doe/bin/foo` and `/usr/local/bin/foo`, and then set `pathsToIgnore` as `/home/john.doe/bin/foo`.

We already have a test that verifies we should normally select the `john.doe` version ahead of the `usr/local` one, so this test will only pass if the ignored paths feature is correct.

src/osx/Installer.Mac/uninstall.sh
```diff
@@ -12,7 +12,7 @@ fi
 
 # Unconfigure (as the current user)
 echo "Unconfiguring credential helper..."
-sudo -u `/usr/bin/logname` "$GCMBIN" unconfigure
+sudo -u `/usr/bin/logname` HOMEBREW_PREFIX=$HOMEBREW_PREFIX "$GCMBIN" unconfigure
```
mjcheetham, line 15: Aha! The reason `HOMEBREW_PREFIX` wasn't coming through to the GCM process was because we're going via `sudo`. Perhaps we should instead use the following option to preserve _all_ environment variables?

```man
-E, --preserve-env
                   Indicates to the security policy that the user wishes to preserve their existing
                   environment variables.  The security policy may return an error if the user does not have
                   permission to preserve the environment.

```

mjcheetham (changes requested):
A few comments on using `sudo -E` instead and caching the ignore list.

src/shared/Core/Interop/MacOS/MacOSEnvironment.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        public override bool TryLocateExecutable(string program, out string path)
+        {
+            string homebrewPrefix = Environment.GetEnvironmentVariable("HOMEBREW_PREFIX");
+            List<string> pathsToIgnore = new List<string>();
+
+            if (homebrewPrefix != null)
+            {
+                pathsToIgnore.Add($"{homebrewPrefix}/Homebrew/Library/Homebrew/shims/shared/git");
+            }
```
MikeMcQuaid, line 29: Could probably just consider ignoring _any_ path which ends in `Homebrew/Library/Homebrew/shims/shared/git`. That'll save you needing to have `HOMEBREW_PREFIX` be present or passed.

src/shared/Core/EnvironmentBase.cs
ldennington: The UT started complaining with `protected` so I went with `internal` instead.

src/shared/Core.Tests/EnvironmentTests.cs
ldennington: I _think_ I'm picking up what you're putting down. I was at least able to re-configure the test so that it was failing and returning the `/home/john.doe/bin/foo` before I passed `pathsToIgnore`. It then started passing after I passed `pathsToIgnore`. Let me know if further changes are needed, though.

src/shared/Core/Interop/MacOS/MacOSEnvironment.cs
ldennington: I'm going to roll with the `-E` change Matthew suggested above. Can't hurt to have access to all the `HOMEBREW` variables in future.

src/shared/Core.Tests/EnvironmentTests.cs
mjcheetham: Looks good!

mjcheetham (approved):
:shipit:

src/shared/Core/Interop/MacOS/MacOSEnvironment.cs
```diff
@@ -0,0 +1,38 @@
...
+            EnsureArgument.NotNull(variables, nameof(variables));
+            Variables = variables;
+        }
+
+        public override bool TryLocateExecutable(string program, out string path)
+        {
+            if (_pathsToIgnore is null)
+            {
+                _pathsToIgnore = new List<string>();
+                if (Variables.TryGetValue("HOMEBREW_PREFIX", out string homebrewPrefix))
```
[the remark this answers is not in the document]
ldennington: Comment added! Thanks so much again for all your help here, Mike!

PR #739, opened 2022-06-13
Avoid writing credentials when account & password have not changed

mjcheetham:
Do not needlessly serialise or otherwise write out the credential again if nothing has changed from an existing stored credential.

This update is applied to all stores except "cache" because values are short lived/not persisted anyway.


ldennington (approved):
Nice! ⭐

PR #748, opened 2022-06-16
Fix bug in Windows credential manager credential matching on port numbers

### LINKED ISSUE #747: [bug] Reversed logic in WindowsCredentialManager for matching port number
Looking at the WindowsCredentialManager code, I spotted a logic error when matching the port number in a stored credential:
https://github.com/GitCredentialManager/git-credential-manager/blob/3ff096f91ce09d11ecbaca1e1f55f31cf2aece8e/src/shared/Core/Interop/Windows/WindowsCredentialManager.cs#L296

Currently if the port numbers match, the IsMatch function returns not a match, and vice versa. This could lead to confusing authentication issues if a user were to clone a repo specifying a port number.

mjcheetham:
Fix a bug whereby we would not correctly match existing credentials when port numbers were used in the search. Additionally, actually now store the non-default port number in the target URI if there is one given.

Fixes #747

### DIFF
--- a/src/shared/Core/Interop/Windows/WindowsCredentialManager.cs
+++ b/src/shared/Core/Interop/Windows/WindowsCredentialManager.cs
@@ -9,7 +9,7 @@ namespace GitCredentialManager.Interop.Windows
 {
     public class WindowsCredentialManager : ICredentialStore
     {
-        private const string TargetNameLegacyGenericPrefix = "LegacyGeneric:target=";
+        internal const string TargetNameLegacyGenericPrefix = "LegacyGeneric:target=";
 
         private readonly string _namespace;
 
@@ -260,7 +260,7 @@ namespace GitCredentialManager.Interop.Windows
             return url;
         }
 
-        private bool IsMatch(string service, string account, Win32Credential credential)
+        internal /* for testing */ bool IsMatch(string service, string account, Win32Credential credential)
         {
             // Match against the username first
             if (!string.IsNullOrWhiteSpace(account) &&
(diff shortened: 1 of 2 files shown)


src/shared/Core/Interop/Windows/WindowsCredentialManager.cs
```diff
@@ -339,6 +345,11 @@ private string CreateTargetName(string service, string account)
...
             }
 
+            if (!serviceUri.IsDefaultPort)
+            {
+                sb.AppendFormat(":{0}", serviceUri.Port);
+            }
+
             if (!string.IsNullOrWhiteSpace(serviceUri.AbsolutePath.TrimEnd('/')))
             {
                 sb.Append(serviceUri.AbsolutePath);
```
mjcheetham, line 355: @ldennington the failing unit test case fails because of this (arguable) error in the code. Today we only effectively trim the trailing '/' if there is no path in the URI.

I.e,

```text
https://example.com/ ==> https://github.com
https://example.com/path/ ==> https://github.com/path/
```

I think we should probably be doing as the test is looking for (to always trim), but for people who have stored creds against the full path already (using `credential.useHttpPath=true`, we will be storing/updating a second credential (one without a slash) and then depending on the _order_ we look up existing credentials, match against _both_.

This may leave users with expired creds that we are not deleting but keep recalling.

What do you think we should do here?

mjcheetham: Spoke offline.. updated as per new revision and commit message:

```text
We also now always trim any trailing '/' from the end of the path of a
service/target URI. Previously we erronously only trimmed the '/' from a
'pathless' URI, i.e.:

  https://example.com/      => https://example.com/
  https://example.com/path/ => https://example.com/path/

This is a safe change since the `Enumerate` method that is used by both
`Get` and `Remove` already matches _both_ trailing and non-trailing
slashes by virtue of creating and comparing components of a `System.Uri`
which does normalisation of the path.

Both `Get` and `Remove` would act (eventually) to remove any bad
credentials stored with a trailing slash, even if new credentials are
only written out without the slash.
```

ldennington (approved):
Ran through on my Windows machine - looks great 🌟.

huajie2020 (approved):

PR #758, opened 2022-06-23
Use /usr/bin/env which instead of /usr/bin/which

echuber2:
It may cause trouble to hard-code `/usr/bin/which` anywhere, so for now, why not use `/usr/bin/env` to resolve which? See https://github.com/GitCredentialManager/git-credential-manager/pull/717 for example.

(In the long run it may cause trouble to use `which` at all since it's not POSIX standard and Debian is thinking of removing it. But then you could probably just ensure another package is installed first.)

### DIFF
--- a/src/shared/TestInfrastructure/GitTestUtilities.cs
+++ b/src/shared/TestInfrastructure/GitTestUtilities.cs
@@ -21,7 +21,7 @@ namespace GitCredentialManager.Tests
             }
             else
             {
-                psi = new ProcessStartInfo("/usr/bin/which", "git");
+                psi = new ProcessStartInfo("/usr/bin/env", "which git");
             }
 
             psi.RedirectStandardOutput = true;


dscho:
Why would we go out of our way to solve the same problem as #717 in a different way? Would it not make much more sense to port `GetGitPath()` to use `TryLocateExecutable()` instead of spawning a process at all (whether that process might be running `env which` or `which` does not matter)?

echuber2:
I mean, sure, but this was the minimal edit. I didn't analyze whether TryLocateExecutable would have identical behavior in this position either.

dscho:
Uh, I didn't mean to imply that you should close this PR...

mjcheetham:
The code in question being updated relates to the testing infrastructure locating a version of Git to test our Git-interop code. On our supported Linux platforms, the `which` command is located in the expected/hardcoded location, so we've not had a problem with this before.

It's non-trivial to replace this `which` invocation with a call to the `TryLocateExecutable()` method used in product code due to the abstractions involved. What would be easier would be to just 'inline' the code, something like this:

```csharp
public static string GetGitPath()
{
    string gitExecName = PlatformUtils.IsWindows()
        ? "git.exe"
        : "git";

    char splitChar = PlatformUtils.IsWindows()
        ? ';'
        : ':';

    string pathValue = Environment.GetEnvironmentVariable("PATH");
    if (!string.IsNullOrWhiteSpace(pathValue))
    {
        string[] paths = pathValue.Split(new[] { splitChar }, StringSplitOptions.RemoveEmptyEntries);

        foreach (string basePath in paths)
        {
            string candidatePath = Path.Combine(basePath, gitExecName);
            if (File.Exists(candidatePath))
            {
                return candidatePath;
            }
        }
    }

    throw new Exception("Failed to locate Git on the PATH");
}
```

PR #759, opened 2022-06-23
docs: update readme to show osx-arm64 support

mjcheetham:
Also while I'm here, simplify how we show the "credential store" links – they're all the same.


hudsontrius (commented):
Yes I approve

ldennington (approved):
Delightful!

PR #798, opened 2022-07-20
macos arm64: Homebrew updates

### LINKED ISSUE #772: Upload osx-arm64 package to brew
Thanks for adding osx-arm64 support! I was really stoked to try it out. But unless I'm doing something wrong (which is very possible), I believe the arm64 package has not been uploaded to Homebrew yet.

This is what I tried:
- Remove current package via `brew uninstall --cask git-credential-manager-core`
- `brew update`
- `brew install --cask git-credential-manager-core`

After this `file /usr/local/share/gcm-core/git-credential-manager-core` still reports the following:
> Mach-O 64-bit executable x86_64

`brew info git-credential-manager-core` also only seems to report to x86_64 package:
```
==> Artifacts
gcm-osx-x64-2.0.779.pkg (Pkg)
```

ldennington:
Allow users to install/uninstall the new `arm64` package via Homebrew by updating the Homebrew release asset regex to separate runtime and version and to remove the PR requirement (which was tested with [these changes](https://github.com/mjcheetham/update-homebrew/pull/10) to the `update-homebrew` task).

Fixes #772

### DIFF
--- a/.github/workflows/release-homebrew.yaml
+++ b/.github/workflows/release-homebrew.yaml
@@ -8,11 +8,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Update Homebrew tap
-      uses: mjcheetham/update-homebrew@v1.1
+      uses: mjcheetham/update-homebrew@v1.2
       with:
         token: ${{ secrets.HOMEBREW_TOKEN }}
         tap: microsoft/git
         name: git-credential-manager-core
         type: cask
-        releaseAsset: gcm-osx-x64-(.*)\.pkg
-        alwaysUsePullRequest: true
+        releaseAsset: |
+          gcm-osx-x64-(.*)\.pkg
+          gcm-osx-arm64-(.*)\.pkg


.github/workflows/release-homebrew.yaml
```diff
@@ -14,5 +14,5 @@ jobs:
         tap: microsoft/git
         name: git-credential-manager-core
         type: cask
-        releaseAsset: gcm-osx-x64-(.*)\.pkg
+        releaseAsset: gcm-osx-(?<runtime>.*)-(?<version>.*)\.pkg
         alwaysUsePullRequest: true
```
mjcheetham, line 18: Remember we'll need to update the action version once https://github.com/mjcheetham/update-homebrew/pull/10 is done and a release made :wink:

src/osx/Installer.Mac/uninstall.sh
```diff
@@ -3,6 +3,15 @@
...
 
+# Ensure we use the correct Homebrew prefix for
+# x64 and arm64 binaries.
+if [ ! -z "$HOMEBREW_PREFIX" ]
+then
+	INSTALLATIONPREFIX="$HOMEBREW_PREFIX"
+else
+	INSTALLATIONPREFIX=/usr/local
+fi
+
```
mjcheetham, line 14: This assumes that if we're being run outside of Homebrew (`$HOMEBREW_PREFIX` is unset) then we must be on an Intel Mac.

Could we not instead do a test for either:

1. The current CPU architecture, or
2. Where the uninstall.sh shell script is located? It should be co-located with the GCM exectuables.

me4502: This information is also available via the `brew config` command. 

`brew config | grep HOMEBREW_PREFIX` will output `HOMEBREW_PREFIX: /opt/homebrew` on an M1 Mac

mjcheetham: My point was that this `uninstall.sh` script is also included and used for the normal `.pkg` installation route, without Homebrew at all.. so the script should also work for ARM64 Macs without Homebrew.

In this case, uninstall depends on where the `.pkg` picks to install when run on ARM64.

ldennington: Yep, I agree with this - I've got a fix in place and am planning to submit today.

mjcheetham (approved):
🎉 ✨ ❤️

Favourjacobmudiaga:
Nice

PR #804, opened 2022-07-25
Reorder Git config list process exit wait in diagnose command

### LINKED ISSUE #795: "git-credential-manager-core.exe diagnose" hangs when "git config --list" output is long
Repro:
 - In a blank folder, `git init` and then modify `.git\config` to make the file bigger than 4K
 - Run `diagnose` sub-command, it will hang at waiting for `git config --list` output.

 [ OK ] Environment
 [ OK ] File system
 [ OK ] Networking
 [ xxx ] Git

The underline issue is that GCM spawns a child process and redirects child's stdout. There is a size limit for the pipe buffer, and if the parent process does not read data out quickly, the child process will stall at writing to stdout.

A fix would be to add `Process.OutputDataReceived` to `GitDiagnostic.RunInternalAsync`.

I'd suggest reviewing all places that spawn child process for similar issue.

mjcheetham:
Do not wait for the Git process to exit until we start reading from the stdout stream, to prevent a deadlock. This is what we do in the product code too.

Fixes #795

### DIFF
--- a/src/shared/Core/Diagnostics/GitDiagnostic.cs
+++ b/src/shared/Core/Diagnostics/GitDiagnostic.cs
@@ -32,8 +32,10 @@ namespace GitCredentialManager.Diagnostics
             log.Append("Listing all Git configuration...");
             Process configProc = _git.CreateProcess("config --list --show-origin");
             configProc.Start();
-            configProc.WaitForExit();
+            // To avoid deadlocks, always read the output stream first and then wait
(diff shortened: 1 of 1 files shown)


dscho (approved):
Technically, we await Git exit, still, but only after reading the output.

mjcheetham:
> Technically, we await Git exit, still, but only after reading the output.

Sorry; clumsy wording! Reworded description (and commit on next push).

PR #813, opened 2022-07-29
release: add installer validation

ldennington:
Add basic installer validation to release pipeline for Windows,
macOS (x64 only), and Linux (Debian package only). Validation runs the
installers/any necessary setup and `git-credential-manager-core
--version`.

Test run validating these changes can be found [here](https://github.com/ldennington/git-credential-manager/actions/runs/2757819748).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -438,12 +438,80 @@ jobs:
           signed
 
 # ================================
+#           Validate
+# ================================
+  validate:
+    name: Validate installers
+    strategy:
+      matrix:
+        component:
+          - os: ubuntu-latest
+            artifact: linux-sign
+            command: git-credential-manager-core
+          - os: macos-latest
+            artifact: osx-x64-sign
+            command: git-credential-manager-core
+          - os: windows-latest
+            artifact: win-sign
+            # Even when a standalone GCM version is installed, GitHub actions
+            # runners still only recognize the version bundled with Git for
+            # Windows due to its placement on the PATH. For this reason, we use
+            # the full path to our installation to validate the Windows version.
+            command: "$PROGRAMFILES (x86)/Git Credential Manager/git-credential-manager-core.exe"
+    runs-on: ${{ matrix.component.os }}
+    needs: [ osx-sign, win-sign, linux-sign ]
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0 # Indicate full history so Nerdbank.GitVersioning works.
+
+      - uses: dotnet/nbgv@master
+        with:
+          setCommonVars: true
+
+      - name: Download artifacts
+        uses: actions/download-artifact@v3
+        with:
+          name: ${{ matrix.component.artifact }}
+
+      - name: Install Windows
+        if: contains(matrix.component.os, 'windows')
+        shell: pwsh
+        run: |
+          $exePaths = Get-ChildItem -Path ./signed/*.exe | %{$_.FullName}
+          foreach ($exePath in $exePaths)
+          {
+            Start-Process -Wait -FilePath "$exePath" -ArgumentList "/SILENT /VERYSILENT /NORESTART"
+          }
+
+      - name: Install Linux
+        if: contains(matrix.component.os, 'ubuntu')
+        run: |
+          debpath=$(find ./*.deb)
+          sudo apt install $debpath
+          "${{ matrix.component.command }}" configure
+
+      - name: Install macOS
+        if: contains(matrix.component.os, 'macos')
+        run: |
+          # Only validate x64, given arm64 agents are not available
+          pkgpath=$(find ./*.pkg)
+          sudo installer -pkg $pkgpath -target /
+
+      - name: Validate
+        shell: bash
+        run: |
+          "${{ matrix.component.command }}" --version | sed 's/+.*//' >actual
+          echo $GitBuildVersionSimple >expect
+          cmp expect actual || exit 1
+
+# ================================
 #             Publish
 # ================================
   create-github-release:
     name: Publish GitHub draft release
     runs-on: ubuntu-latest
-    needs: [ osx-sign, win-sign, linux-sign ]
+    needs: [ validate ]
     steps:      
       - name: Check out repository
         uses: actions/checkout@v3


.github/workflows/release.yml
```diff
@@ -437,13 +437,53 @@ jobs:
...
 
+# ================================
+#           Validate
+# ================================
+  validate:
+    name: Validate installers
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ ubuntu-latest, macos-latest, windows-latest ]
```
dscho, line 448: This looks great!

One suggestion: It might make sense to imitate the matrix style of [`monitor-components.yml`](https://github.com/git-for-windows/git/blob/79972f5b45e66439383630844fb843a84ccf5cce/.github/workflows/monitor-components.yml#L23) where the matrix values specify more than a single feature. That way, we could express the `needs` value in terms of a matrix variable, and we could likewise avoid downloading unneeded artifacts, again by using a matrix variable.

.github/workflows/release.yml
```diff
@@ -437,13 +437,53 @@ jobs:
...
+    needs: [ osx-sign, win-sign, linux-sign ]
+    steps:
+      - name: Download artifacts
+        uses: actions/download-artifact@v3
```
dscho, line 452: We could restrict this via `uses: name: <matrix-variable>`

.github/workflows/release.yml
```diff
@@ -437,13 +437,53 @@ jobs:
...
+
+      - name: Build & Install Windows
+        if: contains(matrix.os, 'windows')
+        shell: sh
```
dscho, line 456: Oh, that's allowed? I thought the permissible value was called `bash`...

.github/workflows/release.yml
```diff
@@ -437,13 +437,53 @@ jobs:
...
+        run: |
+          exePaths=$(find ./win-sign/signed/*.exe) &&
+          for path in $exePaths; do
+            start "" "$path" /WAIT //SILENT //NORESTART &&
```
dscho, line 460: We probably also want `//VERYSILENT`.

dscho (approved):
Excellent work!

I had a few suggestions, but would also be fine with the current iteration.

.github/workflows/release.yml
```diff
@@ -437,13 +437,53 @@ jobs:
...
+            git-credential-manager-core --version
+          done
+
+      - name: Build & Install Linux
+        if: contains(matrix.os, 'ubuntu')
+        run: |
+          debpath=$(find ./linux-sign/*.deb)
+          sudo apt install $debpath
+          git-credential-manager-core configure
+          git-credential-manager-core --version
```
derrickstolee, line 470: Since we know that GCM is not installed _a priori_ on the build machines, this demonstrates that the executable was installed on the `PATH`. Is there any way we could validate the output of `git-credential-manager-core --version` to be sure we have the right version number?

I see this as a really important validation for the Git installer validation, but might be helpful here.

derrickstolee (approved):

.github/workflows/release.yml
mjcheetham: If we could print the output that would be great indeed! I think the script task here should do this already, no?

.github/workflows/release.yml
mjcheetham: 👍

.github/workflows/release.yml
mjcheetham: +1 for the matrix/vector

mjcheetham (approved):
This looks great! Some of the comments from the others would help debugging problems and speed things up.

.github/workflows/release.yml
derrickstolee: I was thinking something like

```
git-credential-manager-core --version >actual
cat >expect <<-EOF
$GCM_VERSION_NUMBER
EOF
cmp expect actual || return 1
```

so we get an automated build failure if they are different.

.github/workflows/release.yml
ldennington: This is great!

.github/workflows/release.yml
ldennington: It does seem to be allowed, but I can update it to bash.

.github/workflows/release.yml
ldennington: To clarify - do we want `//VERYSILENT` in addition to `//SILENT` or instead of `//SILENT`?

mjcheetham: I think `//VERYSILENT` includes `//SILENT`, but not 100% sure. Docs don't make it too clear:

https://jrsoftware.org/ishelp/topic_setupcmdline.htm#SILENT

dscho: In Git for Windows, we always use `//VERYSILENT` _in addition to_ `//SILENT`, but it might be a "better safe 'n sorry" thing. I have no idea whether `//VERYSILENT` _implies_ `//SILENT`. I guess it's time to consult the source.

*clicketyclick*

Seems that [`//VERYSILENT` sets the mode to something different than `//SILENT`](https://github.com/jrsoftware/issrc/blob/cb9c3e94593724ea062bb3070bb034abdd009652/Projects/Main.pas#L2781-L2784), but it would appear as if `//VERYSILENT` indeed does imply `//SILENT` (there is only [this](https://github.com/jrsoftware/issrc/blob/cb9c3e94593724ea062bb3070bb034abdd009652/Projects/Main.pas#L2786-L2792) and [this](https://github.com/jrsoftware/issrc/blob/cb9c3e94593724ea062bb3070bb034abdd009652/Projects/Main.pas#L3069-L3070) and [this](https://github.com/jrsoftware/issrc/blob/cb9c3e94593724ea062bb3070bb034abdd009652/Projects/Main.pas#L3265) and [this](https://github.com/jrsoftware/issrc/blob/cb9c3e94593724ea062bb3070bb034abdd009652/Projects/Main.pas#L4267-L4278) part of the code, and it all looks as if the `imVerySilent` mode does more than `imSilent`, not different things).

tl;dr it does seem as if `//VERYSILENT` should be enough. To be safe, though, I would use both (I vaguely remember that we had to do this at some stage, probably due to a hence-fixed bug in InnoSetup).

ldennington: Okie doke, I'll add both!

ldennington:
I believe we are all good on this! Here is the latest successful test run: https://github.com/ldennington/git-credential-manager/actions/runs/2852519783.

.github/workflows/release.yml
```diff
@@ -437,13 +437,81 @@ jobs:
...
+          # Only validate x64, given arm64 agents are not available
+          pkgpath=$(find ./*.pkg)
+          sudo installer -pkg $pkgpath -target /
+
+      - name: Validate
+        shell: bash
+        run: |
+          "${{ matrix.component.command }}" --version | sed 's/+.*//' >actual
+          echo $GitBuildVersionSimple >expect
+          cmp expect actual || exit 1
```
derrickstolee, line 506: Very nice!

derrickstolee (approved):

Favourjacobmudiaga:
Alright

PR #825, opened 2022-08-08
uri: support ports and split queries/fragments from path

### LINKED ISSUE #608: GCM Settings does not support port numbers

All versions

* [ ] Azure DevOps
* [ ] GitHub
* [ ] GitHub Enterprise
* [X] Bitbucket
* [ ] Other - please describe

* [X] Yes

---

* [X] Not applicable
* [ ] https://dev.azure.com/`{org}`/...
* [ ] https://`{org}`@dev.azure.com/`{org}`/...
* [ ] https://`{org}`.visualstudio.com/...

* [X] Not applicable
* [ ] I only see one identity
* [ ] I checked each identity and none worked

---

**Expected behavior**

I can specify git configuration overrides using the exact url of my host server and have them picked up, e.g.

For my development Bitbucket server it runs at http://localhost:7990/bitbucket. I want to shortcut the auto-discovery process and tell GCM this URL refers to a Bitbucket instance, so I try:

    ❯ git config --global credential.http://localhost:7990/bitbucket.provider bitbucket

I would expect this stop the auto-detect process, but it doesn't

**Actual behavior**

In order to get the desired result I have to use:

    ❯ git config --global credential.http://localhost/bitbucket.provider bitbucket
    
This is because `GitCredentialManager.UriExtensions.GetGitConfigurationScopes()` ignores the port number.

To add context for a Git URL like:

    > git clone http://localhost:7990/bitbucket/scm/project_1/rep_1.git

Git will pass the following to GCM

    protocol=http
    host=localhost:7990

But the port is ignored in `Settings`

Further context Git itself correctly matches the following:

    > git config --global credential.http://localhost:7990/bitbucket.usehttppath true

**Logs**

N/A

ldennington:
GCM currently ignores ports in URIs. This means attempting to authenticate
to a URI with a port can lead to some unexpected behavior (e.g.
shortcutting the provider auto-detect process won't work, even if the
provider is set in config). This change adds support for ports by updating
GetGitConfigurationScopes() to recognize them.

This change also updates GetRemoteUri() to recognize paths with queries
and fragments, as that issue was uncovered during the implementation of
the GetGitConfigurationScopes() fix. Without it, input paths containing
queries and/or fragments get saved as part of Uri.Path, which converts the
query '?' and the fragment '#' to url encoding.

I validated these changes with unit tests for applicable scenarios and by
running a locally-compiled version of GCM with tracing enabled and the 
following config set:

`credential.http://localhost:7990/bitbucket.provider bitbucket`

Trace logs showed that the override was recognized and auto-detection
was skipped.

Fixes #608

### DIFF
--- a/src/shared/Core.Tests/InputArgumentsTests.cs
+++ b/src/shared/Core.Tests/InputArgumentsTests.cs
@@ -220,6 +220,29 @@ namespace GitCredentialManager.Tests
             Assert.Equal(expectedUri, actualUri);
         }
 
+        [Theory]
+        [InlineData("foo?query=true")]
+        [InlineData("foo#fragment")]
+        [InlineData("foo?query=true#fragment")]
+        public void InputArguments_GetRemoteUri_PathQueryFragment_ReturnsCorrectUri(string path)
+        {
+            var expectedUri = new Uri($"https://example.com/{path}");
+
+            var dict = new Dictionary<string, string>
+            {
+                ["protocol"] = "https",
+                ["host"]     = "example.com",
+                ["path"]     = path
+            };
+
+            var inputArgs = new InputArguments(dict);
+
+            Uri actualUri = inputArgs.GetRemoteUri();
+
+            Assert.NotNull(actualUri);
+            Assert.Equal(expectedUri, actualUri);
+        }
+
         [Fact]
         public void InputArguments_GetRemoteUri_IncludeUser_AuthorityPathUserInfo_ReturnsUriWithAll()
         {
--- a/src/shared/Core/InputArguments.cs
+++ b/src/shared/Core/InputArguments.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
+using System.Linq;
 
 namespace GitCredentialManager
 {
@@ -94,10 +95,7 @@ namespace GitCredentialManager
             string[] hostParts = Host.Split(':');
             if (hostParts.Length > 0)
             {
-                var ub = new UriBuilder(Protocol, hostParts[0])
-                {
-                    Path = Path
-                };
+                var ub = new UriBuilder(Protocol, hostParts[0]);
 
                 if (hostParts.Length > 1 && int.TryParse(hostParts[1], out int port))
                 {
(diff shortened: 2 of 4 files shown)


src/shared/Core/UriExtensions.cs
```diff
@@ -88,12 +88,14 @@ public static IEnumerable<string> GetGitConfigurationScopes(this Uri uri)
 
             string schemeAndDelim = $"{uri.Scheme}{Uri.SchemeDelimiter}";
             string host = uri.Host.TrimEnd('/');
+            // If port is default, don't append
+            string port = uri.Port == 80 ? "" : $":{uri.Port}";
```
mjcheetham, line 92: The default port number depends on the scheme/protocol (for `https` it's 443).

You can use the [`System.Uri::IsDefaultPort`](https://learn.microsoft.com/en-us/dotnet/api/system.uri.isdefaultport?view=net-6.0) instance property to check this easily however!

src/shared/Core.Tests/InputArgumentsTests.cs
```diff
@@ -176,6 +176,28 @@ public void InputArguments_GetRemoteUri_AuthorityAndPort_ReturnsUriWithAuthority
             Assert.Equal(expectedUri, actualUri);
         }
 
+        [Theory]
+        [InlineData("example.com", 456, "foo")]
+        [InlineData("example.com", 789, "foo/bar")]
+        [InlineData("example.foo.com", 789, "bar")]
+        public void InputArguments_GetRemoteUri_AuthorityAndPortAndPath_ReturnsUriWithAuthorityAndPortAndPath(string authority, int port, string path)
```
ldennington, line 183: @mjcheetham - just a heads up that I used `Authority` here to align with the names of the other tests in this file.

src/shared/Core/InputArguments.cs
```diff
@@ -91,13 +92,13 @@ public Uri GetRemoteUri(bool includeUser = false)
...
+
             string[] hostParts = Host.Split(':');
             if (hostParts.Length > 0)
             {
-                var ub = new UriBuilder(Protocol, hostParts[0])
-                {
-                    Path = Path
-                };
+                ub.Scheme = Protocol;
+                ub.Host = hostParts[0];
```
mjcheetham, line 101: Is there a reason we can't continue to use the [`UriBuilder(string schemeName, string hostName)`](https://learn.microsoft.com/en-us/dotnet/api/system.uribuilder.-ctor?view=net-7.0#system-uribuilder-ctor(system-string-system-string)) constructor here?

```suggestion
                var ub = new UriBuilder(Protocol, hostParts[0]);
```

src/shared/Core/InputArguments.cs
```diff
@@ -109,6 +110,28 @@ public Uri GetRemoteUri(bool includeUser = false)
...
+
+                    switch (pathParts.Length)
+                    {
+                        // If we have 3 items, that means path, query, and fragment
+                        case 3:
+                            ub.Query = pathParts[1];
+                            ub.Fragment = pathParts[2];
+                            break;
+                        // If we have 2 items, we must distinguish between query and fragment
+                        case 2 when Path.Contains('?'):
```
mjcheetham, line 127: Neat use of `switch` and `when` :)

mjcheetham (approved):
Looks good! Just one 'nit' question.

src/shared/Core/InputArguments.cs
ldennington: Nope, just a hangover from a previous iteration. Will update!

PR #839, opened 2022-08-15
install from source: use native feeds for Ubuntu 22.04

### LINKED ISSUE #841: Install from source script failing (Ubuntu 22.04)
The GCM install from source script is [failing for Ubuntu 22.04](https://github.com/GitCredentialManager/git-credential-manager/actions/runs/2861075334), due to conflicts from recent updates to deploy dotnet natively to Jammy feeds (see [this issue](https://github.com/dotnet/core/issues/7699) for details).

We should update the install from source script to install using native feeds for Ubuntu 22.04 and later and to continue to install from `packages.microsoft.com` for previous versions.

ldennington:
The dotnet team recently added support for installing natively from Jammy feeds on
Ubuntu:

https://github.com/dotnet/core/issues/7699

This caused our current installation to fail, as it caused conflicts with
the packages.microsoft.com feed we use in the install from source script
for Debian/Ubuntu.

This change modifies the Debian/Ubuntu dotnet install process to install from
Jammy feeds for users on Ubuntu 22.04 and greater while continuing to use
the packages.microsoft.com feed for Debian and older Ubuntu versions.

Here is a successful run of the validation workflow with these changes: https://github.com/ldennington/git-credential-manager/actions/runs/2884331734

Fixes #841

### DIFF
--- a/src/linux/Packaging.Linux/install-from-source.sh
+++ b/src/linux/Packaging.Linux/install-from-source.sh
@@ -1,9 +1,9 @@
 #!/bin/sh
 
-# halt execution immediately on failure
-# note there are some scenarios in which this will not exit;
-# see https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html
-# for additional details
+# Halt execution immediately on failure.
+# Note there are some scenarios in which this will not exit; see
+# https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html
+# for additional details.
 set -e
 
 is_ci=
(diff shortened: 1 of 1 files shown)


src/linux/Packaging.Linux/install-from-source.sh
```diff
@@ -116,7 +116,17 @@ fi
 eval "$(sed -n 's/^ID=/distribution=/p' /etc/os-release)"
 eval "$(sed -n 's/^VERSION_ID=/version=/p' /etc/os-release | tr -d '"')"
 case "$distribution" in
-    debian | ubuntu)
+    ubuntu)
```
mjcheetham, line 119: Don't we only want to rely on the Jammy/22.04 feeds in Ubuntu 22.04? What happens running this new block on downlevel Ubuntu versions?

ldennington: Yes, this was a good call. I spent a fair amount of time today playing around with this (particularly trying to make it work with `dpkg --compare-versions`), and I finally settled on this approach because (1) it works and (2) it's simple. However, I do recognize it may be a little _too_ simple, so if you know of a better way feel free to share.

I also added a small second commit to fix up the comments, which were bugging me 😁.

derrickstolee (approved):
This makes sense to me. Ignoring whitespace made the diff look really nice.

PR #842, opened 2022-08-17
workflows: auto-generate maintainer away issue

ldennington:
Add a new workflow to create issues that will notify GCM users and
contributors of when maintainers are away and the repo is not being
actively monitored. This workflow is controlled by a `workflow_dispatch`
trigger that requires both a start date and end date for maintainer
absence before running.

### DIFF
--- a/.github/workflows/maintainer-absence.yml
+++ b/.github/workflows/maintainer-absence.yml
@@ -1,0 +1,60 @@
+name: maintainer-absence
+
+on:
+  workflow_dispatch:
+    inputs:
+      startDate:
+        description: 'First day of maintainer absence [mm-dd-yyyy]'
+        required: true
+      endDate:
+        description:  'Last day of maintainer absence [mm-dd-yyyy]'
+        required: true
+
+permissions:
+  issues: write
+
+jobs:
+  create-issue:
+    name: create-issue
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/github-script@v6
+        with:
+          script: |
+            const startDate = new Date('${{ github.event.inputs.startDate }}');
+            const endDate = new Date('${{ github.event.inputs.endDate }}');
+
+            if (startDate > endDate) {
+              throw 'Start date cannot be later than end date.';
+            }
+
+            // Calculate total days of absence
+            const differenceInDays = endDate.getTime() - startDate.getTime();
+            const lengthOfAbsence = differenceInDays/(1000 * 3600 * 24);
+
+            // Create issue
+            issue = await github.rest.issues.create({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              // Use the briefer input date format in title (instead of JavaScript's full date string)
+              title: `Maintainer(s) will be away from ${{ github.event.inputs.startDate }} until ${{ github.event.inputs.endDate }}`,
+              body: `The ${context.repo.repo} maintainer(s) will be away for ${lengthOfAbsence} day${lengthOfAbsence > 1 ? 's' : ''} beginning on
+              ${startDate.toDateString()} and ending on ${endDate.toDateString()}. During this time, the maintainer(s)
+              will not be actively monitoring PRs, discussions, etc. Please report any issues
+              requiring immediate attention to [@GitCredManager](https://twitter.com/GitCredManager) on Twitter.`
+            });
+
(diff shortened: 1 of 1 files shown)


derrickstolee (commented):
Issues could be created using this template by anyone, right? Should we be inviting that kind of activity, or can maintainers create their own issue on the rare occasions that it happens?

ldennington:
> Issues could be created using this template by anyone, right? Should we be inviting that kind of activity, or can maintainers create their own issue on the rare occasions that it happens?

@mjcheetham and I spoke about this in triage this morning. We decided it would be better to automate creation of this type of issue with a workflow that has a workflow dispatch control only maintainers can use. Stay tuned!

ldennington:
@derrickstolee @mjcheetham PR has been updated with the new workflow. You can find a sample issue [here](https://github.com/ldennington/git-credential-manager/issues/21). In case you're not allowed to view issues in my fork, here's a screenshot as well:

Screen Shot 2022-08-19 at 12 30 48 PM

Any suggestions for improvement are welcome.

.github/workflows/maintainer-absence.yml
```diff
@@ -0,0 +1,44 @@
...
+
+on:
+  workflow_dispatch:
+    inputs:
+      startDate:
+        description: 'First day of maintainer absence [mm-dd-yyyy]'
+        required: true
+      endDate:
+        description:  'Last day of maintainer absence [mm-dd-yyyy]'
+        required: true
```
derrickstolee, line 11: You might get some complaints from @mjcheetham and other Europeans about this date format.

I tend to think that `yyyy-mm-dd` is pretty clear from context, but I'm willing to be wrong here. But of course, what matters is the parsing into a datetime in Javascript, which might not work right with that format. This is a completely optional thing to consider.

derrickstolee (approved):

derrickstolee (approved):

.github/workflows/maintainer-absence.yml
dscho: I guess you could also use something like "6 Oct 2022" and Javascript groks it.

The suggestion in the label is good, because it raises awareness that specifying the date in dd-mm-yy format will either fail or worse: succeed but use the wrong date.

.github/workflows/maintainer-absence.yml
```diff
@@ -0,0 +1,44 @@
...
+            github.rest.issues.create({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              // Use the briefer input date format in title (instead of JavaScript's full date string)
+              title: `Maintainer(s) will be away from ${{ github.event.inputs.startDate }} until ${{ github.event.inputs.endDate }}`,
+              body: `The ${context.repo.repo} maintainer(s) will be away for ${lengthOfAbsence} day${lengthOfAbsence > 1 ? 's' : ''} beginning on
+              ${startDate.toDateString()} and ending on ${endDate.toDateString()}. During this time, the maintainer(s)
+              will not be actively monitoring PRs, discussions, etc. Please report any issues
+              requiring immediate attention to [@GitCredManager](https://twitter.com/GitCredManager) on Twitter.`
+            });
```
dscho, line 44: Could we also pin the issue automatically?

Sadly, there does not seem to be any REST API for that, but here's some GraphQL demonstrating how to do that: https://github.com/helaili/github-graphql-action/blob/master/samples/pinissue.query.yaml

dscho (approved):

.github/workflows/maintainer-absence.yml
dscho: Here's how to use GraphQL in `github-script`: https://github.com/actions/github-script/#run-custom-graphql-queries

ldennington: Nice find! I was disappointed by the lack of REST support for pinning, but this looks promising!!

ldennington: Got it working! Here's a successful run with the addition of pinning: https://github.com/ldennington/git-credential-manager/actions/runs/2899610439.

.github/workflows/maintainer-absence.yml
ldennington: Let's leave as-is for now. If we get complaints, we can re-visit.

dscho (approved):
This is awesome!

Favourjacobmudiaga:
Good

PR #821, opened 2022-08-06
Package as .NET tool

hickford:
First step for #820

https://docs.microsoft.com/en-us/dotnet/core/tools/global-tools
https://docs.microsoft.com/en-us/dotnet/core/tools/global-tools-how-to-create

```
dotnet pack --output out/nupkg
dotnet tool install --global --add-source ./out/nupkg  git-credential-manager-core --version 2.0.798-g57a5791f02
```

This works nicely

```
 which git-credential-manager-core
$HOME/.dotnet/tools/git-credential-manager-core
> git-credential-manager-core diagnose
Running diagnostics...

 [ OK ] Environment
 [ OK ] File system
 [ OK ] Networking
 [ OK ] Git
 [ OK ] Credential storage
 [ OK ] Microsoft authentication (AAD/MSA)
 [ OK ] GitHub API

Diagnostic summary: 7 passed, 0 skipped, 0 failed.
```

PR #844, opened 2022-08-21
Package GitHub.UI as .NET tool

hickford:
Similar to #821

I was also able to package, install and use GitHub.UI making a similar change, although I had to set config variable `githubHelper` to help git-credential-manager-core find it, because of an assembly location complication -- both are installed to `$HOME/.dotnet/tools` but these are shims pointing to  a larger binary in a package-specific location, eg. `$HOME/.dotnet.tools/.store/git-credential-manager-core/2.0.798-g57a5791f02/git-credential-manager-core` , so the two binaries can't find each other.

```
dotnet pack --output out/nupkg src/shared/GitHub.UI.Avalonia/
dotnet tool install --global --add-source ./out/nupkg git-credential-manager-github-ui --prerelease
```

```
> which GitHub.UI
/home/matthickford/.dotnet/tools/GitHub.UI
> GitHub.UI --version
2.0.798+57a5791f02
```

The installed GitHub.UI package is also surprisingly big compared to the other package

```
172M    ./.store/github.ui
4.7M    ./.store/git-credential-manager-core
```

On investigation the install contains binary dependencies for other platforms, which seems unnecessary.

```
896K    ./github.ui/2.0.798-g57a5791f02/tools/net6.0/any/runtimes/win-x64/native/libHarfBuzzSharp.dll
8.5M    ./github.ui/2.0.798-g57a5791f02/tools/net6.0/any/runtimes/win-x64/native/libSkiaSharp.dll
9.4M    ./github.ui/2.0.798-g57a5791f02/tools/net6.0/any/runtimes/win-x64/native
9.4M    ./github.ui/2.0.798-g57a5791f02/tools/net6.0/any/runtimes/win-x64
788K    ./github.ui/2.0.798-g57a5791f02/tools/net6.0/any/runtimes/win-x86/native/libHarfBuzzSharp.dll
7.1M    ./github.ui/2.0.798-g57a5791f02/tools/net6.0/any/runtimes/win-x86/native/libSkiaSharp.dll
```

### DIFF
--- a/src/shared/Git-Credential-Manager/Git-Credential-Manager.csproj
+++ b/src/shared/Git-Credential-Manager/Git-Credential-Manager.csproj
@@ -12,6 +12,13 @@
     <IsTestProject>false</IsTestProject>
     <LangVersion>latest</LangVersion>
     <TargetLatestRuntimePatch>true</TargetLatestRuntimePatch>
+    <PackAsTool>true</PackAsTool>
+    <PackageLicenseExpression>MIT</PackageLicenseExpression>
+    <ProjectUrl>https://github.com/GitCredentialManager/git-credential-manager</ProjectUrl>
+    <Title>Git Credential Manager</Title>
+    <RepositoryUrl>https://github.com/GitCredentialManager/git-credential-manager.git</RepositoryUrl>
+    <Description>Git Credential Manager is a secure Git credential helper providing multi-factor authentication support for Azure DevOps, Azure DevOps Server, GitHub, Bitbucket, and GitLab.</Description>
+    <PackageReadmeFile>README.md</PackageReadmeFile>
   </PropertyGroup>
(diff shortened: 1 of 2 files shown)


ldennington:
This is awesome! Exciting to see that GCM was able to find the GitHub.UI exe with the help of the `githubHelper` variable. I think ideally we'd probably want to package all the exes in the same tool (and I've opened [an issue in the dotnet/sdk repo](https://github.com/dotnet/sdk/issues/27459) to ask about that), but I'm not sure whether that's possible. So, it's good to see that this option could indeed work for us.

hickford:
Actually, I think everything could be packaged together in one package https://github.com/GitCredentialManager/git-credential-manager/issues/820#issuecomment-1229175795

Closing this PR.

PR #851, opened 2022-08-26
release: access Azure Storage Account with login credentials 

ldennington:
Use `--auth-mode login` command to access ESRP Azure Storage Account instead of AZURE_STORAGE_KEY. This allows us to reuse the git-fundamentals-esrp service principal login credentials we're already using for other resources and is a more secure mode of access than using the storage key.

This update was tested with [this abbreviated release workflow](https://github.com/ldennington/git-credential-manager/runs/8045792801?check_suite_focus=true).


dscho (approved):
Makes sense. We're already logged in, and now that login also has the permissions required to download those bits. Nice simplification!

PR #859, opened 2022-09-06
Added check mark for x86 support on macOS

RohanTalip:
There are gcm-osx-x64-2.0.785.* assets in https://github.com/GitCredentialManager/git-credential-manager/releases/tag/v2.0.785 and the gcm-osx-x64-2.0.785.pkg installed fine and worked for me on my Intel MacBook Pro (`uname -m` shows `x86_64`).


mjcheetham (changes requested):
Hello! Thanks for the contribution. Unfortunately, this change is not correct.

> uname -m shows x86_64

`x86_64` is one of the names for the 64-bit Intel Architecture (also known as `AMD64` or `x64`).

The `x86` on it's own in this table refers to 32-bit _only_ Intel/AMD CPUs (also sometimes called `i686`). Your Intel MacBook is still a 64-bit Intel CPU (`x86_64`).

PR #862, opened 2022-09-09
Unify line length across docs

wolf99:
I have seen that the docs do not follow a uniform style of line breaking / line length.

Previously @ldennington has indicated a preference for a <80 char line limit (https://github.com/GitCredentialManager/git-credential-manager/pull/668#issuecomment-1101594445) .

Thus this PR unifies the docs to follow an 80 char line limit. I have been fairly strict and also removed any breaks at end of sentences as that mixes the line breaking styles.

I have enabled the line length check performed by the markdown linter in the CI that was previously disabled since the linter was introduced in #668 .


ldennington (approved):
Looks great! Thanks for your work to standardize this across our docs! 🚀

PR #886, opened 2022-09-30
release: package gcm as dotnet tool

### LINKED ISSUE #820: dotnet tool install git-credential-manager
There are several [Linux distributions that support .NET](https://docs.microsoft.com/en-us/dotnet/core/install/linux) but can’t install deb packages. Installing or updating Git Credential Manager on these platforms is discouragingly awkward -- you have to download the binaries, decompress and copy to the path.  Even on Debian and Ubuntu which can install deb packages, updating GCM is a manual process. GCM is [particularly difficult to install on linux-arm64](https://github.com/GitCredentialManager/git-credential-manager/issues/466). 

It would thus be handy to additionally distribute Git Credential Manager as a [dotnet global tool](https://docs.microsoft.com/en-us/dotnet/core/tools/global-tools ) that could be installed `dotnet tool install -g git-credential-manager`. For users with .NET installed, this is a super convenient way to install. Updating is as simple as `dotnet tool update --global git-credential-manager`. An advantage on all platforms is that the framework-dependent install size (4.5 MB) is much much smaller than self-contained binaries (65 MB).

For applications developed in Python, Go or Nodejs, distributing via language package manager is a popular option offered among alternatives. For example, http client httpie [can be installed](https://httpie.io/docs/cli#installation) from PyPI `pip install httpie` as well as Linux distribution package managers or Snap.

ldennington:
Package GCM as a dotnet tool and publish it to NuGet.org in release pipeline.

**Note 1:** The pipeline is currently set up to publish to my fork of GCM using a GitHub Packages Registry. Here are the instructions to dogfood from this registry:

1. If you have GCM installed, uninstall it (you can find instructions for your distro/install method [here](https://github.com/GitCredentialManager/git-credential-manager#linux)).
2. If you have not already installed dotnet, [do so](https://learn.microsoft.com/en-us/dotnet/core/install/linux).
4. Run the following:
```
export PATH="$PATH:~/.dotnet/tools"
dotnet tool install --global GitCredentialManager.Cli --add-source https://apiint.nugettest.org/v3/index.json
```
5. If you have not already configured a credential store, follow [these instructions](https://github.com/GitCredentialManager/git-credential-manager/blob/main/docs/credstores.md#credential-stores) to do so.

**Note 2:** We use a custom NuSpec file because, unfortunately, using `Content` in the `.csproj` to copy files would not overwrite the `DotnetToolSettings.xml` file. See https://github.com/NuGet/Home/issues/9902 for details.

Fixes #820

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -600,3 +600,29 @@ jobs:
               uploadDirectoryToRelease('linux-sign'),
               uploadDirectoryToRelease('linux-build/tar')
             ]);
+
+  create-dotnet-tool:
+    name: Publish dotnet tool
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0 # Indicate full history so Nerdbank.GitVersioning works.
+
+    - name: Setup .NET
+      uses: actions/setup-dotnet@v2
+      with:
+        dotnet-version: 6.0.201
+
+    - uses: dotnet/nbgv@master
+      with:
+        setCommonVars: true
+
+    - name: Package tool
+      run: |
+        src/shared/DotnetTool/pack-tool.sh
+
+    - name: Publish tool
+      run: |
+        dotnet nuget push ./out/nupkg/*.nupkg \
+          --api-key ${{ secrets.NUGET_API_KEY }} --source https://api.nuget.org/v3/index.json
--- a/src/shared/Core/Authentication/AuthenticationBase.cs
+++ b/src/shared/Core/Authentication/AuthenticationBase.cs
@@ -110,58 +110,111 @@ namespace GitCredentialManager.Authentication
             }
         }
 
-        protected bool TryFindHelperExecutablePath(string envar, string configName, string defaultValue, out string path)
+        protected bool TryFindHelperCommand(string envar, string configName, string defaultValue, out string command, out string args)
         {
-            bool isOverride = false;
+            command = null;
+            args = null;
+
+            //
+            // Search for UI helpers with the following precedence and logic..
+            //
+            // 1. (unset): use the default helper name that's in the source code and go to #3
+            // 2. <absolute>: use the absolute path only and exactly as entered
+            // 3. <relative>: search for..
+            //     a. <appdir>/<relative>(.exe) - run this directly
+            //     b. <appdir>/<relative>(.dll) - use `dotnet exec` to run
+            //     c. Search PATH for <relative>(.exe) - run this directly
+            //        NOTE: do NOT search PATH for <relative>(.dll) as we don't know if this is a dotnet executable..
+            //
+            // We print warning messages for missing helpers specified by the user, not the in-box ones.
+            //
             if (Context.Settings.TryGetPathSetting(
-                envar, Constants.GitConfiguration.Credential.SectionName, configName, out string helperName))
+                   envar, Constants.GitConfiguration.Credential.SectionName, configName, out string helperName))
             {
-                Context.Trace.WriteLine($"UI helper override specified: '{helperName}'.");
-                isOverride = true;
-            }
-            else
-            {
-                // Use the default helper if none was specified.
-                // On Windows append ".exe" for the default helpers only. If a user has specified their own
-                // helper they should append the correct extension.
-                helperName = PlatformUtils.IsWindows() ? $"{defaultValue}.exe" : defaultValue;
-            }
-
-            // If the user set the helper override to the empty string then they are signalling not to use a helper
-            if (string.IsNullOrEmpty(helperName))
-            {
-                path = null;
-                return false;
-            }
-
-            if (Path.IsPathRooted(helperName))
-            {
-                path = helperName;
-            }
-            else
-            {
-                string executableDirectory = Path.GetDirectoryName(Context.ApplicationPath);
-                path = Path.Combine(executableDirectory!, helperName);
-            }
-
-            if (!Context.FileSystem.FileExists(path))
-            {
-                // Only warn for missing helpers specified by the user, not in-box ones
-                if (isOverride)
+                // If the user set the helper override to the empty string then they are signalling not to use a helper
+                if (string.IsNullOrEmpty(helperName))
                 {
-                    Context.Trace.WriteLine($"UI helper '{helperName}' was not found at '{path}'.");
-                    Context.Streams.Error.WriteLine($"warning: could not find configured UI helper '{helperName}'");
+                    Context.Trace.WriteLine("UI helper override specified as the empty string.");
+                    return false;
                 }
 
+                Context.Trace.WriteLine($"UI helper override specified: '{helperName}'.");
+            }
+            else
+            {
+                Context.Trace.WriteLine($"Using default UI helper: '{defaultValue}'.");
+                helperName = defaultValue;
+            }
+
+            //
+            // Check for an absolute path.. run this directly without intermediaries or modification
+            //
+            if (Path.IsPathRooted(helperName))
+            {
+                if (Context.FileSystem.FileExists(helperName))
+                {
+                    Context.Trace.WriteLine($"UI helper found at '{helperName}'.");
+                    command = helperName;
+                    return true;
+                }
+
+                Context.Trace.WriteLine($"UI helper was not found at '{helperName}'.");
+                Context.Streams.Error.WriteLine($"warning: could not find configured UI helper '{helperName}'");
                 return false;
             }
 
-            return true;
+            //
+            // Search the installation directory for an in-box helper
+            //
+            string appDir = Path.GetDirectoryName(Context.ApplicationPath);
+            string inBoxExePath = Path.Combine(appDir, PlatformUtils.IsWindows() ? $"{helperName}.exe" : helperName);
+            string inBoxDllPath = Path.Combine(appDir, $"{helperName}.dll");
+
+            // Look for in-box native executables
+            if (Context.FileSystem.FileExists(inBoxExePath))
+            {
+                Context.Trace.WriteLine($"Found in-box native UI helper: '{inBoxExePath}'");
+                command = inBoxExePath;
+                return true;
+            }
+
+            // Look for in-box .NET framework-dependent executables
+            if (Context.FileSystem.FileExists(inBoxDllPath))
+            {
+                string dotnetName = PlatformUtils.IsWindows() ? "dotnet.exe" : "dotnet";
+                if (!Context.Environment.TryLocateExecutable(dotnetName, out string dotnetPath))
+                {
+                    Context.Trace.WriteLine($"Unable to run UI helper '{inBoxDllPath}' without the .NET CLI.");
+                    Context.Streams.Error.WriteLine($"warning: could not find .NET CLI to run UI helper '{inBoxDllPath}'");
+                    return false;
+                }
+
+                Context.Trace.WriteLine($"Found in-box framework-dependent UI helper: '{inBoxDllPath}'");
+                command = dotnetPath;
+                args = $"exec {QuoteCmdArg(inBoxDllPath)} ";
+                return true;
+            }
+
+            //
+            // Search the PATH for a native executable (do NOT search for out-of-box .NET framework-dependent DLLs)
+            //
+            if (Context.Environment.TryLocateExecutable(helperName, out command))
+            {
+                Context.Trace.WriteLine($"Found UI helper on PATH: '{helperName}'");
+                return true;
+            }
+
+            //
+            // No helper found!
+            //
+            Context.Trace.WriteLine($"UI helper '{helperName}' was not found.");
+            Context.Streams.Error.WriteLine($"warning: could not find UI helper '{helperName}'");
+            return false;
         }
 
         public static string QuoteCmdArg(string str)
         {
-            char[] needsQuoteChars = {'"', ' ', '\\', '\n', '\r', '\t'};
+            char[] needsQuoteChars = { '"', ' ', '\\', '\n', '\r', '\t' };
             bool needsQuotes = str.Any(x => needsQuoteChars.Contains(x));
 
             if (!needsQuotes)
--- a/src/shared/Atlassian.Bitbucket.Tests/BitbucketAuthenticationTest.cs
+++ b/src/shared/Atlassian.Bitbucket.Tests/BitbucketAuthenticationTest.cs
@@ -113,7 +113,8 @@ namespace Atlassian.Bitbucket.Tests
         {
             var targetUri = new Uri("https://bitbucket.org");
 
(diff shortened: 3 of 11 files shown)


derrickstolee (commented):
This looks like a nice and clean result. I'm sure the journey to get here wasn't easy, but thanks for putting in that effort.

src/linux/Packaging.Linux/dotnet-tool.nuspec
```diff
@@ -0,0 +1,30 @@
+
+<package xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd">
+    <metadata>
+        <id>git-credential-manager</id>
+        <!-- A "dummy" version is required, even though we override in csproj -->
+        <version>0.0.0</version>
+        <description>Secure, cross-platform Git credential storage with authentication to Azure Repos, GitHub, and other popular Git hosting services.</description>
+        <authors>The Git Fundamentals Team</authors>
```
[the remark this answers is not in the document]
ldennington: I'm not sure. This is unfortunately a required field, or else I would have removed it. Here's what the docs recommend:

authors
A comma-separated list of packages authors, matching the profile names on nuget.org. These are displayed in the NuGet Gallery on nuget.org and are used to cross-reference packages by the same authors.

Let me take a look at some other packages and think about what we want to do about this.

.github/workflows/release.yml
```diff
@@ -537,3 +538,39 @@ jobs:
...
             ]);
+
+  create-dotnet-tool:
+    name: Publish dotnet tool for Linux
+    runs-on: ubuntu-latest
+    needs: linux-build
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0 # Indicate full history so Nerdbank.GitVersioning works.
```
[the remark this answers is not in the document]
ldennington: You are correct - we need the `.csproj`, `.nuspec`, and a new script that will be coming with my next change.

hickford:
Exciting! Is there a .nupkg file to try out?

hickford:
> Package GCM for Linux as a dotnet tool

The package is cross platform, right?

ldennington:
> Exciting! Is there a .nupkg file to try out?

The best way to dogfood at the moment is to build a dummy version of the package and install it locally (it's only published to my fork's GitHub Packages registry at the moment which has some...issues). For the local install, from the root of your gcm repo run:

```
git fetch <your name for my fork>
git checkout dotnet-tool
dotnet build --configuration=LinuxRelease -o payload
./src/linux/Packaging.Linux/pack-tool.sh --version=1.2.3 --release=./out/linux/Packaging.Linux/payload/Release
dotnet tool install --global --add-source ./src/linux/Packaging.Linux/out/nupkg GitCredentialManager.Cli --prerelease
```

I will keep you posted with new instructions if we're able to resolve the GitHub registry issues.

ldennington:
Going to go ahead and promote this to a full PR - I feel good about our process for dogfooding from the registry while we wait to publish to NuGet.

mjcheetham:
> > Package GCM for Linux as a dotnet tool
> 
> The package itself is cross platform, right?

This is not a cross platform package. It contains compiled/self-contained binaries specifically for Linux (the `linux-x64` RID).

To make this a cross platform package we'd need to `dotnet publish` without `--self-contained` or any `-r <RID>` options, and also modify the code that locates and launches the UI helper executables to also try `dotnet exec <helper>.dll` if the self-contained executable version is not present.

However, the goal here with this change is to help Linux users specifically. For Windows and macOS we already have robust installation methods (via Git for Windows and Homebrew respectively), so the cross platform idea would be a future idea that we'd welcome community contribution for.

src/linux/Packaging.Linux/dotnet-tool.nuspec
```diff
@@ -0,0 +1,15 @@
+<package xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd">
+    <metadata>
+        <id>GitCredentialManager.Cli</id>
+        <version>$version$</version>
+        <description>Secure, cross-platform Git credential storage with authentication to Azure Repos, GitHub, and other popular Git hosting services.</description>
+        <authors>The Git Fundamentals Team</authors>
```
mjcheetham, line 6: I would change this to `GitCredentialManager`.. the project exists separately from GitHub, and would then match the project org name.

src/linux/Packaging.Linux/pack-tool.sh
```diff
@@ -0,0 +1,70 @@
...
+esac
+done
+
+if [ -z "$VERSION" ]; then
+    VERSION="$GitBuildVersionSimple"
+fi
+
+THISDIR="$( cd "$(dirname "$0")" ; pwd -P )"
+ROOT="$( cd "$THISDIR"/../../.. ; pwd -P )"
+echo "ROOT is $ROOT"
```
mjcheetham, line 33: debugging?

.github/workflows/release.yml
```diff
@@ -377,14 +377,23 @@ jobs:
...
+    - uses: dotnet/nbgv@master
+      with:
+        setCommonVars: true
+
     - name: Build
-      run: dotnet build --configuration=LinuxRelease
+      run: |
+        dotnet build --configuration=LinuxRelease -o payload
+        ./src/linux/Packaging.Linux/pack-tar.sh --version=$GitBuildVersionSimple --configuration=Release
+        ./src/linux/Packaging.Linux/pack-deb.sh --version=$GitBuildVersionSimple --configuration=Release
```
mjcheetham, line 388: Just to clarify, on a local dev build we are no longer packing the Debian package or tarball (since we've moved them to scripts that are not in the MSBuild call chain; the `pack-X.sh` scripts must be called manually)?

src/linux/Packaging.Linux/pack-tool.sh
```diff
@@ -0,0 +1,70 @@
...
+    shift # past argument=value
+    ;;
+    --version=*)
+    VERSION="${i#*=}"
+    shift # past argument=value
+    ;;
+    --release=*)
+    RELEASE="${i#*=}"
+    shift # past argument=value
+    ;;
```
mjcheetham, line 20: What is this argument used for? I don't see a place where we call the script with this.

mjcheetham (approved):
This looks and works great (tested on one of my WSL instances)!

I had a few comments and questions around package metadata and which scripts are called on a dev box.

src/linux/Packaging.Linux/dotnet-tool.nuspec
mjcheetham: ..or rather "Git Credential Manager" with spaces :)

ldennington: Sounds good. After re-reviewing the docs, I don't think `Company` is a required field either, so I'll go ahead and remove that.

src/linux/Packaging.Linux/pack-tool.sh
ldennington: Oh drat! 🐛

.github/workflows/release.yml
ldennington: This is true. Probably need to consolidate them for local development - I can make that update.

src/linux/Packaging.Linux/pack-tool.sh
ldennington: It defaults to the location we expect when running in GH Actions (definitely the happy path), but if you run it locally you need to specify this directory ([here's an example](https://github.com/GitCredentialManager/git-credential-manager/pull/886#issuecomment-1267698180) from my local install instructions).

hickford:
> Here are the instructions to dogfood from this registry

Installed successfully. Tested git push to GitHub.

Package size is 147 MB (too big to upload here). Install size is 493 MB.

ldennington:
> > Package size is 147 MB, install size is 493 MB.
> 
> For me, that's prohibitively big. **I want to use GCM everywhere I use Git** including where disk space is precious. Git is 35 MB. The .NET runtime is 70 MB. Have you considered omitting the UI helpers? Advantages:
> 
> 1. Small install size (5 MB)
> 2. Cross platform (in particular supporting linux-arm64, on which otherwise [difficult to install GCM](https://github.com/GitCredentialManager/git-credential-manager/issues/466))

@mjcheetham and I talked this over. The best path forward is probably to go ahead and make this cross-plat and update the way we call the helper exes. Hopefully no longer bundling the runtime multiple times will bring this down to an acceptable size. Planning to get as far as I can on this today.

src/linux/Packaging.Linux/pack-tool.sh
ldennington: Ended up removing after all!

.github/workflows/release.yml
ldennington: Ended up moving this into its own directory beneath `shared` since it's now cross-platform - just lmk if that needs to change!

ldennington:
@mjcheetham @hickford - I believe this is now in a state where it's prepared for final review.

src/shared/Core/Authentication/AuthenticationBase.cs
```diff
@@ -110,28 +110,44 @@ protected void ThrowIfTerminalPromptsDisabled()
...
 
-        protected bool TryFindHelperExecutablePath(string envar, string configName, string defaultValue, out string path)
+        protected bool TryFindHelperCommand(string envar, string configName, string defaultValue, out string command, out string args)
+        {
+            if (TryFindHelperPath(envar, configName, defaultValue, out string path))
+            {
+                if (path.EndsWith("dll"))
+                {
+                    command = $"dotnet";
+                    args = $" exec {path} ";
```
mjcheetham, line 120: For added safety, end the args with `-- ` to ensure that no args that callers may add can be confused as `dotnet` args.

`--` is used to signify the end of options; the rest are always passed straight to the child process.

https://learn.microsoft.com/en-us/dotnet/standard/commandline/syntax#the----token

```suggestion
                    args = $" exec {path} -- ";
```

src/shared/Core/Authentication/AuthenticationBase.cs
```diff
@@ -140,8 +156,11 @@ protected bool TryFindHelperExecutablePath(string envar, string configName, stri
             }
             else
             {
-                string executableDirectory = Path.GetDirectoryName(Context.ApplicationPath);
-                path = Path.Combine(executableDirectory!, helperName);
+                string applicationDirectory = Path.GetDirectoryName(Context.ApplicationPath);
+                string exePath = Path.Combine(applicationDirectory, PlatformUtils.IsWindows() ? $"{defaultValue}.exe" : defaultValue);
+                string dllPath = Path.Combine(applicationDirectory, $"{defaultValue}.dll");
```
mjcheetham, line 161: Should this be using `helperName` and not `defaultValue`? How else can a user specify a custom helper that isn't rooted?

I guess what do we want to allow here is the real question? Today if the user specifies `"foo"` as a custom helper name, we'd look for `"<gcmroot>/foo(.exe)"`.

There's probably questions around if this is correct or not.. but the change here is a change in behaviour.

Perhaps we should allow something like the following:

- _(unset)_: use the default helper name that's in the source code
- `<absolute>`: use the absolute path only and exactly as entered
- `<relative>`: search for..
  - `<appdir>/<relative>(.exe)` - run this directly
  - `<appdir>/<relative>(.dll)` - use `dotnet exec` to run
  - Search `$PATH` for `<relative>(.exe) - run this directly
  - Do NOT search `$PATH` for `<relative>(.dll) as we don't know if this is a `dotnet` executable..

Aside: should we also ensure we can locate the `dotnet` command directly on the `$PATH`?

src/shared/Core/Authentication/AuthenticationBase.cs
```diff
@@ -110,28 +110,44 @@ protected void ThrowIfTerminalPromptsDisabled()
             }
         }
 
-        protected bool TryFindHelperExecutablePath(string envar, string configName, string defaultValue, out string path)
+        protected bool TryFindHelperCommand(string envar, string configName, string defaultValue, out string command, out string args)
+        {
+            if (TryFindHelperPath(envar, configName, defaultValue, out string path))
+            {
+                if (path.EndsWith("dll"))
```
mjcheetham, line 117:
```suggestion
                if (path.EndsWith(".dll"))
```

..else we'd match `"my-native-programdll"` on UNIX platforms as a `dotnet` program.

mjcheetham (changes requested):
Overall this looks and works great from my testing! 🎉 

I've left some comments about how we locate and execute the `.dll` framework-dependent UI helpers.

src/shared/Core/Authentication/AuthenticationBase.cs
mjcheetham: ignore me.. this isn't true for `dotnet exec` it seems!

src/shared/Core/Authentication/AuthenticationBase.cs
```diff
@@ -110,58 +110,110 @@ protected void ThrowIfTerminalPromptsDisabled()
...
+                return true;
             }
 
-            if (!Context.FileSystem.FileExists(path))
+            // Look for in-box .NET framework-dependent executables
+            if (Context.FileSystem.FileExists(inBoxDllPath))
             {
-                // Only warn for missing helpers specified by the user, not in-box ones
-                if (isOverride)
+                if (!Context.Environment.TryLocateExecutable("dotnet", out string dotnetPath))
```
mjcheetham, line 184: This is my bad... 😢 

```suggestion
                string dotnetName = PlatformUtils.IsWindows() ? "dotnet.exe" : "dotnet";
                if (!Context.Environment.TryLocateExecutable(dotnetName, out string dotnetPath))
```

mjcheetham (approved):
🎉 This is great! Sorry for the bad patches I sent you off-PR.. 😞

hickford:
> @hickford: A new package has been pushed to my registry (`GitCredentialManager.Cli` version `2.0.807`) that is cross-platform compatible and has a much smaller `.nupkg` size of 53.1 MB. I'm curious, how did you measure the install size? I'd like to compare that as well.

@ldennington Opened https://github.com/git-ecosystem/git-credential-manager/issues/1212 with measurement details

Sandman1225737:
Let's just say some people may know alot more about SHIBA an the November
2021 crypto crash an maybe some other things that no one else does an if I
tried my hardest every day of my life I will never be poor.

0N$W¥9€^1225737

mohamadfaqani:
سلام

Sent from Yahoo Mail on Android 
 
  On Fri, Apr 21, 2023 at 1:08, CHARLES EDWARD FLEMING ***@***.***> wrote:   

Let's just say some people may know alot more about SHIBA an the November
2021 crypto crash an maybe some other things that no one else does an if I
tried my hardest every day of my life I will never be poor.

0N$W¥9€^1225737

PR #892, opened 2022-10-07
docs: use short link for Linux install src

mjcheetham:
Use a short link for the Linux install from source script.


ldennington (approved):
Much cleaner, thanks!

PR #893, opened 2022-10-08
Updates RFC links to authoritative URL

wolf99:
The current URLs use an outdated source.

The rfc-editor.org site is stated to be the authoritative source, see paragraph 3 on https://www.ietf.org/standards/rfcs/.

The redirection used at the existing URL cause the later versions of the link checker to fail for the instances in the documentation and so part of the reason that #840 needed to be closed without merging, see: https://github.com/GitCredentialManager/git-credential-manager/actions/runs/2863387781/jobs/4540421391. 

However, instances in the source code comments have also been updated.


mjcheetham (approved):
Thanks for keeping up to date with these! 👍

PR #896, opened 2022-10-09
Store GitLab token expiry date

hickford:
No longer necessary to query token info API. This removes a round trip.


hickford:
Better would be to use Git 2.40's new [password_expiry_utc attribute](https://git-scm.com/docs/git-credential#Documentation/git-credential.txt-codepasswordexpiryutccode). This would require plumbing through GCM and into backing stores.

PR #901, opened 2022-10-13
dotnet tool: rename package

ldennington:
Change the name of the dotnet tool package from GitCredentialManager.Cli to git-credential-manager. The reason for this change is that we were able to transfer ownership of this package (which was previously owned by JetBrains) to the git-credential-manager organization on NuGet.org. There should not be any customer impact related to this change, as the GitCredentialManager.Cli package had yet to be published.

### DIFF
--- a/src/shared/DotnetTool/dotnet-tool.nuspec
+++ b/src/shared/DotnetTool/dotnet-tool.nuspec
@@ -1,10 +1,10 @@
 
 <package xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd">
     <metadata>
-        <id>GitCredentialManager.Cli</id>
+        <id>git-credential-manager</id>
         <version>$version$</version>
         <description>Secure, cross-platform Git credential storage with authentication to Azure Repos, GitHub, and other popular Git hosting services.</description>
-        <authors>Git Credential Manager</authors>
+        <authors>git-credential-manager</authors>
         <packageTypes>
             <packageType name="DotnetTool" />
         </packageTypes>


src/shared/DotnetTool/dotnet-tool.nuspec
```diff
@@ -1,7 +1,7 @@
 
 <package xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd">
     <metadata>
-        <id>GitCredentialManager.Cli</id>
+        <id>git-credential-manager</id>
         <version>$version$</version>
         <description>Secure, cross-platform Git credential storage with authentication to Azure Repos, GitHub, and other popular Git hosting services.</description>
         <authors>Git Credential Manager</authors>
```
mjcheetham, line 7: Does the authors (or owners?) field need to match the NuGet.org org name?

mjcheetham (approved):

src/shared/DotnetTool/dotnet-tool.nuspec
hickford: I don't believe so

hickford (approved):

src/shared/DotnetTool/dotnet-tool.nuspec
ldennington: I don't think it needs to, but I think it's a good idea. Apparently [owners](https://learn.microsoft.com/en-us/nuget/reference/nuspec#optional-metadata-elements) is deprecated in favor of authors.

derrickstolee (approved):

mjcheetham (approved):

hickford:
As a dotnet tool user, it's confusing when the package name `git-credential-manager` doesn't match the command name `git-credential-manager-core`, but I still think this is the best choice. https://github.com/GitCredentialManager/git-credential-manager/pull/551/ will improve this.

PR #905, opened 2022-10-14
Add icon to dotnet tool NuGet package

mjcheetham:
Add the GCM icon to the NuGet package used for the dotnet tool.

The image is copied to the root of the package under a new `images` directory (/images). The tool payload continues to be copied under the required `/tool/net6.0/any` directory.

Modify the pack-tool.sh script to keep all build output under the /out directory of the repository (which is in .gitignore). Output from running pack-tool.sh now looks like this:

```text
.
└── out
    └── shared
        └── DotnetTool
            └── nupkg
                └── Debug
                    ├── git-credential-manager.$VERSION.nupkg
                    ├── images       # icon
                    ├── payload      # dotnet publish output
                    └── payload.sym  # pdbs
```

The contents of the NuGet package looks like this:

```text
.
├── [Content_Types].xml
├── _rels
├── git-credential-manager.nuspec
├── images
│   └── icon.png          # copied from the output images/ dir
├── package
│   └── services
│       └── metadata/...
└── tools
    └── net6.0
        └── any/...       # copied from the output payload/ dir
```


src/shared/DotnetTool/DotnetTool.csproj
```diff
@@ -9,6 +9,7 @@
       publishDir=$(PublishDir);
     </NuspecProperties>
     <NoBuild>true</NoBuild>
-    <OutputPath>../../../out/nupkg</OutputPath>
+    <OutputPath>$(ProjectOutPath)nupkg\$(Configuration)\</OutputPath>
+    <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
```
mjcheetham, line 13: This just prevents MSBuild from creating the empty/not used `net6.0` directory under `$(OutputPath)/`.

src/shared/DotnetTool/dotnet-tool.nuspec
```diff
@@ -5,11 +5,14 @@
         <version>$version$</version>
         <description>Secure, cross-platform Git credential storage with authentication to Azure Repos, GitHub, and other popular Git hosting services.</description>
         <authors>git-credential-manager</authors>
+        <icon>images\icon.png</icon>
+        <iconUrl>https://raw.githubusercontent.com/GitCredentialManager/git-credential-manager/main/assets/gcm-transparent.png</iconUrl>
```
mjcheetham, line 9: `iconUrl` is deprecated, but `icon` is only supported by newer versions of NuGet and VS.. so include both, with `iconUrl` as a fallback.

Newer tools will only use `icon`.

src/shared/DotnetTool/dotnet-tool.nuspec
```diff
@@ -5,11 +5,14 @@
...
         <packageTypes>
             <packageType name="DotnetTool" />
         </packageTypes>
   </metadata>
   <files>
-    <file src="$publishdir$" target="tools/net6.0/any" />
+    <file src="$publishdir$payload/" target="tools/net6.0/any" />
+    <file src="$publishdir$images/icon.png" target="images" />
```
mjcheetham, line 16: `$publishdir$` includes the trailing slash.. so this looks weird but is correct 😁

src/shared/DotnetTool/pack-tool.sh
```diff
@@ -38,7 +37,7 @@ esac
 done
 
 if [ -z "$VERSION" ]; then
-    VERSION="$GitBuildVersionSimple"
```
mjcheetham, line 41: `$GitBuildVersionSimple` isn't set in dev builds, so let's just require it and have the CI pass this in. If `$VERSION` ends up being empty the dev builds ended up half-baked.

src/shared/DotnetTool/pack-tool.sh
```diff
@@ -53,36 +52,25 @@ GITLAB_UI_SRC="$SRC/shared/GitLab.UI.Avalonia"
 DOTNET_TOOL="shared/DotnetTool"
 PROJ_OUT="$OUT/$DOTNET_TOOL"
 
-PACKAGE="$ROOT/nuget"
-CONFIGURATION="${CONFIGURATION:=Release}"
+CONFIGURATION="${CONFIGURATION:=Debug}"
```
mjcheetham, line 55: Switched the default back to `Debug` from `Release`.

Thinking about it again, we should make these scripts easy to call locally in dev (where we'd want debug bits) since CI doesn't care how long command line args get 😁

src/shared/DotnetTool/pack-tool.sh
```diff
@@ -53,36 +52,25 @@ GITLAB_UI_SRC="$SRC/shared/GitLab.UI.Avalonia"
...
-fi
-
-# Cleanup package directory
-if [ -d "$PACKAGE" ]; then
-    echo "Cleaning existing package directory '$PACKAGE'..."
-    rm -rf "$PACKAGE"
+OUTDIR="$PROJ_OUT/nupkg/$CONFIGURATION"
+IMGOUT="$OUTDIR/images"
+PAYLOAD="$OUTDIR/payload"
+SYMBOLOUT="$OUTDIR/payload.sym"
```
mjcheetham, line 64: Everything sits under `$OUTDIR` now which is `/out/shared/DotnetTool/nupkg/$CONFIGURATION`.

src/shared/DotnetTool/pack-tool.sh
```diff
@@ -91,49 +79,54 @@ fi
...
 #####################################################################
 # Pack dotnet tool
 #####################################################################
 echo "Creating dotnet tool package..."
 
-mkdir -p "$PACKAGE" || exit 1
-echo "Laying out files..."
-cp -r "$SRC/$DOTNET_TOOL/DotnetToolSettings.xml" \
-    "$PAYLOAD/." \
-    "$PACKAGE/"
```
mjcheetham, line 135: No need to copy the output to another directory just to pack it. Instead we're now `dotnet publish`-ing outputs to a payload directory, and copying associated .xml/.png files to the right dir structure in the first place.

ldennington (approved):

PR #913, opened 2022-10-18
Update README.md

aspata11a:
Given the EOL path of CentOS, other successor Distros like Rocky, Alma, and others are RHEL compatible should be added to the distribution case statement.


ldennington:
Closing, as we have [decided to align with dotnet-supported distributions](https://github.com/GitCredentialManager/git-credential-manager/pull/953).

PR #924, opened 2022-10-24
build: fix/speed-up build graph dependencies

mjcheetham:
When building the entire solution, we need to ensure the various "Installer.<OS>" and "Packaging.<OS>" projects are not built at the same time as other projects, especially on Windows, where we may hit file locking problems.

The installer projects call a build script internally that calls a `dotnet publish` on various projects. By making the installer projects reference (but not consume) the dependent projects we ensure MSBuild won't build them until the end, and prevent file access errors.

At the same time, we can probably remove the "-m:1" flags that tell MSBuild to only build one project at a time.

### DIFF
--- a/src/linux/Packaging.Linux/Packaging.Linux.csproj
+++ b/src/linux/Packaging.Linux/Packaging.Linux.csproj
@@ -17,6 +17,10 @@
 
   <ItemGroup>
     <ProjectReference Include="../../shared/Git-Credential-Manager/Git-Credential-Manager.csproj" ReferenceOutputAssembly="false" />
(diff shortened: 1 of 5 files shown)


derrickstolee (approved):
I took a quick glance at our CI history to see how things improved. The end-to-end time seems to have a moderate speedup of 10-15 seconds (based on very limited samples) but individual jobs do get a lot faster. Windows seems to speed up by a minute. The macOS build seems to be the limiter in the build I saw.

Still, the logic here is sound. Thanks!

Favourjacobmudiaga:
Wow

Favourjacobmudiaga:
Good

Favourjacobmudiaga:
Good

PR #926, opened 2022-10-25
WSL documentation updated

rawfg:
After installation of the Git-2.38.1-64-bit.exe I've noticed the instruction for WSL is no longer up-to-date.

The latest Git Credential Manager binary name is `git-credential-manager-core.exe`. See attached screenshot:

image

### DIFF
--- a/docs/wsl.md
+++ b/docs/wsl.md
@@ -22,7 +22,7 @@ _Inside your WSL installation_, run the following command to set GCM as the Git
 credential helper:
 
 ```shell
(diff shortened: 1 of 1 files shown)


ldennington:
Hello @rawfg! Thank you for your submission.

We actually updated our docs in anticipation of our next release, in which we're dropping the `core` suffix from the executable. See [this PR](https://github.com/GitCredentialManager/git-credential-manager/pull/551) for details. Although we're not scheduled to release until November 16, I will chat with other maintainers about a potential out-of-band release before that time since this is causing some confusion.

PR #940, opened 2022-11-03
Fix command-line instructions for .deb install

rendall:
The installation instructions for .deb are incorrect. 

Currently the instructions for .deb installation are to use `git-credential-manager configure`. However, the command line `git-credential-manager configure` returns `git-credential-manager: command not found`. 

Using `git credential-manager` (no dash) returns
```
git: 'credential-manager' is not a git command. See 'git --help'.

The most similar command is
	credential-manager-core
```

Using `git credential-manager-core` works correctly as expected. This PR updates the installation instructions to use `git credential-manager-core`.

Note: I only verified this for the debian package and so this PR only changes those instructions. The instructions for "Tarball" and "Install from source helper script" also use the probably incorrect `git-credential-manager` command.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -77,13 +77,13 @@ Download the latest [.deb package][latest-release], and run the following:
 
 ```shell
 sudo dpkg -i <path-to-package>
-git-credential-manager configure
+git credential-manager-store configure
 ```
 
 #### Uninstall
 
 ```shell
-git-credential-manager unconfigure
+git credential-manager-store unconfigure
 sudo dpkg -r gcmcore
 ```
 


ldennington:
Closing since these docs were purposefully updated to reflect [the rename of the GCM executable](https://github.com/GitCredentialManager/git-credential-manager/pull/551).

rendall:
Ok, but I just followed these instructions and they do not work.

ldennington:
Did you try on the [latest pre-release](https://github.com/GitCredentialManager/git-credential-manager/releases/tag/v2.0.866)?

rendall:
No. 

The instructions as given do not enable the user to configure and install the application.  The PR reflects the way that I was able to do complete the task. If some future release fixes the issue, then great.

If you'd rather, I can create it as an issue or just drop it. Which would you prefer?

ldennington:
Let's create it as an issue with the details of the error you're seeing included. Thanks!

rendall:
Definitely. 👍

PR #942, opened 2022-11-03
diagnose: always use IHttpClientFactory for diagnose

### LINKED ISSUE #927: Command "git credential-manager-core diagnose" fails Networking diagnostic when used with NTLM proxy

From a terminal, run `git credential-manager-core --version` and paste the output.

2.0.785+6074e4e3d3

**Problem description**

I am on a Windows 10 corporate computer behind a proxy that uses NTLM authentication. 

My .gitconfig file has a configuration like this, that is, with empty credentials:

```
[http]
	proxy = http://:@myproxy.mycompany.com:80
```

Commands such as the following work as expected showing the browser window to authenticate:

```
git clone https://myorg@dev.azure.com/myorg/myteamproject/_git/myrepo
```

So far so good. That works (as expected) because credential-manager-core, when gitconfig has empty credentials internally creates an instance of `WebProxy` with `UseDefaultCredentials = true`. See:

https://github.com/GitCredentialManager/git-credential-manager/blob/main/src/shared/Core/HttpClientFactory.cs#L240-L243

However, when I run this command:

```
git credential-manager-core diagnose
```
**Expected behavior**

No exception, all (seven) diagnostics should pass.

**Actual behavior**

The Networking diagnostic fails because it requests proxy authentication (message translated to English):

```
[FAIL] Networking

    [!] Encountered an exception [!]
    System.Net.Http.HttpRequestException: An error occurred while sending the request. ---> System.Net.WebException: The remote server returned an error: (407) Proxy Authentication Required.
```

**Root cause**

The root cause is that the Networking diagnostic creates a `rawHttpClient = new HttpClient()` without proxy, whose call fails and returns:

https://github.com/GitCredentialManager/git-credential-manager/blob/main/src/shared/Core/Diagnostics/NetworkingDiagnostic.cs#L30-L33

and there is another bug a few lines later: the `contextHttpClient = _httpFactory.CreateClient()`, who is created with proxy and that would succeed, is never used, the code of the call uses `rawHttpClient` again:

https://github.com/GitCredentialManager/git-credential-manager/blob/main/src/shared/Core/Diagnostics/NetworkingDiagnostic.cs#L37-L40

mjcheetham:
The Network diagnostic currently makes two passes of the tests, one using a 'raw' `HttpClient` and another using one created by the product `HttpClientFactory`.

However, we've not actually been using the factory-built client in the tests! Furthermore, the raw HTTP client will not exercise test correctly if the machine is behind a corporate proxy, so let's just drop that option.

Fixes #927

### DIFF
--- a/src/shared/Core/Diagnostics/NetworkingDiagnostic.cs
+++ b/src/shared/Core/Diagnostics/NetworkingDiagnostic.cs
@@ -25,23 +25,11 @@ namespace GitCredentialManager.Diagnostics
 
         protected override async Task<bool> RunInternalAsync(StringBuilder log, IList<string> additionalFiles)
         {
-            log.AppendLine("Checking basic networking and HTTP stack...");
+            log.AppendLine("Checking networking and HTTP stack...");
             log.Append("Creating HTTP client...");
-            using var rawHttpClient = new HttpClient();
+            using var httpClient = _httpFactory.CreateClient();
             log.AppendLine(" OK");
 
-            if (!await RunTestAsync(log, rawHttpClient)) return false;
-
-            log.AppendLine("Testing with IHttpClientFactory created HTTP client...");
-            log.Append("Creating HTTP client...");
-            using var contextHttpClient = _httpFactory.CreateClient();
-            log.AppendLine(" OK");
-
-            return await RunTestAsync(log, rawHttpClient);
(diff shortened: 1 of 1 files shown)


ldennington (approved):
Nice cleanup!

PR #951, opened 2022-11-07
Fix problems with application path and installation directory resolution

mjcheetham:
When we introduced the rename warning in #551, there was broken logic for resolving the actual invoked program name on Mac and Linux.

**Windows**
We continue to use `CommandLineToArgvW(GetCommandLine(), ..)` to return the _absolute, full path_ to the entry executable.

**macOS**
Switch from `_NSGetArgv()` to `_NSGetExecutablePath(..)` which, like the Windows APIs above, returns the full path to the entry executable (or symlink).

**Linux**
As there is no equivalent to `CommandLineToArgvW` or `_NSGetExecutablePath` here, we instead do some path computation based on the form of `argv[0]` that we get from `/proc/self/cmdline`.
- If the value is an absolute path, just use that.
- If the value is relative to the current directory (`./name`) then combine this with the current directory.
- If the value contains a directory separator (`dir/name`) then also resolve this from the current directory.
- Otherwise, this `argv[0]` value must have been a file name (`name`) resolved from the `$PATH`.

If we still don't manage to resolve from the `$PATH`, try and resolve the symlink `/proc/self/exe` that points to the executable image that was loaded from disk. Note that we may miss any intermediate link names here, but it's better than nothing.

---

In addition, we also never tested the .NET Tool scenario after updating the `ICommandContext.ApplicationPath` to use the entry executable name, whereas previously this was computed from the .NET assembly path.

Introduce a separate concept, the `InstallationDirectory` that always points to the home of the core assemblies. This is used for resolving things like in-box UI helpers.

### DIFF
--- a/src/shared/Core.Tests/Interop/Windows/WindowsFileSystemTests.cs
+++ b/src/shared/Core.Tests/Interop/Windows/WindowsFileSystemTests.cs
@@ -1,0 +1,77 @@
+using System.IO;
+using GitCredentialManager.Interop.Windows;
+using Xunit;
+using static GitCredentialManager.Tests.TestUtils;
+
+namespace GitCredentialManager.Tests.Interop.Windows
+{
+    public class WindowsFileSystemTests
+    {
+        [PlatformFact(Platforms.Windows)]
+        public static void WindowsFileSystem_IsSamePath_SamePath_ReturnsTrue()
+        {
+            var fs = new WindowsFileSystem();
+
+            string baseDir = GetTempDirectory();
+            string fileA = CreateFile(baseDir, "a.file");
+
+            Assert.True(fs.IsSamePath(fileA, fileA));
+        }
+
+        [PlatformFact(Platforms.Windows)]
+        public static void WindowsFileSystem_IsSamePath_DifferentFile_ReturnsFalse()
+        {
+            var fs = new WindowsFileSystem();
+
+            string baseDir = GetTempDirectory();
+            string fileA = CreateFile(baseDir, "a.file");
+            string fileB = CreateFile(baseDir, "b.file");
+
+            Assert.False(fs.IsSamePath(fileA, fileB));
+            Assert.False(fs.IsSamePath(fileB, fileA));
+        }
+
+        [PlatformFact(Platforms.Windows)]
+        public static void WindowsFileSystem_IsSamePath_SameFileDifferentCase_ReturnsTrue()
+        {
+            var fs = new WindowsFileSystem();
+
+            string baseDir = GetTempDirectory();
+            string fileA1 = CreateFile(baseDir, "a.file");
+            string fileA2 = Path.Combine(baseDir, "A.file");
+
+            Assert.True(fs.IsSamePath(fileA1, fileA2));
+            Assert.True(fs.IsSamePath(fileA2, fileA1));
+        }
+
+        [PlatformFact(Platforms.Windows)]
+        public static void WindowsFileSystem_IsSamePath_SameFileDifferentPathNormalization_ReturnsTrue()
+        {
+            var fs = new WindowsFileSystem();
+
+            string baseDir = GetTempDirectory();
+            string subDir = CreateDirectory(baseDir, "subDir1", "subDir2");
+            string fileA1 = CreateFile(baseDir, "a.file");
+            string fileA2 = Path.Combine(subDir, "..", "..", "a.file");
+
+            Assert.True(fs.IsSamePath(fileA1, fileA2));
+            Assert.True(fs.IsSamePath(fileA2, fileA1));
+        }
+
+        [PlatformFact(Platforms.Windows)]
+        public static void WindowsFileSystem_IsSamePath_SameFileRelativePath_ReturnsTrue()
+        {
+            var fs = new WindowsFileSystem();
+
+            string baseDir = GetTempDirectory();
+            string fileA1 = CreateFile(baseDir, "a.file");
+            string fileA2 = @".\a.file";
+
+            using (ChangeDirectory(baseDir))
+            {
+                Assert.True(fs.IsSamePath(fileA1, fileA2));
+                Assert.True(fs.IsSamePath(fileA2, fileA1));
+            }
+        }
+    }
+}
--- a/src/shared/Core/Interop/Linux/LinuxFileSystem.cs
+++ b/src/shared/Core/Interop/Linux/LinuxFileSystem.cs
@@ -8,8 +8,18 @@ namespace GitCredentialManager.Interop.Linux
     {
         public override bool IsSamePath(string a, string b)
         {
-            a = Path.GetFileName(a);
-            b = Path.GetFileName(b);
+            if (string.IsNullOrWhiteSpace(a) || string.IsNullOrWhiteSpace(b))
+            {
+                return false;
+            }
+
+            // Normalize paths
+            a = Path.GetFullPath(a);
+            b = Path.GetFullPath(b);
+
+            // Resolve symbolic links
+            a = ResolveSymbolicLinks(a);
+            b = ResolveSymbolicLinks(b);
 
             return StringComparer.Ordinal.Equals(a, b);
         }
--- a/src/shared/Core/Interop/Posix/Native/Unistd.cs
+++ b/src/shared/Core/Interop/Posix/Native/Unistd.cs
@@ -1,5 +1,4 @@
 using System.Runtime.InteropServices;
-using System.Text;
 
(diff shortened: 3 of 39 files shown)


src/shared/Core/Interop/Linux/LinuxFileSystem.cs
```diff
@@ -8,8 +8,18 @@ public class LinuxFileSystem : PosixFileSystem
...
+            {
+                return false;
+            }
+
+            // Normalize paths
+            a = Path.GetFullPath(a);
+            b = Path.GetFullPath(b);
+
+            // Resolve symbolic links
+            a = ResolveSymbolicLink(a);
```
dscho, line 21: Does this resolve symlinks recursively? The singular in the function name suggests that it handles only one iteration, i.e. the result of `ResolveSymbolicLink()` might point to _another_ symlink...

src/shared/Core/Interop/Windows/WindowsFileSystem.cs
```diff
@@ -7,8 +7,16 @@ public class WindowsFileSystem : FileSystem
...
+            if (string.IsNullOrWhiteSpace(a) || string.IsNullOrWhiteSpace(b))
+            {
+                return false;
+            }
+
+            a = Path.GetFullPath(a);
+            b = Path.GetFullPath(b);
+
+            // Note: we do not resolve or handle symlinks on Windows
+            // because they require administrator permissions to even create!
```
dscho, line 19: That's not true on Windows 10 in Developer Mode, though.

src/shared/Core/PlatformUtils.cs
```diff
@@ -176,52 +177,109 @@ public static bool IsElevatedUser()
...
+                    }
+                }
+            }
+
+            //
+            // We cannot determine the absolute file path from argv[0]
+            // (how we were launched), so let's now try to extract the
+            // fully resolved executable path from /proc/self/exe.
+            // Note that this means we may miss if we've been invoked
+            // via a symlink, but it's better than nothing at this point!
```
dscho, line 261: I'll play devil's advocate here for a moment: why do we need the absolute path? Is the purpose of this functionality not to determine whether the program was called via the `git-credential-manager-core` symlink?

If that is the case, using `/proc/self/exe` won't give us what we need because it already points to the resolved path of the actual executable, not to the symlink.

And if that is the case, we actually do not need the absolute path at all, right? If we see that `argv[0]` is `git-credential-manager-core` (without any slash, i.e. not an absolute nor a relative path), we know that we want to warn, right?

src/shared/Core/PlatformUtils.cs
```diff
@@ -176,52 +177,109 @@ public static bool IsElevatedUser()
...
+            //
+            FileSystemInfo fsi = File.ResolveLinkTarget("/proc/self/exe", returnFinalTarget: false);
```
dscho, line 263: Since this is Linux-specific code, I guess it is okay to hard-code `/proc/self/exe`. But please note that for OpenBSD, it is [`/proc/curproc/file`](https://github.com/git/git/blob/v2.38.1/config.mak.uname#L292) and for NetBSD it is [`/proc/curproc/exe`](https://github.com/git/git/blob/v2.38.1/config.mak.uname#L316) (yay standards!).

dscho (approved):
Looks good!

I just left a bit of feedback that is hopefully helpful and constructive.

src/shared/Core/Interop/Linux/LinuxFileSystem.cs
mjcheetham: The implementation currently does indeed resolve _all_ links (up to some hardcoded max in the .NET libraries.. like 40 for POSIX and 60 for Windows).

I can rename this to `ResolveSymbolicLinks` (to make it a bit clearer) :-)

src/shared/Core/Interop/Windows/WindowsFileSystem.cs
mjcheetham: But until there's enough of the world on a version of Windows where you don't need admin to create a symlink, then it's not worth our time supporting symlinks at all in Windows, IMO.

src/shared/Core/PlatformUtils.cs
mjcheetham: We don't currently support BSD-variants :)

ldennington (approved):
Great work! ⭐ This was really nice to review - the code is very clean and easy to follow.

src/shared/Core/Interop/Linux/LinuxFileSystem.cs
dscho: > I can rename this to `ResolveSymbolicLinks` (to make it a bit clearer) :-)

I would like that ;-)

src/shared/Core/Interop/Windows/WindowsFileSystem.cs
dscho: We do not have _any_ reliable indicator in Git for Windows as to how many users run in Developer Mode and use symlinks, but there have been enough complaints that we had to implement patches to support that (git-for-windows/git@9086edc588a7e88d9b071fbf030436e5919dd58b, git-for-windows/build-extra@e6085cb6248c10584d0c43168edecf4cf914b556, just to name two).

So I do not believe that the fraction of symlink/Developer Mode users is actually small.

src/shared/Core/PlatformUtils.cs
dscho: Okay, that makes sense. I do suspect, though, that we need to look at `argv[0]` instead of `/proc/self/exe` for that `-core` rename warning.

src/shared/Core/PlatformUtils.cs
dscho: Fair enough. I just wanted to bring it up in case GCM was ever added *BSD support.

src/shared/Core.Tests/Interop/Windows/WindowsFileSystemTests.cs
```diff
@@ -0,0 +1,77 @@
...
+            var fs = new WindowsFileSystem();
+
+            string baseDir = GetTempDirectory();
+            string fileA = CreateFile(baseDir, "a.file");
+            string fileB = CreateFile(baseDir, "b.file");
+
+            Assert.False(fs.IsSamePath(fileA, fileB));
+            Assert.False(fs.IsSamePath(fileB, fileA));
+        }
+
```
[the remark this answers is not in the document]
mjcheetham: Will remove from the Mac file.

src/shared/Core/Interop/Posix/Native/Unistd.cs
```diff
@@ -1,3 +1,4 @@
+using System;
```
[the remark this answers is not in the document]
mjcheetham: It is not.. not sure how it got there.. probably to bring in `IntPtr` at some point that's no longer required. Will cleanup.

src/shared/Core/Interop/Windows/WindowsFileSystem.cs
mjcheetham: But how many people are going to be installing and configuring standalone GCM via a symlink? :-)

src/shared/Core/PlatformUtils.cs
mjcheetham: If we added BSD support (which, the .NET CLR does support..) there are several places we'd need to review. If/when we add support, Linux != BSD. It'll be it's own platform.

src/shared/Core/PlatformUtils.cs
mjcheetham: Indeed we will. But since this is just a warning, and there may be some cases where `argv[0]` is missing/empty, we can fall-back on `/proc/self/exe` to get _a_ valid(*) path to the real executable for the more important `(un)configure` command usage.

(*) unless someone deletes us before we read this link of course 🤡

src/shared/Core/Interop/Windows/WindowsFileSystem.cs
mjcheetham: Thanks for the context however on symlink usage for G4W 👍

src/shared/Core/Interop/Linux/LinuxFileSystem.cs
FranciscoGuerrerosaavedra: [](url)

Favourjacobmudiaga:
Good

Favourjacobmudiaga:
Good

Favourjacobmudiaga:
Good

PR #953, opened 2022-11-08
linux: align with dotnet distros

ldennington:
Update `install-from-source.sh` and `validate-install-from-source` to align with dotnet-supported Linux distributions. Clarify in `README.md` that the dotnet-supported distributions are the only ones for which GCM guarantees support.


derrickstolee (approved):
Wow, this is excellent. I'm really happy to see this beautiful job list in your test run of the workflow:
image

vtbassmatt (approved):

PR #954, opened 2022-11-08
Ensure certificate validation works on Mac/Linux

mjcheetham:
In 65cead2b5caa we removed the .NET Standard target in favour of a direct .NET 6 target, however we missed this conditional compilation symbol referrering to `NETSTANDARD`.


ldennington (approved):
Good catch!

PR #957, opened 2022-11-15
docs: fix executable name in Linux install instructions

RussellJoyce:
Executable installed by Debian package or Tarball is `git-credential-manager-core`, not `git-credential-manager`, so this should be used for the configure command.


RussellJoyce:
Sorry, I should have checked closed PRs before submitting this - I now see this has already been addressed in #940.

It still feels like a problem though, given the installation instructions also link to a release that contains this old executable.

ldennington:
See https://github.com/GitCredentialManager/git-credential-manager/issues/958

PR #963, opened 2022-11-17
git@github.com:admin-01-web

### LINKED ISSUE #820: dotnet tool install git-credential-manager

admin-01-web:
[5f58e77fe5d9b5d350337fdf281cdbcb0d2ad0c1]
Fixes #820
# Security

If you discover a security issue in this repo, please submit it through the. [GitHub Security Bug Bounty][hackerone-github]

Thanks for helping make GitHub products safe for everyone.

[hackerone-github+https://hackerone.com/github]
[492x0w.webp.pdf](https://github.com/GitCredentialManager/git-credential-manager/files/10032941/492x0w.webp.pdf)
[downloadfile.pdf]+git@github.com:admin-01-web/git-credential-manager.git


ytarsha:
Not me.. I think. I've been hacked

admin-01-web:
>  @`[git@github.com](url)`

admin-01-web:
> Not me.. I think. I've been hacked

5f58e77fe5d9b5d350337fdf281cdbcb0d2ad0c1

PR #966, opened 2022-11-18
homebrew: update install location for arm64

ldennington:
Homebrew updated the install location for Apple Silicon machines to
`/opt/homebrew` (instead of `/usr/local`). Pass runtime information to pack.sh
to determine when this new prefix should be used and update the install
location accordingly.

### DIFF
--- a/src/osx/Installer.Mac/pack.sh
+++ b/src/osx/Installer.Mac/pack.sh
@@ -13,7 +13,6 @@ INSTALLER_SRC="$SRC/osx/Installer.Mac"
 
 # Product information
 IDENTIFIER="com.microsoft.gitcredentialmanager"
-INSTALL_LOCATION="/usr/local/share/gcm-core"
 
 # Parse script arguments
 for i in "$@"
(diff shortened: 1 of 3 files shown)


src/osx/Installer.Mac/pack.sh
```diff
@@ -11,9 +11,16 @@ SRC="$ROOT/src"
 OUT="$ROOT/out"
 INSTALLER_SRC="$SRC/osx/Installer.Mac"
 
+if [ ! -z "$HOMEBREW_PREFIX" ]; then
+    echo "Running in Homebrew environment, installing to Homebrew-recognized path..."
+    PREFIX="$HOMEBREW_PREFIX"
+else
+    PREFIX="/usr/local"
+fi
```
mjcheetham, line 19: `pack.sh` isn't going to be run in a Homebrew environment. We need to switch and hardcode the default prefixes for the different target runtimes:

Runtime|`$PREFIX`
-|-
`osx-x64`|`/usr/local`
`osx-arm64`|`/opt/homebrew`

We should add a `--runtime` argument to this script and pass the correct RID through from `build.sh` as well as from `release.yml`.

```shell
# Parse script arguments
for i in "$@"
do
...
    --runtime=*)
    RUNTIME="${i#*=}"
    shift # past argument=value
    ;;
...
```

```shell
if [ -z "$RUNTIME" ]; then
    die "--runtime was not set"
fi

case "$RUNTIME" in
    "osx-x64")
        PREFIX="/usr/local"
        ;;
    "osx-arm64")
        PREFIX="/opt/homebrew"
        ;;
    *)
        die "Unknown runtime '$RUNTIME'"
        ;;
esac

...

INSTALL_LOCATION="$PREFIX/share/gcm-core"
```

mjcheetham (changes requested):

src/osx/Installer.Mac/pack.sh
mjcheetham: I should also mention that the `conclusion.html` resource file that gets included in the installer package also references `/usr/local` so that will need to be parameterised/swapped like we do for the `distribution.$ARCH.xml` file.

ldennington: Gotcha. I think the issue I have with this is that we're always going to install `osx-arm64` to `/opt/homebrew`, right? Is this desired behavior when a user is installing the `.pkg` directly from the releases page, for example?

ldennington: Resolved offline!

derrickstolee (approved):
Thanks for working so hard on this. Do you need to wait for @mjcheetham to test this, or are you confident in this change?

PR #987, opened 2022-12-08
install from source: publish before attempting install

ldennington:
The changes from 3978b0e introduced a bug in the install from source workflow - we were attempting to copy binaries before publishing them! Fixing with this change.

Validated with [this passing workflow](https://github.com/ldennington/git-credential-manager/actions/runs/3652773502) in my fork.


mjcheetham (approved):
Good catch! I didn't test the install from source option :(

ldennington:
> I didn't test the install from source option :(

Yea - I accidentally tested it on a machine that already had the published binaries in the expected location. Whoops!

PR #994, opened 2022-12-14
GCM Release

ldennington:
Changes:
- Fix various problems with install path and app path variables #968
- Sign Linux tarball and .NET tool #985


mjcheetham (approved):
:shipit:

PR #995, opened 2022-12-14
release: correct nupkg path for publishing

ldennington:
When we began signing the .NET tool in 80cc677, we did not update the path used for publishing to nuget.org. Fixing with this change.


dscho (approved):
Looks obviously correct. Thank you for working on this!

PR #996, opened 2022-12-16
secret service: fix error creating credential

### LINKED ISSUE #793: libsecret: sec_free: Assertion `cell->requested > 0' failed

Credential Manager: 2.0.778+0d6f1c846f (Arch Linux)
Linux Kernel: 5.18.7-zen1-1-zen
Libsecret: 0.20.5-2

* [ ] Azure DevOps
* [ ] GitHub
* [ ] GitHub Enterprise
* [ ] Bitbucket
* [x] Other - please describe: **GitLab** self hosted instance

* [x] Yes

---

* [x] Not applicable
* [ ] https://dev.azure.com/`{org}`/...
* [ ] https://`{org}`@dev.azure.com/`{org}`/...
* [ ] https://`{org}`.visualstudio.com/...

* [x] Not applicable
* [ ] I only see one identity
* [ ] I checked each identity and none worked

---

**Expected behavior**

I am authenticated and my Git operation completes successfully.

**Actual behavior**

An exception is thrown:
```
OAuth token refresh failed: invalid_grant: The provided authorization grant is invalid, expired, revoked, does not match the redirection URI used in the authorization request, or was issued to another client.
git-credential-manager-core: ../libsecret/egg/egg-secure-memory.c:596: sec_free: Assertion `cell->requested > 0' failed.
error: /usr/lib/share/git-credential-manager-core/git-credential-manager-core get died of signal 6
remote: HTTP Basic: Access denied
```

OAuth response shows `Authentication successful` in the browser though.

**Logs**

I will attach as soon as it happens again. The error happened multiple times, so I believe it will be thrown again shortly.

ldennington:
Multiple Linux users have reported that they are unable to use Secret Service as their credential store, as GCM throws the following error:

sec_free: Assertion `cell->requested > 0' failed.

The root cause is that we're using the libsecret secret_value_get function to obtain secret data, then attempting to free the string with secret_password_free. It appears that secret_password_free is only meant to be used to free nonpageable memory [1], however. Removing this call fixes the issue, as verified with a successful git-credential-manager diagnose (which was previously failing with the above error).

[1] secret_password_free manpage
https://www.manpagez.com/html/libsecret-1/libsecret-1-0.18.6/libsecret-Password-storage.php#secret-password-free

Fixes #793

### DIFF
--- a/src/shared/Core/Interop/Linux/SecretServiceCollection.cs
+++ b/src/shared/Core/Interop/Linux/SecretServiceCollection.cs
@@ -291,7 +291,6 @@ namespace GitCredentialManager.Interop.Linux
                 if (accountKeyPtr != IntPtr.Zero) Marshal.FreeHGlobal(accountKeyPtr);
                 if (serviceKeyPtr != IntPtr.Zero) Marshal.FreeHGlobal(serviceKeyPtr);
                 if (value != null) secret_value_unref(value);
-                if (passwordPtr != IntPtr.Zero) secret_password_free(passwordPtr);
                 if (error != null) g_error_free(error);
             }
         }


dscho (approved):
This will make so many libsecret users happy!

mjcheetham (commented):
Nice job finding this!

js6pak:
> Nice job finding this!

Thanks :)

> only meant to be used to free nonpageable memory

[You have to free pageable memory too](https://www.manpagez.com/html/libsecret-1/libsecret-1-0.18.6/libsecret-Password-storage.php#secret-password-lookup-sync), its just that GCM doesn't use `secret_password_lookup` apis for whatever reason

ldennington:
Lol indeed I can't take credit for identifying the fix. I just followed @js6pak's suggestion.

PR #1007, opened 2022-12-22
docs(install): fix link to the ".NET tool" section

dscho:
Something I noticed today, while looking for a way to install Git Credential Manager for Windows/ARM64.


ldennington (approved):
Thanks very much @dscho!

Issue #997, opened 2022-12-16
Can't unpack latest GCM deb package


dotCore-off:
## GCM Version
> 2.0.886

## OS Version
> Debian 11 Bullseye

## Description
> Using ``dpkg -i`` on latest available .deb package returns an error and prevent unpacking GCM

## Error log
```log
~ sudo dpkg -i gcm-linux_amd64.2.0.886.deb
dpkg-deb: error: archive 'gcm-linux_amd64.2.0.886.deb' uses unknown compression for member 'control.tar.zst', giving up
dpkg: error processing archive gcm-linux_amd64.2.0.886.deb (--install):
 dpkg-deb --control subprocess returned error exit status 2
Errors were encountered while processing:
 gcm-linux_amd64.2.0.886.deb
```

## Expected log (using previous GCM release)
```log
~ sudo dpkg -i gcm-linux_amd64.2.0.877.deb
Selecting previously unselected package gcm.
(Reading database ... 20058 files and directories currently installed.)
Preparing to unpack gcm-linux_amd64.2.0.877.deb ...
Unpacking gcm (2.0.877) ...

Setting up gcm (2.0.877) ...
```

ldennington:
This appears to be a bug in Debian's `dpkg` tool (based on what I'm reading [here](https://unix.stackexchange.com/questions/669004/zst-compression-not-supported-by-apt-dpkg)). I would recommend instead attempting to install GCM as a [.NET tool](https://github.com/GitCredentialManager/git-credential-manager/blob/main/docs/install.md#install-5), using the [tarball](https://github.com/GitCredentialManager/git-credential-manager/blob/main/docs/install.md#install-3), or [installing from source](https://github.com/GitCredentialManager/git-credential-manager/blob/main/docs/install.md#install-4).

dennmart:
Not sure if this is related but there might be an issue in Linux builds in general. I noticed when getting the latest Linux tarball from the [v2.0.886 release](https://github.com/GitCredentialManager/git-credential-manager/releases/tag/v2.0.886) that the file is only 481 bytes. Opening the file in a text editor shows a PGP signature.

dotCore-off:
Indeed, it seems to be an issue from ``dpkg``, but something changed on GCM side.
Latest release doesn't work while previous one ([GCM 2.0.877](https://github.com/GitCredentialManager/git-credential-manager/releases/tag/v2.0.877)) does. 
[...]

PR #1009, opened 2022-12-28
release: fix tarball signing and Debian compression

ldennington:
While we added PGP signatures for tarballs in 7baac73, we did not notice
that, while ESRP returns a file with the tar.gz extension, it is actually
the signature file, not the tarball itself. Correct with this change and
validate tarball moving forward so it doesn't happen again!

Additionally, a user reported in #997 that the latest GCM Debian package
didn't work on Debian distributions. It appears that the version of dpkg
that ships with Debian [does not support zstd compression](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=892664). Enforcing xz
compression resolves the issue.
    
Finally, this provided an opportunity to clean up some unused variables in
pack.sh for Linux and to ensure we check the architecture is found before
attemping to use the ARCH variable.

These changes were validated with [this successful test run](https://github.com/ldennington/git-credential-manager/actions/runs/3795232469) in my fork.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -384,7 +384,7 @@ jobs:
     - name: Upload artifacts
       uses: actions/upload-artifact@v3
       with:
-        name: tmp.linux-build
+        name: linux-build
         path: |
           linux-build
 
@@ -399,7 +399,11 @@ jobs:
     - name: Download artifacts
       uses: actions/download-artifact@v3
       with:
-        name: tmp.linux-build
+        name: linux-build
+    
+    - name: Remove symbols
+      run: |
+        rm tar/*symbols*
 
     - uses: azure/login@v1
       with:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -423,6 +427,12 @@ jobs:
       run: |
         python .github/run_esrp_signing.py deb $env:LINUX_KEY_CODE $env:LINUX_OP_CODE
         python .github/run_esrp_signing.py tar $env:LINUX_KEY_CODE $env:LINUX_OP_CODE
+    
+    - name: Re-name tarball signature file
+      shell: bash
+      run: |
+        signaturepath=$(find signed/*.tar.gz)
+        mv "$signaturepath" "${signaturepath%.tar.gz}.asc"
```
mjcheetham, line 435: My only question is should we not also upload/save the signature too?

mjcheetham (approved):
LGTM

.github/workflows/release.yml
ldennington: Do you mean upload to the release? If so, we're already doing that! Here's the dummy release from my test run:

https://github.com/ldennington/git-credential-manager/releases/tag/untagged-4129f63ec90f2c23944a

mjcheetham: Oh! If we're already uploading the signature then that's great!

PR #1037, opened 2023-01-18
Git Credential Manager Path Update

aarondake:
I am running WSL2 and just installed the latest version of Git For Windows, and this was the path I needed to use.


mjcheetham:
Hello @aarondake! 👋 

This change is not correct to configure WSL2 Git with GCM. The path you are using is for the `wincred` credential helper, separate from the GCM helper (`git-credential-manager.exe`), but just also shipped with Git for Windows.

Issue #789, opened 2022-07-12
Refresh tokens for GitHub.


adam-azarchs:
I'm not 100% sure, but as far as I can tell, at the moment the GitHub OAuth2 flow is generating long-lived tokens. It would be nice to have the ability to use [short-lived credentials + refresh tokens](https://docs.github.com/en/developers/apps/building-github-apps/refreshing-user-to-server-access-tokens).  That flow appears to already be mandatory for BitBucket; it would be nice to have that as at least an option for GitHub as well.

Short-lived oauth tokens with refresh tokens are more secure than long-lived tokens because if your token gets exfiltrated then it'll expire quickly, and if the attacker uses the refresh token then you should at least be able to detect that quickly.  I'm sure long-lived tokens are fine for many situations, but for organizations which are more paranoid, or where the risk of exposure is higher (e.g. situations where developers are logging in from a shared machine) it would be helpful to have the ability to enforce that higher level of security.

Implementation-wise this should be pretty straightforward, since it's already supported for BitBucket and GitLab, though it might require making a separate client secret for GCM in this mode, since it looks from the docs like GitHub's configuration on whether oauth credentials are short-lived or not is on a per-application basis.

hickford:
Good question. [Might access token expiry even have prevented or reduced the 2022 stolen OAuth token attack?](
https://security.stackexchange.com/questions/263314/would-refresh-tokens-have-prevented-the-github-stolen-oauth-token-attack) FWIW [GitHub *Apps*](https://docs.github.com/en/developers/apps/getting-started-with-apps/differences-between-github-apps-and-oauth-apps#token-based-identification) offer expiring access tokens [as an optional feature](https://docs.github.com/en/developers/apps/building-github-apps/identifying-and-authorizing-users-for-github-apps).  Why not offer expiring access tokens for OAuth too?

Interestingly Git*Lab* recently [removed non-expiring access tokens in version 15.0](https://docs.gitlab.com/ee/update/removals.html#oauth-tokens-without-an-expiration). This is breaking change for apps which don't support refresh flow -- users would have to reauthenticate every 2 hours.

mjcheetham:
This issue requires migration to the GitHub App infrastructure, which has several issues when using a public client app rather than a confidential client (specifically around org installation and consent).

Blocked on GitHub platform enhancements.

PR #1062, opened 2023-01-27
Add OAuth support to generic host provider

mjcheetham:
Add ability to provide OAuth-based authentication for generic hosts by way of simple Git configuration.

When a remote URL does not match any known host provider plugin, the generic provider will now first check for OAuth configuration in the Git config or environment variables. If such config is available then we try and perform OAuth authentication. Support for device code flow is optional, and refresh tokens will be used if the service supports and returns them.

Users can make use of existing Git config `include` to easily organise and share custom OAuth configurations:

_~/.gitconfig_

```ini
[include]
    path = ~/.gcm/oauth/example.ini
```

_example.ini_

```ini
[credential "https://example.com"]
	provider = generic
	oauthClientId = <client-id>
	oauthClientSecret = <client-secret>
	oauthRedirectUri = <redirect-uri>
	oauthAuthorizeEndpoint = <auth-endpoint>
	oauthTokenEndpoint = <token-endpoint>
	oauthScopes = <scopes>
	oauthDeviceEndpoint = <device-endpoint>
	oauthDefaultUserName = <username>
	oauthUseClientAuthHeader = <true|false>
```

image

image

Fixes https://github.com/GitCredentialManager/git-credential-manager/issues/1047

### DIFF
--- a/docs/generic-oauth.md
+++ b/docs/generic-oauth.md
@@ -1,0 +1,116 @@
+# Generic Host Provider OAuth
+
+Many Git hosts use the popular standard OAuth2 or OpenID Connect (OIDC)
+authentication mechanisms to secure repositories they host.
+Git Credential Manager supports any generic OAuth2-based Git host by simply
+setting some configuration.
+
+## Registering an OAuth application
+
+In order to use GCM with a Git host that supports OAuth you must first have
+registered an OAuth application with your host. The instructions on how to do
+this can be found with your Git host provider's documentation.
+
+When registering a new application, you should make sure to set an HTTP-based
+redirect URL that points to `localhost`; for example:
+
+```text
+http://localhost
+http://localhost:<port>
+http://127.0.0.1
+http://127.0.0.1:<port>
+```
+
+Note that you cannot use an HTTPS redirect URL. GCM does not require a specific
+port number be used; if your Git host requires you to specify a port number in
+the redirect URL then GCM will use that. Otherwise an available port will be
+selected at the point authentication starts.
+
+You must ensure that all scopes required to read and write to Git repositories
+have been granted for the application or else credentials that are generated
+will cause errors when pushing or fetching using Git.
+
+As part of the registration process you should also be given a Client ID and,
+optionally, a Client Secret. You will need both of these to configure GCM.
+
+## Configure GCM
+
+In order to configure GCM to use OAuth with your Git host you need to set the
+following values in your Git configuration:
+
+- Client ID
+- Client Secret (optional)
+- Redirect URL
+- Scopes (optional)
+- OAuth Endpoints
+  - Authorization Endpoint
+  - Token Endpoint
+  - Device Code Authorization Endpoint (optional)
+
+OAuth endpoints can be found by consulting your Git host's OAuth app development
+documentation. The URLs can be either absolute or relative to the host name;
+for example: `https://example.com/oauth/authorize` or `/oauth/authorize`.
+
+In order to set these values, you can run the following commands, where `<HOST>`
+is the hostname of your Git host:
+
+```shell
+git config --global credential.<HOST>.oauthClientId <ClientID>
+git config --global credential.<HOST>.oauthClientSecret <ClientSecret>
+git config --global credential.<HOST>.oauthRedirectUri <RedirectURL>
+git config --global credential.<HOST>.oauthAuthorizeEndpoint <AuthEndpoint>
+git config --global credential.<HOST>.oauthTokenEndpoint <TokenEndpoint>
+git config --global credential.<HOST>.oauthScopes <Scopes>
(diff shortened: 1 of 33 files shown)


src/shared/Core/GenericHostProvider.cs
```diff
@@ -86,10 +109,94 @@ public override async Task<ICredential> GenerateCredentialAsync(InputArguments i
...
+            var client = new OAuth2Client(
+                HttpClient,
+                config.Endpoints,
+                config.ClientId,
+                config.RedirectUri,
+                config.ClientSecret,
+                Context.Trace,
+                config.UseAuthHeader);
+
+            string refreshService = remoteUri.AbsoluteUri.TrimEnd('/') + "/refresh_token";
```
mjcheetham, line 131: I know there are some concerns about how we generate a service name for the refresh tokens with collisions with other credentials.

@hickford, you had suggested in the past using a subdomain to differentiate ATs and RTs:

```text
https://foo.example.com/path/repo
https://refresh_token.foo.example.com/path/repo
```

Did you or @ldennington have any other suggestions? I was going to change to the `refresh_token` subdomain example above if there was no objections.

mjcheetham:
Right now there's no facility to 'test' an existing OAuth credential is valid, so we rely on the less-than-optimal flow of Git trying, failing, telling us to erase, and then expect the user to retry and then we'll use the refresh token.

In the future we may wish to allow a 'validity' endpoint to be configured where the generic implementation can make a GET request with the bearer token, and use the HTTP status code as a signal for validity (200 = OK, 401/403 => invalid).

docs/generic-oauth.md
```diff
@@ -0,0 +1,116 @@
...
+following values in your Git configuration:
+
+- Client ID
+- Client Secret (optional)
+- Redirect URL
+- Scopes (optional)
+- OAuth Endpoints
+  - Authorization Endpoint
+  - Token Endpoint
+  - Device Code Authorzation Endpoint (optional)
```
ldennington, line 48:
``` suggestion
  - Device Code Authorization Endpoint (optional)
```

docs/generic-oauth.md
```diff
@@ -0,0 +1,116 @@
...
+setting some configuration.
+
+## Registering an OAuth application
+
+In order to use GCM with a Git host that supports OAuth you must first have
+registered an OAuth application with your host. The instructions on how to do
+this can be found with your Git host provider's documentation.
+
+When registering a new application you should make sure to set an HTTP based
+redirect URL that points to `localhost`; for example:
```
ldennington, line 15:
```suggestion
When registering a new application, you should make sure to set an HTTP-based
redirect URL that points to `localhost`; for example:
```

src/shared/Core/GenericHostProvider.cs
ldennington: I'm fine with the subdomain implementation.

src/shared/Core/GenericHostProvider.cs
```diff
@@ -128,6 +128,33 @@ private async Task<ICredential> GetOAuthAccessToken(Uri remoteUri, string userNa
                 Context.Trace,
                 config.UseAuthHeader);
 
+            //
```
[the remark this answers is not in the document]
mjcheetham: I've used this pattern in other places in the code base to visually make a block of explanatory comments standout more.

src/shared/Git-Credential-Manager.UI.Avalonia/Views/OAuthView.axaml.cs
```diff
@@ -0,0 +1,43 @@
...
+using Avalonia.Markup.Xaml;
+using GitCredentialManager.UI.Controls;
+using GitCredentialManager.UI.ViewModels;
+
+namespace GitCredentialManager.UI.Views
+{
+    public class OAuthView : UserControl, IFocusable
+    {
+        private TextBox _userNameTextBox;
+        private TextBox _passwordTextBox;
```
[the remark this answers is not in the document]
mjcheetham: ..because I didn't delete this from the code-behind file after copying it from another view 😅

src/shared/Core/GenericOAuthConfig.cs
```diff
@@ -0,0 +1,138 @@
...
+                    Constants.GitConfiguration.Credential.OAuthClientAuthHeader,
+                    out string useHeader))
+            {
+                config.UseAuthHeader = useHeader.IsTruthy();
+            }
+            else
+            {
+                // Default to true
+                config.UseAuthHeader = true;
+            }
```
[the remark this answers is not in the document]
mjcheetham: Removed the default property value from the class.

ANGGBELPAS:
Duplicate of #789

Issue #694, opened 2022-05-12
When using the Secret Service credential store, if the keyring is locked at the time when I run a Git operation then GCM deletes the credential from the keyring and fails to authenticate with the remote


gaazkam:
**Which version of GCM are you using?**

Git Credential Manager version 2.0.696+4365b917da (Ubuntu 20.04.4 LTS on VirtualBox, almost fresh installation)

As per your instructions https://github.com/GitCredentialManager/git-credential-manager/blob/main/docs/credstores.md I instructed GCM to use the freedesktop.org Secret Service API as its credential store.

This is how I set up GCM:

```
sudo dpkg -i Downloads/gcmcore-linux_amd64.2.0.696.deb
git-credential-manager-core configure
git config --global user.name gaazkam
git config --global user.email "26407962+gaazkam@users.noreply.github.com"
git config --global credential.credentialStore secretservice
git clone https://github.com/gaazkam/mon.git
cd mon
git push
```

After the last command, as expected, I had to log in into Github and a keyring was created holding the credentials to my Github account.

Problems described below start when the keyring gets locked (eg because I reboot the system; note that the login keyring does not get unlocked on login if the system is set to automatic login, without having to enter the password).

**Which Git host provider are you trying to connect to?**

* [ ] Azure DevOps
* [ ] Azure DevOps Server (TFS/on-prem)
* [x] GitHub
* [ ] GitHub Enterprise
* [ ] Bitbucket
* [ ] Other - please describe

**Can you access the remote repository directly in the browser using the remote URL?**

* [x] Yes
* [ ] No, I get a permission error
* [ ] No, for a different reason - please describe

---

**_[Azure DevOps only]_ What format is your remote URL?**

* [x] Not applicable
* [ ] https://dev.azure.com/`{org}`/...
* [ ] https://`{org}`@dev.azure.com/`{org}`/...
* [ ] https://`{org}`.visualstudio.com/...

**_[Azure DevOps only]_ If the account picker shows more than one identity as you authenticate, check that you selected the same one that has access on the web.**

* [x] Not applicable
* [ ] I only see one identity
* [ ] I checked each identity and none worked

---

**Expected behavior**

If the keyring that holds the credentials to my Github account is locked at the time when GCM needs to access it then a popup shows asking me to enter the password to unlock this keyring. When I enter the correct password I am authenticated and my Git operation completes succesfully. The credential is not deleted.

**Actual behavior**

[...]

PR #1080, opened 2023-01-31
Warn that using the secure keystore on Linux might fail

fatso83:
I was trying to get working for the longest time, until I tried to diagnose and found it failed on #694.


fatso83:
Fix in place that will be released on Feb 17:
https://github.com/GitCredentialManager/git-credential-manager/issues/694#issuecomment-1410845266

PR #1131, opened 2023-02-27
trace2: add child start and exit events

ldennington:
This change adds the `child_start` and `child_exit` events to GCM's `TRACE2` tracing system.

The first two patches are small bug fixes. The first fix ensures all events are written in snake case. Up until this point, we've only written single word events, but the `TRACE2` convention for multi-word events is to write in snake case, so for simplicity we update all events to use this convention. The second fix ensures we do _not_ redirect standard error for Git-related processes, as this causes deadlocks with `TRACE2` tracing enabled. 

The third patch contains a refactor of the `Trace2` class necessary to support `child_start` and `child_exit`. 
The main change is addition of an `Initialize()` method to `Trace2.cs` to detect whether `TRACE2` tracing is enabled and to set long-lived variables. This also sets a new `_initialized` property that allows us to determine whether `TRACE2` tracing has been initialized. GCM `TRACE2` tracing is disabled until `_initialized` is set to true by the `Initialize()` method (although it is worth noting that Git's `TRACE2` tracing is not disabled and will be reflected in logs).

The third patch pulls Sid-related logic into a separate `SidManager` class to  improve separation of concerns and make Sid logic less error-prone/easier to reason about.

The fourth patch adds a `ChildProcess` class that wraps `Process` and gives us the required control to write process-related `TRACE2` events. It also decouples process-related logic from environment logic to maintain a functional, non-circular order of dependencies with the new wrapper. All `Process` invocations have been updated to use `ChildProcess`.

The fifth patch adds the appropriate `Trace2` methods to write `child_start` and `child_exit` events.

### DIFF
--- a/src/shared/Atlassian.Bitbucket.UI.Avalonia/Program.cs
+++ b/src/shared/Atlassian.Bitbucket.UI.Avalonia/Program.cs
@@ -45,6 +45,9 @@ namespace Atlassian.Bitbucket.UI
         {
             string[] args = (string[]) o;
 
+            // Set the session id (sid) for the helper process, to be
+            // used when TRACE2 tracing is enabled.
+            SidManager.CreateSid();
             using (var context = new CommandContext(args))
             using (var app = new HelperApplication(context))
             {
--- a/src/shared/Atlassian.Bitbucket.UI.Avalonia/Program.cs
+++ b/src/shared/Atlassian.Bitbucket.UI.Avalonia/Program.cs
@@ -48,9 +48,15 @@ namespace Atlassian.Bitbucket.UI
             // Set the session id (sid) for the helper process, to be
             // used when TRACE2 tracing is enabled.
             SidManager.CreateSid();
-            using (var context = new CommandContext(args))
+            using (var context = new CommandContext())
             using (var app = new HelperApplication(context))
             {
+                // Initialize TRACE2 system
+                context.Trace2.Initialize(DateTimeOffset.UtcNow);
+
+                // Write the start and version events
+                context.Trace2.Start(context.ApplicationPath, args);
+
                 app.RegisterCommand(new CredentialsCommandImpl(context));
 
                 int exitCode = app.RunAsync(args)
--- a/src/shared/Core.Tests/StringExtensionsTests.cs
+++ b/src/shared/Core.Tests/StringExtensionsTests.cs
@@ -293,5 +293,20 @@ namespace GitCredentialManager.Tests
             string actual = StringExtensions.TrimMiddle(input, trim, comparisonType);
             Assert.Equal(expected, actual);
         }
+
+        [Theory]
+        [InlineData("FooBar", "foo_bar")]
+        [InlineData("fooBar", "foo_bar")]
+        [InlineData("FBBaz", "fb_baz")]
+        [InlineData("Foo", "foo")]
+        [InlineData("Fo", "fo")]
+        [InlineData("fO", "f_o")]
+        [InlineData("OO", "oo")]
+        [InlineData("F", "f")]
+        [InlineData("", "")]
+        public void StringExtensions_ToSnakeCase_Converts_Correctly(string input, string expected)
+        {
+            Assert.Equal(expected, input.ToSnakeCase());
+        }
     }
 }
--- a/src/shared/Core.Tests/WslUtilsTests.cs
+++ b/src/shared/Core.Tests/WslUtilsTests.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Diagnostics;
+using Moq;
 using Xunit;
 
 namespace GitCredentialManager.Tests
@@ -100,7 +101,7 @@ namespace GitCredentialManager.Tests
             string expectedFileName = WslUtils.GetWslPath();
             string expectedArgs = $"--distribution {distribution} --exec {command}";
 
-            Process process = WslUtils.CreateWslProcess(distribution, command);
+            ChildProcess process = WslUtils.CreateWslProcess(distribution, command, Mock.Of<ITrace2>());
 
             Assert.NotNull(process);
             Assert.Equal(expectedArgs, process.StartInfo.Arguments);
@@ -121,7 +122,7 @@ namespace GitCredentialManager.Tests
             string expectedFileName = WslUtils.GetWslPath();
             string expectedArgs = $"--distribution {distribution} --exec {command}";
 
-            Process process = WslUtils.CreateWslProcess(distribution, command, expectedWorkingDirectory);
+            ChildProcess process = WslUtils.CreateWslProcess(distribution, command, Mock.Of<ITrace2>(), expectedWorkingDirectory);
 
             Assert.NotNull(process);
             Assert.Equal(expectedArgs, process.StartInfo.Arguments);
--- a/src/shared/Core.Tests/WslUtilsTests.cs
+++ b/src/shared/Core.Tests/WslUtilsTests.cs
@@ -107,7 +107,7 @@ namespace GitCredentialManager.Tests
             Assert.Equal(expectedFileName, process.StartInfo.FileName);
             Assert.True(process.StartInfo.RedirectStandardInput);
             Assert.True(process.StartInfo.RedirectStandardOutput);
-            Assert.True(process.StartInfo.RedirectStandardError);
+            Assert.False(process.StartInfo.RedirectStandardError);
             Assert.False(process.StartInfo.UseShellExecute);
         }
 
@@ -128,7 +128,7 @@ namespace GitCredentialManager.Tests
             Assert.Equal(expectedFileName, process.StartInfo.FileName);
             Assert.True(process.StartInfo.RedirectStandardInput);
             Assert.True(process.StartInfo.RedirectStandardOutput);
-            Assert.True(process.StartInfo.RedirectStandardError);
+            Assert.False(process.StartInfo.RedirectStandardError);
             Assert.False(process.StartInfo.UseShellExecute);
             Assert.Equal(expectedWorkingDirectory, process.StartInfo.WorkingDirectory);
         }
--- a/src/shared/Core/ChildProcess.cs
+++ b/src/shared/Core/ChildProcess.cs
@@ -11,6 +11,7 @@ public class ChildProcess : DisposableObject
 
     private DateTimeOffset _startTime;
     private DateTimeOffset _exitTime => Process.ExitTime;
+    private ProcessStartInfo _startInfo => Process.StartInfo;
 
     private int _id => Process.Id;
 
@@ -19,13 +20,12 @@ public class ChildProcess : DisposableObject
     public StreamWriter StandardInput => Process.StandardInput;
     public StreamReader StandardOutput => Process.StandardOutput;
     public StreamReader StandardError => Process.StandardError;
-    public int Id => Process.Id;
     public int ExitCode => Process.ExitCode;
 
-    public static ChildProcess Start(ITrace2 trace2, ProcessStartInfo startInfo)
+    public static ChildProcess Start(ITrace2 trace2, ProcessStartInfo startInfo, Trace2ProcessClass processClass)
     {
         var childProc = new ChildProcess(trace2, startInfo);
-        childProc.Start();
+        childProc.Start(processClass);
         return childProc;
     }
 
@@ -33,11 +33,24 @@ public class ChildProcess : DisposableObject
     {
         _trace2 = trace2;
         Process = new Process() { StartInfo = startInfo };
+        Process.Exited += ProcessOnExited;
     }
 
-    public void Start()
+    public void Start(Trace2ProcessClass processClass)
     {
         ThrowIfDisposed();
+        // Record the time just before the process starts, since:
+        // (1) There is no event related to Start as there is with Exit.
+        // (2) Using Process.StartTime causes a race condition that leads
+        // to an exception if the process finishes executing before the
+        // variable is passed to Trace2.
+        _startTime = DateTimeOffset.UtcNow;
+        _trace2.WriteChildStart(
+            _startTime,
+            processClass,
+            _startInfo.UseShellExecute,
+            _startInfo.FileName,
+            _startInfo.Arguments);
         Process.Start();
     }
 
@@ -47,7 +60,20 @@ public class ChildProcess : DisposableObject
 
     protected override void ReleaseManagedResources()
     {
+        Process.Exited -= ProcessOnExited;
         Process.Dispose();
         base.ReleaseUnmanagedResources();
+    }
+
+    private void ProcessOnExited(object sender, EventArgs e)
+    {
+        if (sender is Process)
+        {
+            double elapsedTime = (_exitTime - _startTime).TotalSeconds;
+            _trace2.WriteChildExit(
+                elapsedTime,
+                _id,
+                Process.ExitCode);
+        }
     }
 }
--- a/src/shared/Core/ChildProcess.cs
+++ b/src/shared/Core/ChildProcess.cs
@@ -1,0 +1,53 @@
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Threading.Tasks;
+
+namespace GitCredentialManager;
+
+public class ChildProcess : DisposableObject
+{
+    private readonly ITrace2 _trace2;
+
+    private DateTimeOffset _startTime;
+    private DateTimeOffset _exitTime => Process.ExitTime;
+
+    private int _id => Process.Id;
+
+    public ProcessStartInfo StartInfo => Process.StartInfo;
+    public Process Process { get; }
+    public StreamWriter StandardInput => Process.StandardInput;
+    public StreamReader StandardOutput => Process.StandardOutput;
+    public StreamReader StandardError => Process.StandardError;
+    public int Id => Process.Id;
+    public int ExitCode => Process.ExitCode;
+
+    public static ChildProcess Start(ITrace2 trace2, ProcessStartInfo startInfo)
+    {
+        var childProc = new ChildProcess(trace2, startInfo);
+        childProc.Start();
+        return childProc;
+    }
+
+    public ChildProcess(ITrace2 trace2, ProcessStartInfo startInfo)
+    {
+        _trace2 = trace2;
+        Process = new Process() { StartInfo = startInfo };
+    }
+
+    public void Start()
+    {
+        ThrowIfDisposed();
+        Process.Start();
+    }
+
+    public void WaitForExit() => Process.WaitForExit();
+
+    public void Kill() => Process.Kill();
+
+    protected override void ReleaseManagedResources()
+    {
+        Process.Dispose();
+        base.ReleaseUnmanagedResources();
+    }
+}
--- a/src/shared/Core/Diagnostics/EnvironmentDiagnostic.cs
+++ b/src/shared/Core/Diagnostics/EnvironmentDiagnostic.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Net.Mime;
 using System.Text;
 using System.Threading.Tasks;
 
@@ -14,7 +15,7 @@ namespace GitCredentialManager.Diagnostics
 
         protected override Task<bool> RunInternalAsync(StringBuilder log, IList<string> additionalFiles)
         {
-            PlatformInformation platformInfo = PlatformUtils.GetPlatformInformation();
+            PlatformInformation platformInfo = PlatformUtils.GetPlatformInformation(CommandContext.Trace2);
             log.AppendLine($"OSType: {platformInfo.OperatingSystemType}");
             log.AppendLine($"OSVersion: {platformInfo.OperatingSystemVersion}");
 
--- a/src/shared/Core/Diagnostics/EnvironmentDiagnostic.cs
+++ b/src/shared/Core/Diagnostics/EnvironmentDiagnostic.cs
@@ -8,15 +8,9 @@ namespace GitCredentialManager.Diagnostics
 {
     public class EnvironmentDiagnostic : Diagnostic
     {
-        private readonly IEnvironment _env;
-
-        public EnvironmentDiagnostic(IEnvironment env)
-            : base("Environment")
-        {
-            EnsureArgument.NotNull(env, nameof(env));
-
-            _env = env;
-        }
+        public EnvironmentDiagnostic(ICommandContext commandContext)
+            : base("Environment", commandContext)
+        { }
 
         protected override Task<bool> RunInternalAsync(StringBuilder log, IList<string> additionalFiles)
         {
--- a/src/shared/Core/SidManager.cs
+++ b/src/shared/Core/SidManager.cs
@@ -1,0 +1,27 @@
+using System;
+
+namespace GitCredentialManager;
+
+public class SidManager
+{
+    private const string SidEnvar = "GIT_TRACE2_PARENT_SID";
+
+    public static string Sid { get; private set; }
+
+    public static void CreateSid()
+    {
+        Sid = Environment.GetEnvironmentVariable(SidEnvar);
+
+        if (!string.IsNullOrEmpty(Sid))
+        {
+            Sid = $"{Sid}/{Guid.NewGuid():D}";
+        }
+        else
+        {
+            // We are the root process; create our own 'root' SID
+            Sid = Guid.NewGuid().ToString("D");
+        }
+
+        Environment.SetEnvironmentVariable(SidEnvar, Sid);
+    }
+}
--- a/src/shared/Core/StringExtensions.cs
+++ b/src/shared/Core/StringExtensions.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Text;
 
 namespace GitCredentialManager
 {
@@ -239,6 +240,29 @@ namespace GitCredentialManager
         public static bool Contains(this string str, string value, StringComparison comparisonType)
         {
             return str?.IndexOf(value, comparisonType) >= 0;
+        }
+
+        /// <summary>
+        /// Convert string to snake case.
+        /// </summary>
+        /// <param name="str">String to convert.</param>
+        /// <returns>Input string converted to snake case.</returns>
+        public static string ToSnakeCase(this string str)
+        {
+            int len = str.Length;
+            var sb = new StringBuilder(2*len);
+            for (int i = 0; i < len; i++)
+            {
+                if (i > 0 && char.IsUpper(str[i]) &&
+                    (char.IsLower(str[i - 1]) || i < len - 1 && char.IsLower(str[i + 1])))
+                {
+                    sb.Append('_');
+
+                }
+                sb.Append(char.ToLower(str[i]));
+            }
+
+            return sb.ToString();
         }
     }
 }
--- a/src/shared/Core/Trace2.cs
+++ b/src/shared/Core/Trace2.cs
@@ -20,7 +20,20 @@ public enum Trace2Event
 {
     Version = 0,
     Start = 1,
-    Exit = 2
+    Exit = 2,
+    ChildStart = 3,
+    ChildExit = 4
+}
+
+/// <summary>
+/// Classifications of processes invoked by GCM.
+/// </summary>
+public enum Trace2ProcessClass
+{
+    None = 0,
+    UIHelper = 1,
+    Git = 2,
+    Other = 3
 }
 
 public class Trace2Settings
@@ -61,6 +74,45 @@ public interface ITrace2 : IDisposable
     void Stop(int exitCode,
         [System.Runtime.CompilerServices.CallerFilePath] string filePath = "",
         [System.Runtime.CompilerServices.CallerLineNumber] int lineNumber = 0);
+
+    /// <summary>
+    /// Writes information related to startup of child process to trace writer.
+    /// </summary>
+    /// <param name="startTime">Time at which child process began executing.</param>
+    /// <param name="processClass">Process classification.</param>
+    /// <param name="useShell">Specifies whether or not OS shell was used to start the process.</param>
+    /// <param name="appName">Name of application running in child process.</param>
+    /// <param name="argv">Arguments specific to the child process.</param>
+    /// <param name="sid">The child process's session id.</param>
+    /// <param name="filePath">Path of the file this method is called from.</param>
+    /// <param name="lineNumber">Line number of file this method is called from.</param>
+    void WriteChildStart(DateTimeOffset startTime,
+        Trace2ProcessClass processClass,
+        bool useShell,
+        string appName,
+        string argv,
+        [System.Runtime.CompilerServices.CallerFilePath]
+        string filePath = "",
+        [System.Runtime.CompilerServices.CallerLineNumber]
+        int lineNumber = 0);
+
+    /// <summary>
+    /// Writes information related to exit of child process to trace writer.
+    /// </summary>
+    /// <param name="elapsedTime">Runtime of child process.</param>
+    /// <param name="pid">Id of exiting process.</param>
+    /// <param name="code">Process exit code.</param>
+    /// <param name="sid">The child process's session id.</param>
+    /// <param name="filePath">Path of the file this method is called from.</param>
+    /// <param name="lineNumber">Line number of file this method is called from.</param>
+    void WriteChildExit(
+        double elapsedTime,
+        int pid,
+        int code,
+        [System.Runtime.CompilerServices.CallerFilePath]
+        string filePath = "",
+        [System.Runtime.CompilerServices.CallerLineNumber]
+        int lineNumber = 0);
 }
 
 public class Trace2 : DisposableObject, ITrace2
@@ -77,6 +129,8 @@ public class Trace2 : DisposableObject, ITrace2
     private string _sid;
 
     private bool _initialized;
+    // Increment with each new child process that is tracked
+    private int _childProcCounter = 0;
 
     public Trace2(ICommandContext commandContext)
     {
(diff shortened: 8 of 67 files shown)


ldennington:
> I'm not sure I understand what you're doing with the SidManager class. A first skim makes me think you're making it more complicated that it needs to be.

You were right - I went ahead and dropped that commit. The sid for a child process is now appended using the `SetSid` method, called from the `ChildProcess` class to ensure that it can be passed to each child process-related `Trace2` method.

We could also set the sid in `Trace2.cs`, but then we'd have to modify `Trace2`'s `_sid` variable both to append to the child process GUID and remove the appended value which felt a bit messy/dangerous.

src/shared/Core.Tests/StringExtensionsTests.cs
```diff
@@ -293,5 +293,14 @@ public void StringExtensions_TrimMiddle_String_ComparisonType(string input, stri
             string actual = StringExtensions.TrimMiddle(input, trim, comparisonType);
             Assert.Equal(expected, actual);
         }
+
+        [Theory]
+        [InlineData("FooBar", "foo_bar")]
+        [InlineData("FBBaz", "fb_baz")]
+        [InlineData("Foo", "foo")]
```
mjcheetham, line 300:
```suggestion
        [InlineData("FooBar", "foo_bar")]
        [InlineData("fooBar", "foo_bar")]
        [InlineData("FBBaz", "fb_baz")]
        [InlineData("Foo", "foo")]
        [InlineData("Fo", "fo")]
        [InlineData("fO", "f_o")]
        [InlineData("OO", "oo")]
        [InlineData("F", "f")]
        [InlineData("", "")]
```

Here are some more test data to use. One of them reveals a bug 🐛

src/shared/Core.Tests/StringExtensionsTests.cs
```diff
@@ -293,5 +293,14 @@ public void StringExtensions_TrimMiddle_String_ComparisonType(string input, stri
...
+        public void StringExtensions_ToSnakeCase_Converts_Correctly(string input, string expected)
+        {
+            Assert.Equal(input.ToSnakeCase(), expected);
```
mjcheetham, line 303: The actual and expected arguments are reversed.

```suggestion
            Assert.Equal(expected, input.ToSnakeCase());
```

src/shared/Core/StringExtensions.cs
```diff
@@ -240,5 +241,15 @@ public static bool Contains(this string str, string value, StringComparison comp
...
+
+        /// <summary>
+        /// Convert string to snake case.
+        /// </summary>
+        /// <param name="str">String to convert.</param>
+        /// <returns>Input string converted to snake case.</returns>
+        public static string ToSnakeCase(this string str) =>
+            string.Concat(str.Select((x, i) => (i > 0 && char.IsUpper(x) &&
+                (char.IsLower(str[i - 1]) || char.IsLower(str[i + 1])))
+                ? "_" + x : x.ToString())).ToLower();
```
mjcheetham, line 253: This was a little hard to read 😵‍💫

But I think there's also a bug here. Try using "OO" as an input.. we should get "oo" but we get an exception.

The problem is we are doing `str[i+1]` without checking that `i < str.Length - 1`.

mjcheetham: I wondered if we could use `StringBuilder` here and avoid LINQ to improve readability and improve perf.

I used BenchmarkDotnet to compare an (expanded out) LINQ version (with the `i < str.Length - 1` check) and a `StringBuilder + for-loop` version; I found the latter to be 42% of the compute and 22% of the memory of the former.

```csharp
[MemoryDiagnoser]
public class SnakeCaseBenchmark
{
    private const string str = "oFoObArFoObAzF";

    [Benchmark]
    public string Builder()
    {
        int len = str.Length;
        var sb = new StringBuilder(2*len);
        for (int i = 0; i < len; i++)
        {
            if (i > 0 && char.IsUpper(str[i]) &&
                (char.IsLower(str[i - 1]) || i < len - 1 && char.IsLower(str[i + 1])))
            {
                sb.Append('_');
            }

            sb.Append(char.ToLower(str[i]));
        }

        return sb.ToString();
    }

    [Benchmark(Baseline = true)]
    public string Linq()
    {
        int len = str.Length;
        var x = str.Select((x, i) =>
        {
            if (i > 0 && char.IsUpper(x) && (char.IsLower(str[i - 1]) || i < len - 1 && char.IsLower(str[i + 1])))
            {
                return "_" + x;
            }

            return x.ToString();
        });

        return string.Concat(x).ToLower();
    }
}
```


```text
BenchmarkDotNet=v0.13.5, OS=macOS Ventura 13.2.1 (22D68) [Darwin 22.3.0]
Apple M1 Max, 1 CPU, 10 logical and 10 physical cores
.NET SDK=8.0.100-preview.1.23115.2
  [Host]     : .NET 6.0.13 (6.0.1322.58009), Arm64 RyuJIT AdvSIMD
  DefaultJob : .NET 6.0.13 (6.0.1322.58009), Arm64 RyuJIT AdvSIMD


|  Method |     Mean |   Error |  StdDev | Ratio |   Gen0 | Allocated | Alloc Ratio |
|-------- |---------:|--------:|--------:|------:|-------:|----------:|------------:|
| Builder | 165.6 ns | 0.47 ns | 0.42 ns |  0.42 | 0.0918 |     192 B |        0.22 |
|    Linq | 390.6 ns | 0.78 ns | 0.73 ns |  1.00 | 0.4244 |     888 B |        1.00 |
```

src/shared/Core/Trace2.cs
```diff
@@ -164,7 +171,7 @@ internal bool TryGetPipeName(string eventTarget, out string name)
         return false;
     }
 
-    private void TryParseSettings(TextWriter error, IFileSystem fileSystem)
+    private void InitializeWriters()
```
mjcheetham, line 317: **Doesn't need to stop this PR**, but I wonder if in the future we could look to just having `Trace2`'s constructor take in the entire `ICommandContext`?

Then we can use the `IStandardStreams` and `IFileSystem` components here, given the 'no-op until `Initialize` is called' approach to resolving the dependency loop.

```patch
diff --git a/src/shared/Atlassian.Bitbucket.UI.Avalonia/Program.cs b/src/shared/Atlassian.Bitbucket.UI.Avalonia/Program.cs
index e2e231ac..03dcc322 100644
--- a/src/shared/Atlassian.Bitbucket.UI.Avalonia/Program.cs
+++ b/src/shared/Atlassian.Bitbucket.UI.Avalonia/Program.cs
@@ -49,7 +49,7 @@ namespace Atlassian.Bitbucket.UI
             using (var app = new HelperApplication(context))
             {
                 // Initialize TRACE2 system
-                context.Trace2.Initialize(context.Environment, context.Settings.GetTrace2Settings(), DateTimeOffset.UtcNow);
+                context.Trace2.Initialize(context.Settings.GetTrace2Settings(), DateTimeOffset.UtcNow);
 
                 // Write the start and version events
                 context.Trace2.Start(context.ApplicationPath, args);
diff --git a/src/shared/Core/CommandContext.cs b/src/shared/Core/CommandContext.cs
index 723d0f3a..21d72b69 100644
--- a/src/shared/Core/CommandContext.cs
+++ b/src/shared/Core/CommandContext.cs
@@ -92,7 +92,7 @@ namespace GitCredentialManager
 
             Streams = new StandardStreams();
             Trace   = new Trace();
-            Trace2  = new Trace2();
+            Trace2  = new Trace2(this);
 
             if (PlatformUtils.IsWindows())
             {
diff --git a/src/shared/Core/Trace2.cs b/src/shared/Core/Trace2.cs
index 62c75070..ae016e23 100644
--- a/src/shared/Core/Trace2.cs
+++ b/src/shared/Core/Trace2.cs
@@ -37,10 +37,9 @@ public interface ITrace2 : IDisposable
     /// <summary>
     /// Initialize TRACE2 tracing by initializing multi-use fields and setting up any configured target formats.
     /// </summary>
-    /// <param name="environment">The environment</param>
     /// <param name="settings">Enabled format targets and values.</param>
     /// <param name="startTime">Approximate time calling application began executing.</param>
-    void Initialize(IEnvironment environment, Trace2Settings settings, DateTimeOffset startTime);
+    void Initialize(Trace2Settings settings, DateTimeOffset startTime);
 
     /// <summary>
     /// Write Version and Start events.
@@ -67,6 +66,7 @@ public interface ITrace2 : IDisposable
 
 public class Trace2 : DisposableObject, ITrace2
 {
+    private readonly ICommandContext _context;
     private readonly object _writersLock = new object();
     private readonly Encoding _utf8NoBomEncoding = new UTF8Encoding(encoderShouldEmitUTF8Identifier: false);
     private readonly List<ITrace2Writer> _writers = new List<ITrace2Writer>();
@@ -79,7 +79,12 @@ public class Trace2 : DisposableObject, ITrace2
 
     private bool _initialized;
 
-    public void Initialize(IEnvironment environment, Trace2Settings settings, DateTimeOffset startTime)
+    public Trace2(ICommandContext context)
+    {
+        _context = context;
+    }
+
+    public void Initialize(Trace2Settings settings, DateTimeOffset startTime)
     {
         if (_initialized)
         {
@@ -88,7 +93,7 @@ public class Trace2 : DisposableObject, ITrace2
 
         _applicationStartTime = startTime;
         _settings = settings;
-        _sid = SetSid(environment);
+        _sid = SetSid(_context.Environment);
 
         InitializeWriters();
 
@@ -187,20 +192,20 @@ public class Trace2 : DisposableObject, ITrace2
             }
             else if (formatTarget.Value.IsTruthy()) // Write to stderr
             {
-                AddWriter(new Trace2StreamWriter(Console.Error, formatTarget.Key));
+                AddWriter(new Trace2StreamWriter(_context.Streams.Error, formatTarget.Key));
             }
             else if (Path.IsPathRooted(formatTarget.Value)) // Write to file
             {
                 try
                 {
-                    Stream stream = File.Open(formatTarget.Value, FileMode.Append,
+                    Stream stream = _context.FileSystem.OpenFileStream(formatTarget.Value, FileMode.Append,
                         FileAccess.Write, FileShare.ReadWrite);
                     AddWriter(new Trace2StreamWriter(new StreamWriter(stream, _utf8NoBomEncoding,
                         4096, leaveOpen: false), formatTarget.Key));
                 }
                 catch (Exception ex)
                 {
-                    Console.Error.WriteLine($"warning: unable to trace to file '{formatTarget.Value}': {ex.Message}");
+                    _context.Streams.Error.WriteLine($"warning: unable to trace to file '{formatTarget.Value}': {ex.Message}");
                 }
             }
         }
diff --git a/src/shared/Git-Credential-Manager.UI.Avalonia/Program.cs b/src/shared/Git-Credential-Manager.UI.Avalonia/Program.cs
index 31df67ff..ed96319a 100644
--- a/src/shared/Git-Credential-Manager.UI.Avalonia/Program.cs
+++ b/src/shared/Git-Credential-Manager.UI.Avalonia/Program.cs
@@ -47,7 +47,7 @@ namespace GitCredentialManager.UI
             using (var app = new HelperApplication(context))
             {
                 // Initialize TRACE2 system
-                context.Trace2.Initialize(context.Environment, context.Settings.GetTrace2Settings(), DateTimeOffset.UtcNow);
+                context.Trace2.Initialize(context.Settings.GetTrace2Settings(), DateTimeOffset.UtcNow);
 
                 // Write the start and version events
                 context.Trace2.Start(context.ApplicationPath, args);
diff --git a/src/shared/Git-Credential-Manager/Program.cs b/src/shared/Git-Credential-Manager/Program.cs
index 8c4dc554..829f187b 100644
--- a/src/shared/Git-Credential-Manager/Program.cs
+++ b/src/shared/Git-Credential-Manager/Program.cs
@@ -15,7 +15,7 @@ namespace GitCredentialManager
             {
                 // Initialize TRACE2 system
-                context.Trace2.Initialize(context.Environment, context.Settings.GetTrace2Settings(), DateTimeOffset.UtcNow);
+                context.Trace2.Initialize(context.Settings.GetTrace2Settings(), DateTimeOffset.UtcNow);
 
                 // Write the start and version events
                 context.Trace2.Start(context.ApplicationPath, args);
diff --git a/src/shared/GitHub.UI.Avalonia/Program.cs b/src/shared/GitHub.UI.Avalonia/Program.cs
index 105047c0..04ff038e 100644
--- a/src/shared/GitHub.UI.Avalonia/Program.cs
+++ b/src/shared/GitHub.UI.Avalonia/Program.cs
@@ -49,7 +49,7 @@ namespace GitHub.UI
             using (var app = new HelperApplication(context))
             {
                 // Initialize TRACE2 system
-                context.Trace2.Initialize(context.Environment, context.Settings.GetTrace2Settings(), DateTimeOffset.UtcNow);
+                context.Trace2.Initialize(context.Settings.GetTrace2Settings(), DateTimeOffset.UtcNow);
 
                 // Write the start and version events
                 context.Trace2.Start(context.ApplicationPath, args);
diff --git a/src/shared/GitLab.UI.Avalonia/Program.cs b/src/shared/GitLab.UI.Avalonia/Program.cs
index 3b096ef8..7768fffa 100644
--- a/src/shared/GitLab.UI.Avalonia/Program.cs
+++ b/src/shared/GitLab.UI.Avalonia/Program.cs
@@ -48,7 +48,7 @@ namespace GitLab.UI
             using (var app = new HelperApplication(context))
             {
                 // Initialize TRACE2 system
-                context.Trace2.Initialize(context.Environment, context.Settings.GetTrace2Settings(), DateTimeOffset.UtcNow);
+                context.Trace2.Initialize(context.Settings.GetTrace2Settings(), DateTimeOffset.UtcNow);
 
                 // Write the start and version events
                 context.Trace2.Start(context.ApplicationPath, args);
diff --git a/src/shared/TestInfrastructure/Objects/NullTrace.cs b/src/shared/TestInfrastructure/Objects/NullTrace.cs
index 5be2d497..9b0f995a 100644
--- a/src/shared/TestInfrastructure/Objects/NullTrace.cs
+++ b/src/shared/TestInfrastructure/Objects/NullTrace.cs
@@ -52,8 +52,7 @@ namespace GitCredentialManager.Tests.Objects
     {
         #region ITrace2
 
-        public void Initialize(IEnvironment environment,
-            Trace2Settings settings,
+        public void Initialize(Trace2Settings settings,
             DateTimeOffset startTime) { }
 
         public void Start(string appPath,

```

src/shared/Core.UI.Avalonia/Controls/AboutWindow.axaml.cs
```diff
@@ -27,7 +40,7 @@ private void ProjectButton_Click(object sender, RoutedEventArgs e)
             {
                 UseShellExecute = true
             };
-            Process.Start(psi);
+            _processManager.CreateProcess(psi);
```
mjcheetham, line 43: Hmm... do we need to wrap this `Process.Start`? It's only launching a browser and we're not waiting for it end.

Plus just creating a `new Trace2()` instance means any child_start events are not going to go anywhere.

src/shared/Core/Authentication/OAuth/OAuth2SystemWebBrowser.cs
```diff
@@ -34,14 +34,16 @@ public class OAuth2WebBrowserOptions
     public class OAuth2SystemWebBrowser : IOAuth2WebBrowser
     {
         private readonly IEnvironment _environment;
+        private readonly IProcessManager _processManager;
         private readonly OAuth2WebBrowserOptions _options;
 
-        public OAuth2SystemWebBrowser(IEnvironment environment, OAuth2WebBrowserOptions options)
+        public OAuth2SystemWebBrowser(IEnvironment environment, IProcessManager processManager, OAuth2WebBrowserOptions options)
```
mjcheetham, line 40: We're not using `IProcessManager` here?

src/shared/Core.Tests/WslUtilsTests.cs
```diff
@@ -100,15 +101,15 @@ public void WslUtils_CreateWslProcess()
...
-            Assert.True(process.StartInfo.RedirectStandardInput);
-            Assert.True(process.StartInfo.RedirectStandardOutput);
-            Assert.False(process.StartInfo.RedirectStandardError);
-            Assert.False(process.StartInfo.UseShellExecute);
+            Assert.Equal(expectedArgs, process.Process.StartInfo.Arguments);
+            Assert.Equal(expectedFileName, process.Process.StartInfo.FileName);
+            Assert.True(process.Process.StartInfo.RedirectStandardInput);
+            Assert.True(process.Process.StartInfo.RedirectStandardOutput);
+            Assert.False(process.Process.StartInfo.RedirectStandardError);
+            Assert.False(process.Process.StartInfo.UseShellExecute);
```
mjcheetham, line 112: I wonder if this is another property on `ChildProcess` we could create to avoid churn..

```csharp
public ProcessStartInfo StartInfo => Process.StartInfo;
```

src/shared/Core/Diagnostics/EnvironmentDiagnostic.cs
```diff
@@ -10,8 +11,8 @@ public class EnvironmentDiagnostic : Diagnostic
     {
         private readonly IEnvironment _env;
 
-        public EnvironmentDiagnostic(IEnvironment env)
-            : base("Environment")
+        public EnvironmentDiagnostic(IEnvironment env, ITrace2 trace2)
```
mjcheetham, line 14: Given that the various `Diagnostics` need many of the components from `ICommandContext`, maybe we should just always pass the entire `ICommandContext` to each one?

src/shared/Core/Interop/Posix/PosixProcessManager.cs
```diff
@@ -0,0 +1,7 @@
+namespace GitCredentialManager.Interop.Posix;
+
+public class PosixProcessManager : ProcessManagerBase
+{
+    public PosixProcessManager(ITrace2 trace2) : base(trace2)
+    { }
+}
```
mjcheetham, line 7: If this doesn't `override` anything from `ProcessManagerBase`, why not just drop `PosixProcessManager` and just use `ProcessManagerBase` directly instead.

mjcheetham: Maybe rename `ProcessManagerBase` to `ProcessManager` too.

src/shared/Core/Trace2.cs
```diff
@@ -20,7 +20,20 @@ public enum Trace2Event
...
+
+/// <summary>
+/// Classifications of processes invoked by GCM.
+/// </summary>
+public enum Trace2ProcessClass
+{
+    None = 0,
+    UIHelper = 1,
+    Git = 2,
+    OperatingSystem = 3
```
mjcheetham, line 36: I wonder if we just rename `OperatingSystem` to `General` or `Other`.

mjcheetham (commented):
Looking good so far! I have some comments initially after a first pass.

jeffhostetler:
> > I'm not sure I understand what you're doing with the SidManager class. A first skim makes me think you're making it more complicated that it needs to be.
> 
> You were right - I went ahead and dropped that commit. The sid for a child process is now appended using the `SetSid` method, called from the `ChildProcess` class to ensure that it can be passed to each child process-related `Trace2` method.
> 
> We could also set the sid in `Trace2.cs`, but then we'd have to modify `Trace2`'s `_sid` variable both to append to the child process GUID and remove the appended value which felt a bit messy/dangerous.

I think the solution is actually simpler than this.  All events generated by the GCM.exe process
should have the same SID on them.   The `child_start` and `child_exit` events are written from
the point of view of the invoking process (in this case GCM.exe) -- in a sense, GCM is saying that
**IT** is creating a child process (and may or may not wait for it to finish) and here are the args
that it is using.   This is the "outer span" on the lifetime of the child process.  That is, the
time from CreateProcess() thru WaitForSingleObject().  I digress, the point is that the SID needs
to be that of GCM.exe so that we can attach it to GCM's trace graph.

The child process itself should receive the inherited parent-SID and compute it's own and use that
for the child's events.  So when the child process does it's `version`, `start`, and `exit` events
with the child's new SID, we get the "internal span" on the lifetime of the child process.  That is,
the internal time between main() and exit()/return.  The child's data with the child's SID will be
used to create a new and independent trace graph for the child's lifetime.

The two trace graphs will be stitched together in the datastore after both have been converted
to OTLP (using OTel TraceIDs/SpanIDs synthesized from the two SIDs).  (And the outer and
inner lifetime spans will be adjacent, but distinct, in the combined graph.)

If the child process is not Trace2-aware, then there will not be a child trace graph for it,
so the parent trace graph needs to be self-sufficient.  That is, we'll only see the outer span.
Which is the case for Git hooks which are usually shell scripts.  But there's nothing we can
do about that.

Anyway, hope this helps.  And sorry for the confusion.

mjcheetham:
> The child process itself should receive the inherited parent-SID and compute it's own and use that
> for the child's events. So when the child process does it's `version`, `start`, and `exit` events
> with the child's new SID, we get the "internal span" on the lifetime of the child process. That is,
> the internal time between main() and exit()/return. The child's data with the child's SID will be
> used to create a new and independent trace graph for the child's lifetime.

Just to check my understanding here, this would look like this in pseudo GCM code?

```csharp
// Program.cs
static void Main(string[] args)
{
    using var context = new CommandContext();
    using var app = new Application(context);

    context.Trace2.Initialize(context.Environment, context.Settings.GetTrace2Settings(), DateTimeOffset.UtcNow);
    context.Trace2.Start(context.ApplicationPath, args);

    var exitCode = app.RunAsync(args)....;

    context.Trace2.Stop(exitCode);
}
```

```csharp
// CommandContext.cs
public CommandContext()
{
    ...
    Trace2 = new Trace2();
    ...
}
```

```csharp
// Trace2.cs
private string _sid;

public void Initialize(IEnvironment env, ...)
{
    ...
    _sid = ComputeSid(env);
    env.SetEnvironmentVariable("GIT_TRACE2_PARENT_SID", _sid);
}

private static string CreateSid(IEnvironment env)
{
    if (env.Variables.TryGet("GIT_TRACE2_PARENT_SID", out string parentSid))
    {
        // We are a child process; append to the parent to create our new SID
        return $"{parentSid}/{Guid.NewGuid():D}"
    }
    else
    {
        // We are the root process; create our own 'root' SID
        return Guid.NewGuid().ToString("D");
    }
}

public void Start(...)
{
    ...
    WriteVersion(...);
    WriteStart(...);
}

private void WriteVersion()
{
    WriteMessage(
        new VersionMessage(...)
        {
            Sid = _sid
        }
    );
}
```

jeffhostetler:
WRT the SID computation, yes.

I saw your note over the weekend and realized that perhaps the original confusion lies in my use of English when naming the event -- a noun-verb vs verb-noun issue, if you will.  Think of the "child-start" event as "i'm-starting-a-child" and likewise for "child-exit" is actually a "my-child-exited-with-code".   The parent emits those events and does so with it's SID.  If the child is Trace2-aware, it will create it's own SID using the append trick; if the child is not Trace2 aware, we won't get any telemetry from it.

Sorry for the confusion.  That naming scheme keep the event pairs symmetric "foo-start" and "foo-exit", "region-enter" and "region-leave" and etc.

mjcheetham:
> That naming scheme keep the event pairs symmetric "foo-start" and "foo-exit", "region-enter" and "region-leave" and etc.

Interesting question.. what should we do, if anything, about tracing child process starts that we don't subsequently wait for? A `child-start` without a `child-end`?

The example we have is clicking a link and having it open a web browser, for example.
I'd imagine we don't care about this, and every `-start` event should only be emitted when we can pair it with an `-end` event.

jeffhostetler:
I have a "child-ready" event for children that we start and release into the wild.  I use it for example when something like "git fsmonitor--daemon start" disassociates with the TTY and spawns a background process "git fsmonitor--daemon run".

The idea was that I wanted to be able to see that the child successfully got started (the exe was found and  we had 0755 access and etc) -- even if it immediately died because of a bad CL arg or something.

In the fsmonitor example, the test suite can use "... start" and know that that process will do whatever is necessary (and not return before) the actual "run" process gets going.  the test can then talk to the background "run" process over the socket/pipe/whatever -- and not have to worry about spinning until it is ready.  (and not have to rely on polling stdin/out or anything).

jeffhostetler:
So we should see either a "child-start" ... "child-exit"  -or-  "child-start" ... "child-ready" pair.  If the child cannot get started (bad exe or accesss or whatever), emit a "child-exit" with the usual exit code.

All of this is a bit hacky, but given the unix and windows process models, my hands were kinda tied...

jeffhostetler:
> private static string GetEventName(Trace2Event event)
{
    return event switch
    {
        Trace2Event.Version => "version",
        Trace2Event.ChildStart => "child_start",
        ...

Yeah, that's one way of doing it that is relatively easy to maintain and efficient.
I do question the need for the `Trace2Event` type at all, but that may just be style
and/or might help in the testing, so that's OK.

I'm just wondering whether we actually use the enum anywhere else?  I mean, in the following explicitly pass the event type to the initializer following the ctor.

```
   public void WriteChildStart(DateTimeOffset startTime,
        Trace2ProcessClass processClass,
        bool useShell,
        string appName,
        string argv,
        string sid,
        string filePath = "",
        int lineNumber = 0)
    {
        ...  
        WriteMessage(new ChildStartMessage()
        {
            Event = Trace2Event.ChildStart,
            Sid = sid,
            Time = startTime,
            File = Path.GetFileName(filePath).ToLower(),
            Line = lineNumber,
            Id = ++_childProcCounter,
            Classification = processClass,
            UseShell = useShell,
            Argv = procArgs
        });
        ...
```

The call to the `ChildStartMessage()` ctor already knows it is a child start event (and the set of required fields for that message is already encoded in the class definition).  The spelling of the event name could also be hidden inside that message class.  Passing in an event type to the ctor (or whatever we call the property bucket attached to the trivial ctor call), seems like an opportunity for error.

Is it possible to do something like (my C# is rusty):

```
public class ChildStartMessage : Trace2Message
{
    static const MyEventName = "child_start"
    public ChildStartMessage()
        : Trace2Message(MyEventName)
    {
    }
    ...
```

and let the rest of the initializer (following the new()) take care of the rest?

Just a thought and I could go either way here.

src/shared/Core/StringExtensions.cs
ldennington: Great analysis! I am more than happy to switch to Builder.

src/shared/Core.UI.Avalonia/Controls/AboutWindow.axaml.cs
ldennington: I kind of just wrapped everything - went ahead and removed this, though!

ldennington:
> ```
> public void WriteChildStart(...)
> {
>     WriteMessage(
>         new ChildStartMessage(...)
>         {
>             Sid = _sid // this is OUR sid, the child will compute its own
>         }
>     );
> }
> ```

@mjcheetham - since the child is computing its own Sid and CreateSid is private, do we need to start passing an `IEnvironment` to `ChildProcess`? Will it have its own `CreateSid` method? Just want to clarify before going down that path.

ldennington:
@mjcheetham, @jeffhostetler - I believe this latest series addresses all your feedback. A couple things to call out:

1. I decided to keep the snake casing patch. Although we don't necessarily need the `Trace2Event` enum (as Jeff pointed out), we _do_ (in my opinion) need the new `Trace2ProcessClass` to structure the information child process callers are passing in. Leaving the `Trace2Event` enum just follows this pattern for consistency.
2. I also kept the SidManager around, although it's now a static class that is invoked directly from `Program.cs` so that we immediately set the environment variable. This means that invoked Git processes will have the correct `Sid`, even if they're called before `TRACE2` is initialized. I also adjusted so that the start time is recorded just as the application starts.

src/shared/Core/SidManager.cs
```diff
@@ -0,0 +1,27 @@
...
+namespace GitCredentialManager;
+
+public class SidManager
+{
+    private const string SidEnvar = "GIT_TRACE2_PARENT_SID";
+
+    public static string Sid { get; private set; }
+
+    public static void CreateSid()
+    {
```
jeffhostetler, line 12: we talked about having a guard here to ensure that this function is only called once or that if it called a second time that it returns value from the first invocation.

src/shared/Core/Trace2FileWriter.cs
```diff
@@ -0,0 +1,18 @@
...
+    private readonly string _path;
+
+    public Trace2FileWriter(Trace2FormatTarget formatTarget, string path) : base(formatTarget)
+    {
+        _path = path;
+    }
+
+    public override void Write(Trace2Message message)
+    {
+        File.AppendAllText(_path, $"{Format(message)}\n");
```
jeffhostetler, line 16: File.AppendAllText() feels expensive, but I guess that you have to do it to get proper APPEND semantics.  Sigh.

If I understand your Trace2Writer.Format() method, it can fail -- set Failed=true and return "" -- do we need to guard against that here ?   Seems like if it does fail, we'll start writing blank lines to the file/socket.  (or maybe only one)

jeffhostetler: Would it be better to append the LF inside the Format() method?  Then we don't need to deal with it here and do a second $"{}" style printf.

src/shared/Core/Trace2FileWriter.cs
```diff
@@ -0,0 +1,18 @@
...
+    }
+}
```
jeffhostetler, line 18: do you folks care about final LFs on the source files?

src/shared/Atlassian.Bitbucket.UI.Avalonia/Program.cs
```diff
@@ -45,9 +45,15 @@ private static void AppMain(object o)
         {
```
jeffhostetler, line 45: I never use Avalonia, so I'm just going to ask a few questions.

Does the Main() and/or AppMain() function run in a separate process from the GCM proper process?

Does this new avalonia process need/want a new SID that is a child of the GCM SID?  I can't tell here
whether the New CommandContext() or context.Tract2.Initialize() generates or calls the SID Manager
class (like we do in the updated GCM Main()).

src/shared/Git-Credential-Manager.UI.Avalonia/Program.cs
```diff
@@ -43,9 +43,15 @@ private static void AppMain(object o)
         {
```
jeffhostetler, line 43: Same question here as in the BitBucket case.

src/shared/Core/ChildProcess.cs
```diff
@@ -19,26 +20,38 @@ public class ChildProcess : DisposableObject
...
-    public void Start()
+    public void Start(Trace2ProcessClass processClass)
     {
         ThrowIfDisposed();
+        // Record the time just before the process starts, since:
+        // (1) There is no event related to Start as there is with Exit.
+        // (2) Using Process.StartTime causes a race condition that leads
+        // to an exception if the process finishes executing before the
+        // variable is passed to Trace2.
+        _startTime = DateTimeOffset.UtcNow;
```
jeffhostetler, line 47: You could also say that we want to include the process startup cost in the timing.  
In Git, I write the child-start message before I spawn the child and the child-exit message after returning from wait3() or WaitForSingleObject(), so that we get the outer timing and all of the process overhead.

Your way is fine -- your child-start message will execute concurrently with the child startup, so it'll be slightly more efficient.  The important thing is using the clock value from before you called Process.Start().

src/shared/Core/Trace2.cs
```diff
@@ -20,7 +20,20 @@ public enum Trace2Event
...
+
+/// <summary>
+/// Classifications of processes invoked by GCM.
+/// </summary>
+public enum Trace2ProcessClass
+{
+    None = 0,
+    UIHelper = 1,
+    Git = 2,
+    Other = 3
```
jeffhostetler, line 36: Feel free to add other classes if it would be helpful.

jeffhostetler (approved):
Nicely done!  Thanks for putting up with me and all my comments.  😄

src/shared/Core/SidManager.cs
ldennington: Ah, I misunderstood that - I thought it was a redundant guard given we're checking whether it's empty when determining if we should append or just create a new GUID. But I see the purpose now and will add it back.

src/shared/Core/Trace2FileWriter.cs
ldennington: Matthew and I wrestled with this for awhile yesterday and came to the conclusion that `File.AppendAllText` is the only way to ensure we're writing correctly 😬.

We won't need to guard, since we ensure we don't call `Write` if the writer has failed [in `Trace2`](https://github.com/git-ecosystem/git-credential-manager/blob/622960273ecf31b2b23d2c93fc68462c7d6cc342/src/shared/Core/Trace2.cs#LL292-L295C18).

I did go ahead and update Format to append the newline - thanks for the pointer!

src/shared/Core/Trace2FileWriter.cs
ldennington: Er, we're not as strict as Git, but we prefer it. Thanks for pointing it out, I've got it fixed locally!

src/shared/Atlassian.Bitbucket.UI.Avalonia/Program.cs
ldennington: I'm fine with giving them child Sids. It will mean removing the guard we spoke about previously, but that's not a big deal.

src/shared/Core/ChildProcess.cs
ldennington: Going to leave as is for now to lessen the (admittedly already significant) churn in this PR.

src/shared/Core/SidManager.cs
ldennington: Did not end up adding per the helper exe conversation below.

src/shared/Core.UI.Avalonia/Controls/AboutWindow.axaml.cs
```diff
@@ -3,6 +3,8 @@
 using Avalonia.Controls;
 using Avalonia.Interactivity;
 using Avalonia.Markup.Xaml;
+using GitCredentialManager.Interop.Posix;
+using GitCredentialManager.Interop.Windows;
```
mjcheetham, line 7: Don't need this?

src/shared/Core/ChildProcess.cs
```diff
@@ -0,0 +1,79 @@
...
+    {
+        ThrowIfDisposed();
+        // Record the time just before the process starts, since:
+        // (1) There is no event related to Start as there is with Exit.
+        // (2) Using Process.StartTime causes a race condition that leads
+        // to an exception if the process finishes executing before the
+        // variable is passed to Trace2.
+        _startTime = DateTimeOffset.UtcNow;
+        Process.Start();
+        _trace2.WriteChildStart(
```
mjcheetham, line 49: Feels like we should still switch these around, in case the process (say, Git) starts up and writes to the same Trace2 target before we write 'child start'?

mjcheetham (approved):
Looks great! Happy to merge as is.

src/shared/Core.UI.Avalonia/Controls/AboutWindow.axaml.cs
ldennington: Good catch! Went ahead and fixed.

src/shared/Core/ChildProcess.cs
ldennington: Sure, I went ahead and made the move, then re-ran my tests. All looks good.

madgbec (commented):
I started and exited device unable to process or click and then I was unsure of action thought should have been more automated

madgbec (commented):
I started and exited device unable to process or click and then I was unsure of action thought should have been more automated

PR #1136, opened 2023-02-28
git: consistently read from stdout before exit wait

mjcheetham:
We should always try and drain standard output before waiting for the Git process to exit. We do this for all other calls that expect output, but missed the `TryGet` method.

This is important if Git ever writes out so much data to stdout that we buffer the output, and then Git will not exit until it can finish writing (and we'd be waiting for Git to exit).


ldennington (approved):

derrickstolee (approved):

FRANKAAAA67 (commented):
Excited

PR #1137, opened 2023-03-01
Add login hint support for GitHub browser-based OAuth authentication

mjcheetham:
If a username is specified in the remote URL then include this as a login hint when performing OAuth authentication for GitHub. This can help users of multiple accounts select the correct account before the OAuth flow completes and returns a token for the wrong account.

GitHub shows this nice prompt when a login hint is provided that does _not match the currently logged-in user_:

image

At the same time, if there is no logged-in user, then the login page's username box is already filled in:

image

The login hint is specified to GitHub via the extra `login` query parameter to the authorization endpoint.


ldennington (approved):
:shipit:

PR #1146, opened 2023-03-11
Ensure we create a `WindowsProcessManager` on Windows

mjcheetham:
We are creating a `WindowsProcessManager` implementation on macOS and a non-Windows/WSL aware, plain `ProcessManager` on Windows - this is the reverse of what we should do. Fix this.


ldennington (approved):
😱 - thanks for the quick fix and added protection moving forward.

PR #1148, opened 2023-03-11
Better browser/interactive session detection inside of WSL

### LINKED ISSUE #878: for wslview: Browser authentication disabled if DISPLAY variable is not set
wslview is supported
see here:
https://github.com/GitCredentialManager/git-credential-manager/blob/7c4381cb754538367a7d0b33ecf3550ca6b5fa79/src/shared/Core/BrowserUtils.cs#L43

wslview does not require a DISPLAY variable.

However, the posix session manager checks for the DISPLAY variable.
https://github.com/GitCredentialManager/git-credential-manager/blob/7c4381cb754538367a7d0b33ecf3550ca6b5fa79/src/shared/Core/Interop/Posix/PosixSessionManager.cs#L13

Therefore, the browser is disabled.
https://github.com/GitCredentialManager/git-credential-manager/blob/dea7cb1afdab43f3852dd897c58b52ae3b6722a3/src/shared/GitHub/GitHubAuthentication.cs#L70

Possibly broken in this commit.
https://github.com/GitCredentialManager/git-credential-manager/commit/28f487ad0931010c657dd919f72c87e52d402e5c

Suggestion:
Bypass the check if wslview is present.

workaround:
Set the display variable.

From a terminal, run `git credential-manager-core --version` and paste the output.
2.0.785.29792

* [ ] Azure DevOps
* [X] GitHub
* [ ] GitHub Enterprise
* [ ] Bitbucket
* [ ] Other - please describe

* [ ] Yes
* [X] No, for a different reason - please describe

Any private repo will work.

---

**Expected behavior**

Browser pops up when authenticating.

**Actual behavior**

Only show oauth device code login.

**Logs**

N/A

mjcheetham:
Today we're gating browser-based authentication mechanisms behind a GUI-possible session, which isn't 100% correct. When inside of a WSL session, without WSLg enabled, it is still possible to launch a web browser in the hosting Windows session.

Split the determination between "GUI session" and "browser possible", and add the special case for WSL inside an interactive Windows session.

Things are complicated here because it is possible to SSH in to the WSL instance from outside of the Windows host, meaning the Windows session also doesn't have UI. This case we still need to prevent browser-based options.

In order to detect these cases we need to 'punch out' to the Windows host and run a script/cmdlet to determine the session type.

Fixes #878

### DIFF
--- a/src/shared/Core/Interop/Linux/LinuxSessionManager.cs
+++ b/src/shared/Core/Interop/Linux/LinuxSessionManager.cs
@@ -1,0 +1,64 @@
+using GitCredentialManager.Interop.Posix;
+
+namespace GitCredentialManager.Interop.Linux;
+
+public class LinuxSessionManager : PosixSessionManager
+{
+    private bool? _isWebBrowserAvailable;
+
+    public LinuxSessionManager(IEnvironment env, IFileSystem fs) : base(env, fs)
+    {
+        PlatformUtils.EnsureLinux();
+    }
+    
+    public override bool IsWebBrowserAvailable
+    {
+        get
+        {
+            return _isWebBrowserAvailable ??= GetWebBrowserAvailable();
+        }
+    }
+    
+    private bool GetWebBrowserAvailable()
+    {
+        // If this is a Windows Subsystem for Linux distribution we may
+        // be able to launch the web browser of the host Windows OS.
+        if (WslUtils.IsWslDistribution(Environment, FileSystem, out _))
+        {
+            // We need a shell execute handler to be able to launch to browser
+            if (!BrowserUtils.TryGetLinuxShellExecuteHandler(Environment, out _))
+            {
+                return false;
+            }
+
+            //
+            // If we are in Windows logon session 0 then the user can never interact,
+            // even in the WinSta0 window station. This is typical when SSH-ing into a
+            // Windows 10+ machine using the default OpenSSH Server configuration,
+            // which runs in the 'services' session 0.
+            //
+            // If we're in any other session, and in the WinSta0 window station then
+            // the user can possibly interact. However, since it's hard to determine
+            // the window station from PowerShell cmdlets (we'd need to write P/Invoke
+            // code and that's just messy and too many levels of indirection quite
+            // frankly!) we just assume any non session 0 is interactive.
+            //
+            // This assumption doesn't hold true if the user has changed the user that
+            // the OpenSSH Server service runs as (not a built-in NT service) *AND*
+            // they've SSH-ed into the Windows host (and then started a WSL shell).
+            // This feels like a very small subset of users...
+            //
+            if (WslUtils.GetWindowsSessionId(FileSystem) == 0)
+            {
+                return false;
+            }
+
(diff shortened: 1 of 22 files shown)


src/shared/Core/Interop/Linux/LinuxSessionManager.cs
```diff
@@ -0,0 +1,64 @@
...
+        // be able to launch the web browser of the host Windows OS.
+        if (WslUtils.IsWslDistribution(Environment, FileSystem, out _))
+        {
+            // We need a shell execute handler to be able to launch to browser
+            if (!BrowserUtils.TryGetLinuxShellExecuteHandler(Environment, out _))
+            {
+                return false;
+            }
+
+            //
```
ldennington, line 34: Out of curiosity, when should one add `// <empty line>` before a comment versus `// <comment>`?

src/shared/Core/Interop/Posix/PosixSessionManager.cs
```diff
@@ -1,17 +1,15 @@
...
         {
             PlatformUtils.EnsurePosix();
         }
 
         // Check if we have an X11 or Wayland display environment available
         public override bool IsDesktopSession =>
-            !string.IsNullOrWhiteSpace(Environment.GetEnvironmentVariable("DISPLAY")) ||
-            !string.IsNullOrWhiteSpace(Environment.GetEnvironmentVariable("WAYLAND_DISPLAY"));
+            !string.IsNullOrWhiteSpace(System.Environment.GetEnvironmentVariable("DISPLAY")) ||
+            !string.IsNullOrWhiteSpace(System.Environment.GetEnvironmentVariable("WAYLAND_DISPLAY"));
```
ldennington, line 13: Why the addition of `System` here and removal of the `using`?

ldennington (approved):
Looks good! Great job on the `INI` file deserialization especially.

src/shared/Core/Interop/Posix/PosixSessionManager.cs
mjcheetham: Ahh.. good point. It's using `System.Environment` because we were using the BCL `Environment` class to read these variables initally.

Now we've added an `IEnvironment` instance to the `SessionManager`, there's a property named `Environment` (same name), so the namespace prefix is required to still reference the BCL static class.

But.. we should probably be using the `IEnvironment` here now that we have it.

mjcheetham: Pushed a new version, with a new commit at the front of the series that refactors the `IEnvironment` hierarchy a bit to remove a code smell.

src/shared/Core/Interop/Linux/LinuxSessionManager.cs
mjcheetham: It's more personal taste. When a comment block ends up feeling like a novel with multiple paragraphs, I think it's easier to read (and stands out more as something important to understand!) with surrounding whitespace.

So let's say "a comment block with multiple paragraphs should have a `//` blank line at start and end"?

PR #1149, opened 2023-03-13
settings: support scoped enterprise settings

### LINKED ISSUE #983: Support per-url enterprise configuration
### Feature description

[Enterprise configuration](https://github.com/GitCredentialManager/git-credential-manager/blob/c7c35983b83d7eeee6c179b22a9d1650d562a8a9/docs/enterprise-config.md) enables setting credentials using the windows registry for GCM.

[GitLab support](https://github.com/GitCredentialManager/git-credential-manager/blob/c7c35983b83d7eeee6c179b22a9d1650d562a8a9/docs/gitlab.md) has settings which are per-url something like: `credential.https://url.something`.

Currently, enterprise configuration only supports settings in the form `credential.something`, not `credential.https://url.something`, which means for example, multiple gitlab instances cannot be configured at the enterprise level.

It looks like support for this is implemented here: https://github.com/GitCredentialManager/git-credential-manager/blob/9ecda7b84778f89de6356c9b3fa47a7587c5ab5f/src/shared/Core/Settings.cs#L420-L424 - it probably just needs a similar block slightly further up in that method.

mjcheetham:
Add support for remote URL-scoped enterprise default settings.

Fixes #983

### DIFF
--- a/src/shared/Core.Tests/SettingsTests.cs
+++ b/src/shared/Core.Tests/SettingsTests.cs
@@ -1,5 +1,4 @@
 using System;
-using System.Collections.Generic;
 using System.Linq;
 using System.Net;
 using GitCredentialManager.Tests.Objects;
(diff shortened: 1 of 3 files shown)


ldennington (approved):
🚀

PR #1151, opened 2023-03-14
trace2: add performance format target

ldennington:
This series adds the performance format target for TRACE2 tracing.

The first two patches are small fixes for bugs I noticed as I was testing the
performance format target. The first moves the `FormatTargetTypes` enum
from the specific `Trace2StreamWriter` class to the shared `Trace2Writer`
class. The second ensures the normal format event names are written in snake
case (rather than lowercase).

The third patch adds the ability to get the "depth" of the process. Because Sids
and depth are process-related concepts, it removes the SidManager class and
moved depth/sid-related properties and methods into the ProcessManager class.

The fourth patch adds the performance format target. It includes a few notable
components:
    
1. A new GetMessage method to ensure message strings written to normal and
performance targets are not duplicated.
2. A new PerformanceFormatComponent class to track sizing of optional
properties of performance format messages.
3. A BuildSpan method that Trace2Message children use to create
correctly-sized "spans" for optional properties of performance format
messages. A span is a piece of the message beginning with a pipe (|) and
ending just before the next pipe or the end of the message.
4. A BuildTimeSpan method that adjusts spans for long/short times. In
future, when we have events that include performance format string span
components, we will need to add an additional method to handle them.


src/shared/Core/Trace2.cs
```diff
@@ -286,6 +301,30 @@ internal static bool TryGetPipeName(string eventTarget, out string name)
...
+
+        int count = 0;
+        // Use AsSpan() for slight performance bump over traditional foreach loop.
+        foreach (var c in _sid.AsSpan())
+        {
+            if (c == processSeparator)
+                count++;
+        }
+
+        return count;
```
mjcheetham, line 325: Could we cache this value, given `_sid` is set once per process? Would avoid rescanning O(N) the `_sid` string each time we called `GetProcessDepth()`.

We could also pre-compute this depth when we do `CreateSid()` too!

src/shared/Core/Trace2.cs
```diff
@@ -449,34 +499,138 @@ public abstract class Trace2Message
...
+        //     BAR SP [<category>] SP+
+        //     BAR SP DOTS* <perf-event-message>
+        //     LF
+        return $"{time} {source,-29}| d{Depth} | {Thread,-24} | {Event.ToString().ToSnakeCase(),-12} {sb} | {message}";
+    }
+
+    internal string BuildSpan(PerformanceFormatComponent component)
+    {
+        var sb = new StringBuilder();
+        var property = GetType().GetProperty(component.Name);
```
mjcheetham, line 551: Could we avoid the use of reflection here? Could we possibly add `string BuildPerformanceSpan()` or something to `Trace2Message` that returns the appropriately built string?

src/shared/Core/Trace2.cs
```diff
@@ -449,34 +499,138 @@ public abstract class Trace2Message
...
+
+            if (double.TryParse(propertyValue, out var value))
+            {
+                // Ensure the correct number of decimal places
+                propertyValue = value.ToString("F6");
+                isTime = true;
+            }
+
+            AdjustPadding(component, propertyValue, isTime, dataSize);
+            sb.Append($"|{new string(' ', component.BeginPadding)}{propertyValue}");
```
mjcheetham, line 567: If we have a `StringBuilder` could we avoid mixing `new string(..)` and `$"interpolated {strings}"` and just use the various `Append*` methods instead? Avoid creating lots of new `string`s for each format operation.

src/shared/Core/Trace2.cs
```diff
@@ -491,16 +645,26 @@ public override string ToJson()
...
     public override string ToNormalString()
     {
-        return BuildNormalString(Exe.ToLower());
+        return BuildNormalString(GetEventMessage());
+    }
+
+    public override string ToPerformanceString()
+    {
+        return BuildPerformanceString(GetEventMessage());
+    }
```
mjcheetham, line 654: I think we can simplify `Trace2Message` here by pulling `string GetEventMessage()` up to the abstract `Trace2Message` class, and then call `GetEventMessage()` inside of the `Build*String()` methods.

```patch
diff --git a/src/shared/Core.Tests/Trace2Tests.cs b/src/shared/Core.Tests/Trace2Tests.cs
index 9d6fed32..e6abc0ed 100644
--- a/src/shared/Core.Tests/Trace2Tests.cs
+++ b/src/shared/Core.Tests/Trace2Tests.cs
@@ -17,17 +17,12 @@ internal class TestTrace2Message : Trace2Message
 
     public override string ToJson()
     {
-        return "";
+        return string.Empty;
     }
 
-    public override string ToNormalString()
+    protected override string GetEventMessage()
     {
-        return "";
-    }
-
-    public override string ToPerformanceString()
-    {
-        return "";
+        return string.Empty;
     }
 }
 
diff --git a/src/shared/Core/Trace2.cs b/src/shared/Core/Trace2.cs
index 95d4a60b..ce2ba7f2 100644
--- a/src/shared/Core/Trace2.cs
+++ b/src/shared/Core/Trace2.cs
@@ -493,7 +493,6 @@ public abstract class Trace2Message
     public DateTimeOffset Time { get; set; }
 
     [JsonProperty("file", Order = 5)]
-
     public string File { get; set; }
 
     [JsonProperty("line", Order = 6)]
@@ -504,12 +503,10 @@ public abstract class Trace2Message
 
     public abstract string ToJson();
 
-    public abstract string ToNormalString();
-
-    public abstract string ToPerformanceString();
-
-    protected string BuildNormalString(string message)
+    public string ToNormalString()
     {
+        string message = GetEventMessage();
+
         // The normal format uses local time rather than UTC time.
         string time = Time.ToLocalTime().ToString(NormalPerfTimeFormat);
         string source = GetSource();
@@ -519,8 +516,10 @@ public abstract class Trace2Message
         return $"{time} {source,-33} {Event.ToString().ToSnakeCase()} {message}";
     }
 
-    protected string BuildPerformanceString(string message)
+    public string ToPerformanceString()
     {
+        string message = GetEventMessage();
+
         // The performance format uses local time rather than UTC time.
         var time = Time.ToLocalTime().ToString(NormalPerfTimeFormat);
         var source = GetSource();
@@ -544,6 +543,8 @@ public abstract class Trace2Message
         //     LF
         return $"{time} {source,-29}| d{Depth} | {Thread,-24} | {Event.ToString().ToSnakeCase(),-12} {sb} | {message}";
     }
+    
+    protected abstract string GetEventMessage();
 
     internal string BuildSpan(PerformanceFormatComponent component)
     {
@@ -643,17 +644,7 @@ public class VersionMessage : Trace2Message
             });
     }
 
-    public override string ToNormalString()
-    {
-        return BuildNormalString(GetEventMessage());
-    }
-
-    public override string ToPerformanceString()
-    {
-        return BuildPerformanceString(GetEventMessage());
-    }
-
-    private string GetEventMessage()
+    protected override string GetEventMessage()
     {
         return Exe.ToLower();
     }
@@ -677,17 +668,7 @@ public class StartMessage : Trace2Message
             });
     }
 
-    public override string ToNormalString()
-    {
-        return BuildNormalString(GetEventMessage());
-    }
-
-    public override string ToPerformanceString()
-    {
-        return BuildPerformanceString(GetEventMessage());
-    }
-
-    private string GetEventMessage()
+    protected override string GetEventMessage()
     {
         return string.Join(" ", Argv);
     }
@@ -711,17 +692,7 @@ public class ExitMessage : Trace2Message
             });
     }
 
-    public override string ToNormalString()
-    {
-        return BuildNormalString(GetEventMessage());
-    }
-
-    public override string ToPerformanceString()
-    {
-        return BuildPerformanceString(GetEventMessage());
-    }
-
-    private string GetEventMessage()
+    protected override string GetEventMessage()
     {
         return $"elapsed:{ElapsedTime} code:{Code}";
     }
@@ -751,21 +722,11 @@ public class ChildStartMessage : Trace2Message
             });
     }
 
-    public override string ToNormalString()
-    {
-        return BuildNormalString(GetEventMessage(Id.ToString()));
-    }
-
-    public override string ToPerformanceString()
-    {
-        return BuildPerformanceString(GetEventMessage($"ch{Id}"));
-    }
-
-    private string GetEventMessage(string id)
+    protected override string GetEventMessage()
     {
         var sb = new StringBuilder();
         AddDots(sb);
-        sb.Append($"[{id}] {string.Join(" ", Argv)}");
+        sb.Append($"[{Id}] {string.Join(" ", Argv)}");
         return sb.ToString();
     }
 }
@@ -794,21 +755,11 @@ public class ChildExitMessage : Trace2Message
             });
     }
 
-    public override string ToNormalString()
-    {
-        return BuildNormalString(GetEventMessage(Id.ToString()));
-    }
-
-    public override string ToPerformanceString()
-    {
-        return BuildPerformanceString(GetEventMessage($"ch{Id}"));
-    }
-
-    private string GetEventMessage(string id)
+    protected override string GetEventMessage()
     {
         var sb = new StringBuilder();
         AddDots(sb);
-        sb.Append($"[{id}] pid:{Pid} code:{Code} elapsed:{ElapsedTime}");
+        sb.Append($"[{Id}] pid:{Pid} code:{Code} elapsed:{ElapsedTime}");
         return sb.ToString();
     }
 }

```

mjcheetham (changes requested):
Had a few suggestions on improving perf/caching values, simplifying the virtual methods on `Trace2Message`, and wondered if we could avoid reflection at all.

src/shared/Core.Tests/Trace2Tests.cs
```diff
@@ -30,4 +56,112 @@ public void TryGetPipeName_Windows_Returns_Expected_Value(string input, string e
...
+        };
+
+        var trace2Message = new TestTrace2Message()
+        {
+            ElapsedTime = input
+        };
+
+        var actual = trace2Message.BuildSpan(component);
+
+        Assert.Equal(expected, actual);
```
jeffhostetler, line 116: Exact matching the string formatting of a float/double is iffy because of round-off and other machine precision issues.  This seems to work at this precision level, but would probably fail to give the right answer if you tried "10000....000.123456" because the mantissa doesn't have enough resolution.  I'm not saying we should change this, but just be aware that testing floats is tricky.

Mabye run the result thru sscanf("|%g ",&v) and see it if is within an epsilon of the input value.  But this may be overkill.

src/shared/Core/Trace2.cs
ldennington: I like this idea, thanks!

src/shared/Core/Trace2.cs
ldennington: How in this case do we add the `ch` prefix to the process id in child start and exit event messages? For example, the last line of the patch will output `[1]` when the correct formatting for the performance format target is `ch[1]`.

mjcheetham: Can we have `GetEventMessage(..)` take a `Trace2FormatTarget`, and all other implementations ignore the arg, but this particular event adds the `ch` in front of the `Id` if `formatTarget == Trace2FormatTarget.Performance`?

```diff
-    private string GetEventMessage(string id)
+    protected override string GetEventMessage(Trace2FormatTarget formatTarget)
     {
         var sb = new StringBuilder();
         AddDots(sb);
-        sb.Append($"[{id}] pid:{Pid} code:{Code} elapsed:{ElapsedTime}");
+        if (formatTarget == Trace2FormatTarget.Performance)
+            sb.Append($"[ch{Id}]");
+        else
+            sb.Append($"[{Id}]");
+        sb.Append($" pid:{Pid} code:{Code} elapsed:{ElapsedTime}");
         return sb.ToString();
     }
 }
```

mjcheetham: Updated patch:

```patch
diff --git a/src/shared/Core.Tests/Trace2Tests.cs b/src/shared/Core.Tests/Trace2Tests.cs
index 9d6fed32..7b3b15b0 100644
--- a/src/shared/Core.Tests/Trace2Tests.cs
+++ b/src/shared/Core.Tests/Trace2Tests.cs
@@ -17,17 +17,12 @@ internal class TestTrace2Message : Trace2Message
 
     public override string ToJson()
     {
-        return "";
+        return string.Empty;
     }
 
-    public override string ToNormalString()
+    protected override string GetEventMessage(Trace2FormatTarget formatTarget)
     {
-        return "";
-    }
-
-    public override string ToPerformanceString()
-    {
-        return "";
+        return string.Empty;
     }
 }
 
diff --git a/src/shared/Core/Trace2.cs b/src/shared/Core/Trace2.cs
index 95d4a60b..31367ee9 100644
--- a/src/shared/Core/Trace2.cs
+++ b/src/shared/Core/Trace2.cs
@@ -493,7 +493,6 @@ public abstract class Trace2Message
     public DateTimeOffset Time { get; set; }
 
     [JsonProperty("file", Order = 5)]
-
     public string File { get; set; }
 
     [JsonProperty("line", Order = 6)]
@@ -504,12 +503,10 @@ public abstract class Trace2Message
 
     public abstract string ToJson();
 
-    public abstract string ToNormalString();
-
-    public abstract string ToPerformanceString();
-
-    protected string BuildNormalString(string message)
+    public string ToNormalString()
     {
+        string message = GetEventMessage(Trace2FormatTarget.Normal);
+
         // The normal format uses local time rather than UTC time.
         string time = Time.ToLocalTime().ToString(NormalPerfTimeFormat);
         string source = GetSource();
@@ -519,8 +516,10 @@ public abstract class Trace2Message
         return $"{time} {source,-33} {Event.ToString().ToSnakeCase()} {message}";
     }
 
-    protected string BuildPerformanceString(string message)
+    public string ToPerformanceString()
     {
+        string message = GetEventMessage(Trace2FormatTarget.Performance);
+
         // The performance format uses local time rather than UTC time.
         var time = Time.ToLocalTime().ToString(NormalPerfTimeFormat);
         var source = GetSource();
@@ -545,6 +544,8 @@ public abstract class Trace2Message
         return $"{time} {source,-29}| d{Depth} | {Thread,-24} | {Event.ToString().ToSnakeCase(),-12} {sb} | {message}";
     }
 
+    protected abstract string GetEventMessage(Trace2FormatTarget formatTarget);
+
     internal string BuildSpan(PerformanceFormatComponent component)
     {
         var sb = new StringBuilder();
@@ -643,17 +644,7 @@ public class VersionMessage : Trace2Message
             });
     }
 
-    public override string ToNormalString()
-    {
-        return BuildNormalString(GetEventMessage());
-    }
-
-    public override string ToPerformanceString()
-    {
-        return BuildPerformanceString(GetEventMessage());
-    }
-
-    private string GetEventMessage()
+    protected override string GetEventMessage(Trace2FormatTarget formatTarget)
     {
         return Exe.ToLower();
     }
@@ -677,17 +668,7 @@ public class StartMessage : Trace2Message
             });
     }
 
-    public override string ToNormalString()
-    {
-        return BuildNormalString(GetEventMessage());
-    }
-
-    public override string ToPerformanceString()
-    {
-        return BuildPerformanceString(GetEventMessage());
-    }
-
-    private string GetEventMessage()
+    protected override string GetEventMessage(Trace2FormatTarget formatTarget)
     {
         return string.Join(" ", Argv);
     }
@@ -711,17 +692,7 @@ public class ExitMessage : Trace2Message
             });
     }
 
-    public override string ToNormalString()
-    {
-        return BuildNormalString(GetEventMessage());
-    }
-
-    public override string ToPerformanceString()
-    {
-        return BuildPerformanceString(GetEventMessage());
-    }
-
-    private string GetEventMessage()
+    protected override string GetEventMessage(Trace2FormatTarget formatTarget)
     {
         return $"elapsed:{ElapsedTime} code:{Code}";
     }
@@ -751,21 +722,18 @@ public class ChildStartMessage : Trace2Message
             });
     }
 
-    public override string ToNormalString()
-    {
-        return BuildNormalString(GetEventMessage(Id.ToString()));
-    }
-
-    public override string ToPerformanceString()
-    {
-        return BuildPerformanceString(GetEventMessage($"ch{Id}"));
-    }
-
-    private string GetEventMessage(string id)
+    protected override string GetEventMessage(Trace2FormatTarget formatTarget)
     {
         var sb = new StringBuilder();
         AddDots(sb);
-        sb.Append($"[{id}] {string.Join(" ", Argv)}");
+
+        if (formatTarget == Trace2FormatTarget.Performance)
+            sb.Append($"[ch{Id}]");
+        else
+            sb.Append($"[{Id}]");
+
+        sb.Append($" {string.Join(" ", Argv)}");
+        
         return sb.ToString();
     }
 }
@@ -794,21 +762,18 @@ public class ChildExitMessage : Trace2Message
             });
     }
 
-    public override string ToNormalString()
-    {
-        return BuildNormalString(GetEventMessage(Id.ToString()));
-    }
-
-    public override string ToPerformanceString()
-    {
-        return BuildPerformanceString(GetEventMessage($"ch{Id}"));
-    }
-
-    private string GetEventMessage(string id)
+    protected override string GetEventMessage(Trace2FormatTarget formatTarget)
     {
         var sb = new StringBuilder();
         AddDots(sb);
-        sb.Append($"[{id}] pid:{Pid} code:{Code} elapsed:{ElapsedTime}");
+        
+        if (formatTarget == Trace2FormatTarget.Performance)
+            sb.Append($"[ch{Id}]");
+        else
+            sb.Append($"[{Id}]");
+        
+        sb.Append($" pid:{Pid} code:{Code} elapsed:{ElapsedTime}");
+        
         return sb.ToString();
     }
 }

```

src/shared/Core.Tests/Trace2Tests.cs
mjcheetham: With Xunit we can use the `Assert.Equals(double, double, int)` overload to specify a precision for comparing the floating point numbers:

image

src/shared/Core/Trace2.cs
ldennington: Now pre-computing with `CreateSid`. Thanks for the idea!

src/shared/Core/Trace2.cs
ldennington: I think this comment was rendered inapplicable with all the other refactorings. Let me know if I'm mistaken though.

src/shared/Core.Tests/Trace2Tests.cs
ldennington: Going to leave as-is for now. Point is completely valid, but I feel the test is useful as it stands for checking the alignment of values we'll see in the majority of cases.

ldennington:
@mjcheetham @jeffhostetler - I believe all feedback has been addressed with my latest push, but please do let me know if I missed anything.

src/shared/Core.Tests/ProcessManagerTests.cs
```diff
@@ -0,0 +1,20 @@
...
+    [InlineData("foo/bar", 1)]
+    [InlineData("foo/bar/baz", 2)]
+    public void TryGetProcessDepth_Returns_Expected_Depth(string input, int expected)
+    {
+        ProcessManager.Sid = input;
+        var actual = ProcessManager.GetProcessDepth();
+
+        Assert.Equal(expected, actual);
+    }
+}
```
jeffhostetler, line 20: Just for completeness, do we care if the input string (ie, the SID that we got from the environment) has a leading or trailing slash?

I'm not saying it should happen, but accidents do...

src/shared/Core/ProcessManager.cs
```diff
@@ -50,9 +57,53 @@ public virtual ChildProcess CreateProcess(string path, string args, bool useShel
...
+            Sid = $"{Sid}/{Guid.NewGuid():D}";
+        }
+        else
+        {
+            // We are the root process; create our own 'root' SID
+            Sid = Guid.NewGuid().ToString("D");
+        }
+
+        Environment.SetEnvironmentVariable(SidEnvar, Sid);
+        Depth = GetProcessDepth();
```
jeffhostetler, line 83: nit: you can save a few cycles if you assign Depth inside the above "if", since we already know the answer in the null case.

src/shared/Core/ProcessManager.cs
```diff
@@ -50,9 +57,53 @@ public virtual ChildProcess CreateProcess(string path, string args, bool useShel
...
+
+    /// <summary>
+    /// Get "depth" of current process relative to top-level GCM process.
+    /// </summary>
+    /// <returns>Depth of current process.</returns>
+    internal static int GetProcessDepth()
+    {
+        char processSeparator = '/';
+        // Convert to string for .NET Framework compatibility.
+        if (!Sid.Contains(processSeparator.ToString()))
```
jeffhostetler, line 94: nit:  Is this test necessary?  it seems like it just adds a second pass. 

That is, when Trace2 is enabled, GCM will almost always be called by Git with a valid SID and we'll already have created our own SID with a slash.  So this test will fail and we'll have to do the second AsSpan() loop anyway.

It feels like we can just omit this and just do the AsSpan() loop and save an extra scan on the string.

jeffhostetler (approved):

src/shared/Core.Tests/ProcessManagerTests.cs
ldennington: Sure - I actually ended up adding a `Trim` in `CreateSid` - felt better to clean up the Sid prior to checking the process depth.

src/shared/Core/ProcessManager.cs
ldennington: Good point!

src/shared/Core/Trace2.cs
```diff
@@ -449,34 +461,122 @@ public abstract class Trace2Message
...
+        // [<time> SP <filename>:<line> SP+] <event-name> [[SP] <event-message>] LF
+        return $"{time} {source,-33} {Event.ToString().ToSnakeCase()} {message}";
+    }
+
+    protected string BuildPerformanceString()
+    {
+        string message = GetEventMessage(Trace2FormatTarget.Performance);
+
+        // The performance format uses local time rather than UTC time.
+        var time = Time.ToLocalTime().ToString(NormalPerfTimeFormat);
```
mjcheetham, line 493: Just double checking that `NormalPerfTimeFormat` is the correct format for the performance target?

mjcheetham (approved):
Looks great! Good stuff :)

src/shared/Core/Trace2.cs
ldennington: Yes! I prefixed `TimeFormat` with `NormalPerf` since the normal and performance format targets use local time while the event format target uses UTC time. Not necessarily my best naming job, though - we can change in future if something better comes to us.

PR #1152, opened 2023-03-14
http: add support for client certificates

### LINKED ISSUE #369: Git Credential Manager Core doesn't support TLS client certificates
With git 2.28.0 using schannel we were able to use client certificates to connect to HTTPS servers secured with mutual TLS with certificates in the user certificate store.  This appears to no longer be possible with GCMC.  Is there an option to use these certificates, especially hardware certificates access through the personal certificate store.

mjcheetham:
Add support for automatically sending client TLS certificates using the Git configuration setting 'http.sslAutoClientCert'.

This setting is currently only [present in Git for Windows](https://github.com/git-for-windows/git/blob/c8edb521bdabec14b07e9142e48cab77a40ba339/http.c#L906-L910), and there is only respected when the SSL backend is "schannel".

Fixes #369

### DIFF
--- a/src/shared/Core/Settings.cs
+++ b/src/shared/Core/Settings.cs
@@ -120,6 +120,11 @@ namespace GitCredentialManager
         bool IsCertificateVerificationEnabled { get; }
 
         /// <summary>
+        /// Automatically send client TLS certificates.
+        /// </summary>
+        bool AutomaticallyUseClientCertificates { get; }
+
+        /// <summary>
         /// Get the proxy setting if configured, or null otherwise.
         /// </summary>
         /// <returns>Proxy setting, or null if not configured.</returns>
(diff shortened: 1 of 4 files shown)


derrickstolee (approved):

src/shared/Core/Settings.cs
```diff
@@ -563,6 +568,9 @@ public bool IsCertificateVerificationEnabled
             }
         }
 
+        public bool AutomaticallyUseClientCertificates =>
+            !TryGetSetting(null, KnownGitCfg.Credential.SectionName, KnownGitCfg.Http.SslAutoClientCert, out string value) && value.ToBooleanyOrDefault(false);
```
mjcheetham, line 572: Just noticed a small logic bug here..

```suggestion
            TryGetSetting(null, KnownGitCfg.Credential.SectionName, KnownGitCfg.Http.SslAutoClientCert, out string value) && value.ToBooleanyOrDefault(false);
```

If there is no setting available we should return `false`!

PR #1154, opened 2023-03-15
Assume GitHub Enterprise instances with a null version number support OAuth

mjcheetham:
Enterprise instances that return a null version are considered to support OAuth. Only disable OAuth as an option if we can verify we are talking to an old Enterprise Server.

This is a less fragile option as we never block the user from using an auth method they could use, and don't need to special case GHAE.

### DIFF
--- a/src/shared/GitHub/GitHubHostProvider.cs
+++ b/src/shared/GitHub/GitHubHostProvider.cs
@@ -263,20 +263,20 @@ namespace GitHub
             {
                 GitHubMetaInfo metaInfo = await _gitHubApi.GetMetaInfoAsync(targetUri);
 
+                // All Enterprise/AE instances support PATs
                 var modes = AuthenticationModes.Pat;
+
+                // If the server says it supports basic auth, we can use that too!
                 if (metaInfo.VerifiablePasswordAuthentication)
                 {
                     modes |= AuthenticationModes.Basic;
                 }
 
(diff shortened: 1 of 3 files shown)


src/shared/GitHub/GitHubHostProvider.cs
```diff
@@ -263,20 +263,20 @@ internal async Task<AuthenticationModes> GetSupportedAuthenticationModesAsync(Ur
...
                 }
 
-                if (StringComparer.OrdinalIgnoreCase.Equals(metaInfo.InstalledVersion, GitHubConstants.GitHubAeVersionString))
-                {
-                    // Assume all GHAE instances have the GCM OAuth application deployed
-                    modes |= AuthenticationModes.OAuth;
-                }
-                else if (Version.TryParse(metaInfo.InstalledVersion, out var version) && version >= GitHubConstants.MinimumOnPremOAuthVersion)
+                // If the version is unknown, we assume it supports OAuth.
+                // If the server version is less than the minimum required, we know we can't use OAuth.
```
derrickstolee, line 276: nit: this line is a bit backwards from what the condition says.

```
// If the server version is at least the minimum, we know it supports OAuth.
```

derrickstolee (approved):

vtbassmatt (approved):
Rubber stamping for the good concept; I didn't review the code 😅

PR #1156, opened 2023-03-17
trace2: add error event

ldennington:
This series adds the `TRACE2` error event.

The first two commits are opportunistic fixes that remove unnecessary components of `Trace2` messages. The next four commits add `Trace2` as a dependency to certain classes where exceptions are thrown in order to capture those exceptions with the new error event. The seventh commit adds the error event, and eighth adds special exceptions that write to `Trace2`. Finally, the ninth commit adds error tracing exceptions and messages throughout the GCM codebase.

### DIFF
--- a/src/shared/Atlassian.Bitbucket/BitbucketAuthentication.cs
+++ b/src/shared/Atlassian.Bitbucket/BitbucketAuthentication.cs
@@ -4,7 +4,6 @@ using System.Net.Http;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
-using Atlassian.Bitbucket.Cloud;
 using GitCredentialManager;
 using GitCredentialManager.Authentication;
 using GitCredentialManager.Authentication.OAuth;
@@ -128,12 +127,12 @@ namespace Atlassian.Bitbucket
                 {
                     if (!output.TryGetValue("username", out userName))
                     {
-                        throw new Exception("Missing username in response");
+                        throw new Trace2Exception(Context.Trace2, "Missing username in response");
                     }
 
                     if (!output.TryGetValue("password", out password))
                     {
-                        throw new Exception("Missing password in response");
+                        throw new Trace2Exception(Context.Trace2, "Missing password in response");
                     }
 
                     return new CredentialsPromptResult(
--- a/src/shared/Atlassian.Bitbucket/BitbucketHostProvider.cs
+++ b/src/shared/Atlassian.Bitbucket/BitbucketHostProvider.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Net;
 using System.Net.Http;
 using System.Threading.Tasks;
 using Atlassian.Bitbucket.Cloud;
@@ -79,7 +78,8 @@ namespace Atlassian.Bitbucket
             if (StringComparer.OrdinalIgnoreCase.Equals(input.Protocol, "http")
                 && BitbucketHelper.IsBitbucketOrg(input))
             {
-                throw new Exception("Unencrypted HTTP is not supported for Bitbucket.org. Ensure the repository remote URL is using HTTPS.");
+                throw new Trace2Exception(_context.Trace2,
+                    "Unencrypted HTTP is not supported for Bitbucket.org. Ensure the repository remote URL is using HTTPS.");
             }
 
             var authModes = await GetSupportedAuthenticationModesAsync(input);
@@ -145,8 +145,9 @@ namespace Atlassian.Bitbucket
                 var result = await _bitbucketAuth.GetCredentialsAsync(remoteUri, input.UserName, authModes);
                 if (result is null || result.AuthenticationMode == AuthenticationModes.None)
                 {
-                    _context.Trace.WriteLine("User cancelled credential prompt");
-                    throw new Exception("User cancelled credential prompt.");
+                    var message = "User cancelled credential prompt";
+                    _context.Trace.WriteLine(message);
+                    throw new Trace2Exception(_context.Trace2, message);
                 }
 
                 switch (result.AuthenticationMode)
@@ -176,8 +177,10 @@ namespace Atlassian.Bitbucket
                 }
                 catch (OAuth2Exception ex)
                 {
-                    _context.Trace.WriteLine("Failed to refresh existing OAuth credential using refresh token");
+                    var message = "Failed to refresh existing OAuth credential using refresh token";
+                    _context.Trace.WriteLine(message);
                     _context.Trace.WriteException(ex);
+                    _context.Trace2.WriteError(message);
 
                     // We failed to refresh the AT using the RT; log the refresh failure and fall through to restart
                     // the OAuth authentication flow
@@ -279,7 +282,7 @@ namespace Atlassian.Bitbucket
             try
             {
                 var authenticationMethods = await _restApiRegistry.Get(input).GetAuthenticationMethodsAsync();
-                
+
                 var modes = AuthenticationModes.None;
 
                 if (authenticationMethods.Contains(AuthenticationMethod.BasicAuth))
@@ -298,10 +301,14 @@ namespace Atlassian.Bitbucket
             }
             catch (Exception ex)
             {
-                _context.Trace.WriteLine($"Failed to query '{input.GetRemoteUri()}' for supported authentication schemes.");
-                _context.Trace.WriteException(ex);
+                var format = "Failed to query '{0}' for supported authentication schemes.";
+                var message = string.Format(format, input.GetRemoteUri());
 
-                _context.Terminal.WriteLine($"warning: failed to query '{input.GetRemoteUri()}' for supported authentication schemes.");
+                _context.Trace.WriteLine(message);
+                _context.Trace.WriteException(ex);
+                _context.Trace2.WriteError(message, format);
+
+                _context.Terminal.WriteLine($"warning: {message}");
 
                 // Fall-back to offering all modes so the user is never blocked from authenticating by at least one mode
                 return AuthenticationModes.All;
@@ -356,7 +363,8 @@ namespace Atlassian.Bitbucket
                 return result.Response.UserName;
             }
 
-            throw new Exception($"Failed to resolve username. HTTP: {result.StatusCode}");
+            throw new Trace2Exception(_context.Trace2,
+                $"Failed to resolve username. HTTP: {result.StatusCode}");
         }
 
         private async Task<string> ResolveBasicAuthUserNameAsync(InputArguments input, string username, string password)
@@ -367,7 +375,8 @@ namespace Atlassian.Bitbucket
                 return result.Response.UserName;
             }
 
-            throw new Exception($"Failed to resolve username. HTTP: {result.StatusCode}");
+            throw new Trace2Exception(_context.Trace2,
+                $"Failed to resolve username. HTTP: {result.StatusCode}");
         }
 
         private async Task<bool> ValidateCredentialsWork(InputArguments input, ICredential credentials, AuthenticationModes authModes)
@@ -404,8 +413,10 @@ namespace Atlassian.Bitbucket
                 }
                 catch (Exception ex)
                 {
-                    _context.Trace.WriteLine($"Failed to validate existing credentials using OAuth");
+                    var message = "Failed to validate existing credentials using OAuth";
+                    _context.Trace.WriteLine(message);
                     _context.Trace.WriteException(ex);
+                    _context.Trace2.WriteError(message);
                 }
             }
 
@@ -419,8 +430,10 @@ namespace Atlassian.Bitbucket
                 }
                 catch (Exception ex)
                 {
-                    _context.Trace.WriteLine($"Failed to validate existing credentials using Basic Auth");
+                    var message = "Failed to validate existing credentials using Basic Auth";
+                    _context.Trace.WriteLine(message);
                     _context.Trace.WriteException(ex);
+                    _context.Trace2.WriteError(message);
                     return false;
                 }
             }
--- a/src/shared/Core.UI/Commands/CredentialsCommand.cs
+++ b/src/shared/Core.UI/Commands/CredentialsCommand.cs
@@ -63,7 +63,7 @@ namespace GitCredentialManager.UI.Commands
 
             if (!viewModel.WindowResult)
             {
-                throw new Exception("User cancelled dialog.");
+                throw new Trace2Exception(Context.Trace2, "User cancelled dialog.");
             }
 
             WriteResult(
--- a/src/shared/Core/Authentication/OAuth/OAuth2Client.cs
+++ b/src/shared/Core/Authentication/OAuth/OAuth2Client.cs
@@ -70,16 +70,24 @@ namespace GitCredentialManager.Authentication.OAuth
         private readonly OAuth2ServerEndpoints _endpoints;
         private readonly Uri _redirectUri;
         private readonly string _clientId;
+        private readonly ITrace2 _trace2;
         private readonly string _clientSecret;
         private readonly bool _addAuthHeader;
 
         private IOAuth2CodeGenerator _codeGenerator;
 
-        public OAuth2Client(HttpClient httpClient, OAuth2ServerEndpoints endpoints, string clientId, Uri redirectUri = null, string clientSecret = null, bool addAuthHeader = true)
+        public OAuth2Client(HttpClient httpClient,
+            OAuth2ServerEndpoints endpoints,
+            string clientId,
+            ITrace2 trace2,
+            Uri redirectUri = null,
+            string clientSecret = null,
+            bool addAuthHeader = true)
         {
             _httpClient = httpClient;
             _endpoints = endpoints;
             _clientId = clientId;
+            _trace2 = trace2;
             _redirectUri = redirectUri;
             _clientSecret = clientSecret;
             _addAuthHeader = addAuthHeader;
--- a/src/shared/Core/Authentication/OAuth/OAuth2Client.cs
+++ b/src/shared/Core/Authentication/OAuth/OAuth2Client.cs
@@ -163,17 +163,19 @@ namespace GitCredentialManager.Authentication.OAuth
             IDictionary<string, string> redirectQueryParams = finalUri.GetQueryParameters();
             if (!redirectQueryParams.TryGetValue(OAuth2Constants.AuthorizationGrantResponse.StateParameter, out string replyState))
             {
-                throw new OAuth2Exception($"Missing '{OAuth2Constants.AuthorizationGrantResponse.StateParameter}' in response.");
+                throw new Trace2OAuth2Exception(_trace2, $"Missing '{OAuth2Constants.AuthorizationGrantResponse.StateParameter}' in response.");
             }
             if (!StringComparer.Ordinal.Equals(state, replyState))
             {
-                throw new OAuth2Exception($"Invalid '{OAuth2Constants.AuthorizationGrantResponse.StateParameter}' in response. Does not match initial request.");
+                throw new Trace2OAuth2Exception(_trace2,
+                    $"Missing '{OAuth2Constants.AuthorizationGrantResponse.StateParameter}' in response.");
             }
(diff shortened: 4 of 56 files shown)


src/shared/Core/Trace2.cs
```diff
@@ -255,6 +271,34 @@ public void Stop(int exitCode, string filePath, int lineNumber)
...
+        {
+            return;
+        }
+
+        WriteMessage(new ErrorMessage()
+        {
+            Event = Trace2Event.Error,
+            Sid = _sid,
+            Time = DateTimeOffset.UtcNow,
+            File = Path.GetFileName(filePath).ToLower(),
```
jeffhostetler, line 294: nit: should we be doing ToLower() on the source pathname?   i guess we are in the other methods too.  Feels Windows-centric maybe.   I'm not sure it matters one way or another, TBH.

src/shared/Core/Trace2.cs
```diff
@@ -796,3 +840,42 @@ protected override string GetEventMessage(Trace2FormatTarget formatTarget)
...
     }
 }
+
+public class ErrorMessage : Trace2Message
+{
+    [JsonProperty("msg", Order = 7)]
+    public string Message { get; set; }
+
+    [JsonProperty("format", Order = 8)]
+    public string FormattedErrorMessage { get; set; }
```
jeffhostetler, line 850: The `FormattedErrorMessage` property name is a bit misleading, I think.  In the Trace2 spec, I have "msg" as the actual error message to be printed and "fmt" as the format string used in a `sprintf()` like call to build "msg".  So would it be better to call this `FormatString` or `ParameterizedMessage` or something??

src/shared/Core/Trace2.cs
```diff
@@ -255,6 +271,34 @@ public void Stop(int exitCode, string filePath, int lineNumber)
...
+
+        WriteMessage(new ErrorMessage()
+        {
+            Event = Trace2Event.Error,
+            Sid = _sid,
+            Time = DateTimeOffset.UtcNow,
+            File = Path.GetFileName(filePath).ToLower(),
+            Line = lineNumber,
+            Message = errorMessage,
+            FormattedErrorMessage = formattedErrorMessage ?? "",
```
jeffhostetler, line 297: Given the C# model of building message strings is different from C, would it be better to default the format string to the message string if the format string is not supplied?

With the assumption being that your call sites in the next commit will send one string if there's no PII/sensitive information (and you can build the message the usual way with inline @"{var}" tokens).  (More on this in the next commit.)

And only if there is something that we think is sensitive, would we use the 2 arg version of `Trace2.WriteError()`.

The original purpose of the "fmt" and "msg" field pair was that we could aggregate errors of the form "invalid pathname: %s" in SQL/kusto/wherever on the "fmt" field without using regex magic -- in addition to the privacy aspects, since the actual pathname might be sensitive.

src/shared/Atlassian.Bitbucket/BitbucketAuthentication.cs
```diff
@@ -88,7 +88,11 @@ public async Task<CredentialsPromptResult> GetCredentialsAsync(Uri targetUri, st
             // We need at least one mode!
             if (modes == AuthenticationModes.None)
             {
-                throw new ArgumentException(@$"Must specify at least one {nameof(AuthenticationModes)}", nameof(modes));
+                var format = @"Must specify at least one {0}";
```
jeffhostetler, line 91: As I said in the previous commit, splitting up the format and message to use the 2 arg version of Trace2.WriteError() seems like a pain for many calls.

I don't have good visibility here to the `nameof(AuthenticationModes)` symbol here, so I can't say if this is useful split or not.  I don't think there's any PII here, but it might be a good use of a split if we want to group OAuth vs XYZ vs ABC schemes.

src/shared/Atlassian.Bitbucket/BitbucketHostProvider.cs
```diff
@@ -404,8 +425,10 @@ private async Task<bool> ValidateCredentialsWork(InputArguments input, ICredenti
                 }
                 catch (Exception ex)
                 {
-                    _context.Trace.WriteLine($"Failed to validate existing credentials using OAuth");
+                    var message = "Failed to validate existing credentials using OAuth";
+                    _context.Trace.WriteLine(message);
                     _context.Trace.WriteException(ex);
+                    _context.Trace2.WriteError(message);
```
jeffhostetler, line 431: I agree here, let's NOT serialize the exception onto the end of the trace2 error message.

src/shared/Core.UI/Commands/CredentialsCommand.cs
```diff
@@ -63,7 +63,9 @@ private async Task<int> ExecuteAsync(CommandOptions options)
 
             if (!viewModel.WindowResult)
             {
-                throw new Exception("User cancelled dialog.");
+                var message = "User cancelled dialog.";
+                Context.Trace2.WriteError(message);
+                throw new Exception(message);
```
jeffhostetler, line 68: Would it help (or would this be a hack) if we made `Context.Trace2.WriteError(m)` return `m`, so that you could do:

```
throw new Exception(Context.Trace2.WriteError(message))`
```

That might be a style clash though, so feel free to disregard.

src/shared/Core/CredentialStore.cs
```diff
@@ -166,32 +169,41 @@ private void ValidateWindowsCredentialManager()
         {
             if (!PlatformUtils.IsWindows())
             {
-                throw new Exception(
-                    $"Can only use the '{StoreNames.WindowsCredentialManager}' credential store on Windows." +
-                    Environment.NewLine +
-                    $"See {Constants.HelpUrls.GcmCredentialStores} for more information."
-                );
+                var format = "Can only use the '{0}' credential store on Windows.{1}" +
+                             "See {3} for more information.";
```
jeffhostetler, line 173: should {3} be {2} ??

src/shared/Core/CredentialStore.cs
```diff
@@ -166,32 +169,41 @@ private void ValidateWindowsCredentialManager()
...
 
             if (!WindowsCredentialManager.CanPersist())
             {
-                throw new Exception(
-                    $"Unable to persist credentials with the '{StoreNames.WindowsCredentialManager}' credential store." +
-                    Environment.NewLine +
-                    $"See {Constants.HelpUrls.GcmCredentialStores} for more information."
-                );
+                var format = "Unable to persist credentials with the '{0}' credential store.{1}" +
+                             "See {3} for more information.";
```
jeffhostetler, line 185: and here??

src/shared/GitHub.UI/Commands/DeviceCommand.cs
```diff
@@ -38,6 +38,8 @@ private async Task<int> ExecuteAsync(string code, string url)
 
             if (!viewModel.WindowResult)
             {
+                var message = "User cancelled dialog.";
+                Context.Trace2.WriteError(message);
                 throw new Exception("User cancelled dialog.");
```
jeffhostetler, line 43: did you mean to use `message` in the throw ?

src/shared/GitHub.UI/Commands/TwoFactorCommand.cs
```diff
@@ -33,6 +33,8 @@ private async Task<int> ExecuteAsync(bool sms)
 
             if (!viewModel.WindowResult)
             {
+                var message = "User cancelled dialog.";
+                Context.Trace2.WriteError(message);
                 throw new Exception("User cancelled dialog.");
```
jeffhostetler, line 38: and here?

jeffhostetler:
This looks good. I did have a few comments. Sorry that it causes so much churn in the code to squeeze it in.

jeffhostetler (approved):

jeffhostetler:
Oh, I did find it a little odd that the Windows terminal didn't need to be updated, but that's more of a curiosity than a question.

jeffhostetler:
And yes, you mentioned somewhere about not getting every error/exception because of excess churn, and that is fine.

src/shared/Core/Trace2.cs
ldennington: Sure, I like `ParameterizedMessage`. That'll be an easy change to make.

src/shared/Core/Trace2.cs
ldennington: I'm fine with excluding this - I think I originally did this because I thought file names were written as all lowercase per `TRACE2` convention, but now I'm not sure where I got that idea.

src/shared/Core/Trace2.cs
ldennington: Hmm...if we do this, won't the error message often be printed twice? Since it will be the same as the format string?

src/shared/Atlassian.Bitbucket/BitbucketAuthentication.cs
ldennington: I honestly just split anything that was interpolated. And yes, it was a pain - it took up a fair part of yesterday 😢 . Probably worth pairing on Monday so I fully understand when to split things vs. when not to so as not to use up more time spinning on it unnecessarily. I'll grab some time!

src/shared/Core.UI/Commands/CredentialsCommand.cs
ldennington: I'd like to keep them separate - I think it's a little more readable that way and I'd like to stay consistent with the other `Trace2` methods that don't return anything.

src/shared/Core/CredentialStore.cs
ldennington: Eagle eye! Thank you.

ldennington:
> Oh, I did find it a little odd that the Windows terminal didn't need to be updated, but that's more of a curiosity than a question.

I just didn't want to pass it an extra parameter that was unused, since that class doesn't have any exceptions.

src/shared/Core/Trace2.cs
```diff
@@ -796,3 +840,42 @@ protected override string GetEventMessage(Trace2FormatTarget formatTarget)
         return sb.ToString();
     }
 }
+
+public class ErrorMessage : Trace2Message
+{
+    [JsonProperty("msg", Order = 7)]
```
jeffhostetler, line 846: nit: does the `Order` attribute here mean that this field will be the 7th in the resulting JSON?

Most consumers of the JSON output don't/shouldn't care about the ordering of the fields. I just read/parse it into a dictionary and access the fields as I need them, for example.  So you could omit them here -- unless it makes testing easier (some readers might want to grep for "\"msg\":.*,\"fmt\":.*" for example, so it is ok to keep it if you think this might be the case).

src/shared/TestInfrastructure/Objects/NullTrace.cs
```diff
@@ -78,6 +78,12 @@ public class NullTrace2 : ITrace2
             string filePath = "",
             int lineNumber = 0) { }
 
+        public void WriteError(string errorMessage,
+            string parameterizedMessage = null,
+            string filePath = "",
```
jeffhostetler, line 83: question: in the ITrace2 definition, you add the `...CompilerServices...` thing to (I presume) override the default value to the call-site's filename.  You also use it in the main implementation's declaration.  Do we need that here or do we inherit the default value override from the interface declaration?

(not a criticism, just asking)

src/shared/Core/Authentication/OAuth/OAuth2Client.cs
```diff
@@ -383,10 +386,10 @@ protected Exception CreateExceptionFromResponse(string json)
         {
             if (TryCreateExceptionFromResponse(json, out OAuth2Exception exception))
             {
-                return exception;
+                return new Trace2OAuth2Exception(_trace2, exception.Message);
             }
 
-            return new OAuth2Exception($"Unknown OAuth error: {json}");
+            return new Trace2OAuth2Exception(_trace2, $"Unknown OAuth error: {json}");
```
jeffhostetler, line 392: Does `json` have PII in it? 

It looks like we only get here if the JSON parser barfs on the string, but do we know it is safe to log it?

src/shared/Core/Authentication/OAuth/OAuth2Client.cs
```diff
@@ -383,10 +386,10 @@ protected Exception CreateExceptionFromResponse(string json)
         {
             if (TryCreateExceptionFromResponse(json, out OAuth2Exception exception))
             {
-                return exception;
+                return new Trace2OAuth2Exception(_trace2, exception.Message);
```
jeffhostetler, line 389: This one feels weird.  We already have an `Exception` from the `Try...()` and we're stealing the message and throwing it away to create a new one -- just to get the `WriteError()` call.  

Won'd this also throw away the original throw context (stack info and etc)?

Maybe we should just manually do the `WriteError()` here and let it return the original exception.

src/shared/Core/CredentialStore.cs
```diff
@@ -107,7 +107,7 @@ private void EnsureBackingStore()
                     AppendAvailableStoreList(sb);
                     sb.AppendLine();
                     sb.AppendLine($"See {Constants.HelpUrls.GcmCredentialStores} for more information.");
-                    throw new Exception(sb.ToString());
+                    throw new Trace2Exception(_context.Trace2, sb.ToString());
```
jeffhostetler, line 110: I'm wondering if we really want a multiline message with hint text in the trace log ??   Especially, the hint text.

Perhaps we should call `WriteError()` after the initial message string is inserted into the buffer (between lines 100 and 101) and before the hint text is appended.  Then just throw using the original Exception() class.

src/shared/Core/Git.cs
```diff
@@ -210,7 +212,7 @@ public ChildProcess CreateProcess(string args)
             var process = _processManager.CreateProcess(procStartInfo);
             if (process is null)
             {
-                throw new Exception($"Failed to start Git helper '{args}'");
+                throw new Trace2Exception(_trace2, $"Failed to start Git helper '{args}'");
```
jeffhostetler, line 215: do we know if `args` contains PII ??

src/shared/Core/HostProviderRegistry.cs
```diff
@@ -66,8 +66,7 @@ public void Register(IHostProvider hostProvider, HostProviderPriority priority)
 
             if (StringComparer.OrdinalIgnoreCase.Equals(hostProvider.Id, Constants.ProviderIdAuto))
             {
-                throw new ArgumentException(
-                    $"A host provider cannot be registered with the ID '{Constants.ProviderIdAuto}'",
+                throw new ArgumentException($"A host provider cannot be registered with the ID '{Constants.ProviderIdAuto}'",
```
jeffhostetler, line 69: This is just a whitespace change, right?

Did you mean to wrap it?

src/shared/Core/Trace2Exception.cs
```diff
@@ -0,0 +1,67 @@
...
+using System.IO;
+using GitCredentialManager.Authentication.OAuth;
+using GitCredentialManager.Interop;
+
+namespace GitCredentialManager;
+
+public class Trace2Exception : Exception
+{
+    public Trace2Exception(ITrace2 trace2, string message) : base(message)
+    {
```
jeffhostetler, line 12: Should all of these be guarded with `if (trace2 != null)` ??

src/shared/Core/Trace2Exception.cs
```diff
@@ -0,0 +1,67 @@
...
+
+    public Trace2InteropException(ITrace2 trace2, string message, Win32Exception ex) : base(message, ex)
+    {
+        trace2.WriteError(message);
+    }
+}
+
+public class Trace2GitException : GitException
+{
+    public Trace2GitException(ITrace2 trace2, string message, string gitMessage, int errorCode) :
```
jeffhostetler, line 53: I stumbled a bit here because this method takes two strings and I was expecting the second to be an instance of a `messageFormat` but it is a different creature.  I realize that that comes from the base class, but still it caused me to double check the caller and callee a few times.

Not sure there is an action item here, just making a note here.

src/shared/Core/Trace2Exception.cs
```diff
@@ -0,0 +1,67 @@
...
+        base(message, gitMessage, errorCode)
+    {
+        trace2.WriteError($"message: {message} git message: {gitMessage} error code: {errorCode}");
```
jeffhostetler, line 56: Do we know if the provided `gitMessage` is PII clean ??

I tracked it back and it looks like it is the contents of stderr of the child Git command.  I'm not sure what that contains in practice, so I'm just asking.

It looks like all of the callers pass a string literal for the `message`, so they should be safe.
I'm wondering if we should do something like:

```
string fmt = $"message: '{message}' error code: {errorCode} git message: '\{0\}'"
string msg = string.Format(fmt, gitMessage)

trace2.WriteError(msg, fmt)
```

jeffhostetler:
I looked thru the entire series.  I had a few comments, but it all looks good.

src/shared/Core/Trace2.cs
ldennington: I'm fine with omitting! I included them to ensure the same order as the fields coming from Git, but since there won't typically be any human readers, it seems unnecessary.

src/shared/TestInfrastructure/Objects/NullTrace.cs
ldennington: It's inherited! Good eye though.

src/shared/Core/Authentication/OAuth/OAuth2Client.cs
ldennington: Mmm...probably not. I'll work on adding the format here.

src/shared/Core/Trace2Exception.cs
ldennington: I don't think so - by virtue of having the `Trace2` instance to pass to this, we know that it's been created. It may not be initialized yet, but we guard against that in the `WriteError` method.

src/shared/Core/Trace2Exception.cs
ldennington: Probably a good idea.

src/shared/Core/Trace2Exception.cs
jeffhostetler: maybe move the error code earlier in the calling sequence, so we would see (effectively)
the current (and later if we have to add the fmt version):

```
Trace2GitException(iTrace2, string,         int, string gitMessage)
Trace2GitException(iTrace2, string, string, int, string gitMessage)
```

src/shared/Core/Interop/Windows/Native/Win32Error.cs
```diff
@@ -123,6 +136,8 @@ public static void ThrowIfError(int error, string defaultErrorMessage = "Unknown
                 default:
                     // The Win32Exception constructor will automatically get the human-readable
                     // message for the error code.
+                    if (trace2 != null)
```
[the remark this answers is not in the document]
jeffhostetler: ah, here was the place i mentioned in the above if !null comment inside the exception classes.

if you hide the `(trace2 != null)` inside the exception class, then you can just always use the new `Trace2...Exception()` here.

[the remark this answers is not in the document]
ldennington: Hmm...maybe we're seeing things differently. I defaulted the `Trace2` parameter to `null` to ensure that callers who didn't want to write to `Trace2` didn't have to change. So, a caller could do something like this:
```
Win32Error.ThrowIfError(error, "Failed to show credential prompt.");
```

In that case, `trace2` would be null by default and the old exception would be thrown.

Alternatively, a caller could do something like this:
```
Win32Error.ThrowIfError(error, "Failed to show credential prompt.", _trace2);
```

At which point trace2 would be the instance passed in (not `null`) and the trace2 exception would be thrown.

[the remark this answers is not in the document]
jeffhostetler: that's fine. i didn't see that you had an optional arg there.

jeffhostetler:
this looks good. thanks!

ldennington:
> this looks good. thanks!

Thank you! I will plan to clean up the test failures and merge!

PR #1162, opened 2023-03-24
trace2: add regions

ldennington:
Add `TRACE2` region tracing capability.

The first six commits contain minor refactors, bug fixes, and cleanups. The first moves Trace2 messages into their own file. The second removes the unnecessary `AddDots()` method. The third fixes certain conflated timing information in `child_start` and `child_exit` events. The fourth ensures we're writing exit message times in UTC. The fifth adds descriptive comments to Trace2-related classes that did not yet have them. The sixth adds best-effort logic to capture thread names. 

The next three commits are dedicated to preparing for and adding region-related logic. The seventh commit updates custom performance format span handling to account for the repo and category fields, which are used in region tracing. The eighth adds region-related events, encapsulated in a class that inherits from `DisposableObject` (which in turn inherits from `IDisposable`). The ninth adds region tracing to methods of interest pertaining to OAuth authentication.

The final commit updates GCM's documentation to inform users of how to enable and utilize Trace2 tracing.

A final note: This implementation does not support nested regions, as it is not required for the current regions of interest.


src/shared/Core/Authentication/OAuth/OAuth2Client.cs
```diff
@@ -184,13 +183,19 @@ public IOAuth2CodeGenerator CodeGenerator
                     $"Missing '{OAuth2Constants.AuthorizationGrantResponse.AuthorizationCodeParameter}' in response.");
             }
 
-            _trace2.WriteRegionLeave((DateTimeOffset.UtcNow - startTime).Seconds,
-                Constants.Trace2Region.Oauth2Category, Constants.Trace2Region.AuthCodeLabel, thread);
-            return new OAuth2AuthorizationCodeResult(authCode, redirectUri, codeVerifier);
+            var result = new OAuth2AuthorizationCodeResult(authCode, redirectUri, codeVerifier);
+            var endTime = (DateTimeOffset.UtcNow - startTime).TotalSeconds;
+            _trace2.WriteRegionLeave(endTime, OAuth2Constants.Trace2Category, label, thread);
```
derrickstolee, line 188: One thing that C# has that I really wish we could do in the C code of Git is the `using` statement. If the trace2 machinery creates an `IDisposable` object that writes the "region enter" event on construction and "region exit" event on disposal, then we can do this kind of pattern:

```cs
using (IDisposable resion = _trace2.StartRegion(...))
{
   // Do stuff
} // exit event written no matter how we exit the block.
```

All of the stuff about thread names, start times, and end times, seem like they would be best encapsulated by that disposable object, to make it incredibly easy for consumers to add regions whenever it seems prudent.

ldennington: I agree this is a much more user-friendly approach. I'll see if I can get it working tomorrow.

ldennington: Interested in your feedback on the implementation.

ldennington:
Marked as ready for review, since the only remaining failures are due to the expired CodeQL certificate.

src/shared/Core/Authentication/OAuth/OAuth2Client.cs
jeffhostetler: I like the `IDisposable` changes.  I'm wondering if we can make (not necessarily now) a `Region` also implement `ITrace2` and holds a reference to the containing `ITrace2` such that it embeds the former.  Then we could do:

```
foo(_trace2 ITrace2, ...)
{
    using (r1 = _trace2.StartRegion(...))
    {
        using (r2 = r1.StartRegion(...))
        {
            // call a function and pass r2 rather than _trace, for example...
        }
    }
}
```

Going one further, if we also defined that for Threads, where the `using` sets the thread name.  This could be
used in lamba expressions and we wouldn't have to worry about thread-pool vs named vs anonymous threads.

```
foo(_trace2 ITrace2, ...)
{
    using (t1 = _trace2.StartThread("xyzzy", ...))
    {
        using (r1_in_t1 = t1.StartRegion(...))
        {
...
```

This may be too much...

ldennington:
Latest push adds a new commit that writes exit message times in UTC rather than local time. @jeffhostetler and I discovered this little bug while doing some testing with App Insights this morning.

src/shared/Core/Trace2Message.cs
```diff
@@ -329,6 +331,8 @@ public class ChildExitMessage : Trace2Message
     public int Code { get; set; }
 
     [JsonProperty("t_rel")]
+    public double RelativeTime { get; set; }
+
     public double ElapsedTime { get; set; }
```
mjcheetham, line 363: This doesn't have a `JsonProperty` anymore? Is `ElapsedTime` still relevant?

src/shared/Core/Trace2Message.cs
```diff
@@ -277,6 +277,8 @@ public class ChildStartMessage : Trace2Message
     [JsonProperty("argv")]
     public IList<string> Argv { get; set; }
 
+    public double ElapsedTime { get; set; }
```
mjcheetham, line 307: Where is this set? Is it always `0`?

src/shared/Core/Trace2.cs
```diff
@@ -31,12 +31,19 @@ public enum Trace2ProcessClass
...
+/// Check <see cref="Trace2FormatTarget"/> for supported formats.
+/// </summary>
 public class Trace2Settings
 {
     public IDictionary<Trace2FormatTarget, string> FormatTargetsAndValues { get; set; } =
         new Dictionary<Trace2FormatTarget, string>();
 }
 
+/// <summary>
+/// Specifies a "span" (i.e. space between two pipes) for the performance format target.
```
mjcheetham, line 50: Maybe say "text span" as a `Span<T>` exists in .NET.

derrickstolee (approved):
The disposable interface looks great. Just one question about a method prototype.

src/shared/Git-Credential-Manager/Program.cs
```diff
@@ -20,10 +21,12 @@ public static void Main(string[] args)
             using (var app = new Application(context))
             {
                 // Initialize TRACE2 system
+                // TRACE2 convention is for main thread of execution to be named "main"
+                Thread.CurrentThread.Name = "main";
```
mjcheetham, line 25: Since each helper is a child processes, should all of these `Program.cs` threads not also be called `main`? It's just like Git starting itself as a child, no?

src/shared/Core/Trace2.cs
```diff
@@ -481,3 +504,16 @@ private void WriteMessage(Trace2Message message)
         }
     }
 }
+
+public static class ThreadHelpers
+{
+    public static string BuildThreadName()
```
mjcheetham, line 655: We are passing `ThreadHelpers.BuildThreadName()` in almost all of these places.. should we just not always use the `BuildThreadName()` logic inside each `WriteX` method?

To capture all possible thread names, including `main` what about this?

```csharp
// If the thread is named, is it!
if (!string.IsNullOrEmpty(Thread.CurrentThread.Name))
{
    return Thread.CurrentThread.Name;
}

// Otherwise if this is the entry thread, call it "main"
if (Thread.CurrentThread.ManagedThreadId == 0)
{
    return "main";
}

// If this is a thread pool thread, name it as such
if (Thread.CurrentThread.IsThreadPoolThread)
{
    return $"thread_pool_{Environment.CurrentManagedThreadId}";
}

// We don't know what this thread is!
return string.Empty;
```

src/shared/Core/Trace2.cs
```diff
@@ -150,6 +190,51 @@ public interface ITrace2 : IDisposable
...
+    /// <param name="message">Message associated with entering region.</param>
+    /// <param name="filePath">Path of the file this method is called from.</param>
+    /// <param name="lineNumber">Line number of file this method is called from.</param>
+    void WriteRegionEnter(
+        string category,
+        string label,
+        string threadName,
+        string message = "",
+        string filePath = "",
+        int lineNumber = 0);
```
mjcheetham, line 218: Are we're missing the `[System.Runtime.CompilerServices.CallerX]` attributes here?

src/shared/Core/Trace2Message.cs
```diff
@@ -447,3 +437,95 @@ protected override string GetEventMessage(Trace2FormatTarget formatTarget)
...
+
+public abstract class RegionMessage : Trace2Message
+{
+    [JsonProperty("repo")]
+
+    // Defaults to 1, as does Git.
+    // See https://git-scm.com/docs/api-trace2#Documentation/technical/api-trace2.txt-codeltrepo-idgtcode for details.
+    public int Repo { get; set; } = 1;
+
+    // Nested regions are not currently supported in GCM, so this value should always be 1.
```
mjcheetham, line 449: Maybe add a "TODO" here.. makes it easy to search for in the future :)

src/shared/Core/Authentication/OAuth/OAuth2Client.cs
```diff
@@ -104,6 +104,9 @@ public IOAuth2CodeGenerator CodeGenerator
         public async Task<OAuth2AuthorizationCodeResult> GetAuthorizationCodeAsync(IEnumerable<string> scopes,
             IOAuth2WebBrowser browser, IDictionary<string, string> extraQueryParams, CancellationToken ct)
         {
+            var label = "get auth code";
+            using IDisposable region = _trace2.CreateRegion(OAuth2Constants.Trace2Category, label);
```
mjcheetham, line 108: The `GetAuthorizationCodeAsync` method (and the `GetTokenByDeviceCodeAsync` method below) involve a period of time where we wait for user interaction. The former being the user interacting with the launched browser to sign in, and the latter being the user navigating to a login page and entering a device code on another device (and we are polling for the server to say "done").

I wonder if we should scope these regions to only parts that don't include user interaction, else these figures will vary widely and be of no use performance tracing-wise.

mjcheetham (commented):
Looking great! I had some comments primarily on removing the need to always call `ThreadHelper.BuildThreadName()` and around specifics of what the OAuth regions cover.

src/shared/Core/Trace2Message.cs
ldennington: Yep, we need it to correctly build the span for the Performance format target. However, it's not needed for the Event format, which is why there is no Json attribute.

src/shared/Core/Trace2Message.cs
ldennington: It's set in `WriteChildStart` in `Trace2.cs`.

src/shared/Core/Trace2.cs
ldennington: Nice! Thanks for the much improved suggestion.

src/shared/Git-Credential-Manager/Program.cs
ldennington: I re-organized the below method you suggested a bit to account for this.

src/shared/Core/Trace2.cs
ldennington: Well spotted.

src/shared/Core/Authentication/OAuth/OAuth2Client.cs
ldennington: That's a good point. Since there isn't server interaction without user interaction in these cases, I will just remove these regions.

mjcheetham: Ahh. I think you may have removed the region from the wrong side of the Device Code flow.

Device code flow consists of two parts:

1. `GetDeviceCodeAsync` which has no user input - it tells the server to start a device code flow operation and prints the code to the user. It should return quickly.

2. `GetTokenByDeviceCodeAsync` which polls the server until the user has completed the sign in on another device. This is the method that is bounded by the user interaction speed.

Looks like you've removed the region from 1 and not 2.

src/windows/GitLab.UI.Windows/Program.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Diagnostics;
+using System.Threading;
```
mjcheetham, line 3: This `using` import is no longer needed (along with the other UI helper `Program.cs`s.

src/shared/Git-Credential-Manager/Program.cs
```diff
@@ -86,7 +89,7 @@ public static void Main(string[] args)
                     .GetAwaiter()
                     .GetResult();
 
-                context.Trace2.Stop(exitCode);
+                context.Trace2.Stop(exitCode, Thread.CurrentThread.Name);
```
mjcheetham, line 92: Every call to `Stop` takes `Thread.CurrentThread.Name`.. perhaps we should also use `BuildThreadName`?

src/shared/Git-Credential-Manager/Program.cs
mjcheetham: @ldennington do we still need to explicitly name this thread?
It's the entry thread so will be `ManagedThreadId == 0`, which we call `"main"`.

src/shared/Core/Trace2.cs
```diff
@@ -1,7 +1,9 @@
 using System;
 using System.Collections.Generic;
+using System.ComponentModel;
 using System.IO;
 using System.IO.Pipes;
+using System.Reflection.Emit;
```
mjcheetham, line 6: Is this used? I don't see any changes here that require reflection. Perhaps I'm missing something though.

mjcheetham (approved):
A few nits and one extra possible cleanup in `Stop` with ThreadName.

src/windows/GitLab.UI.Windows/Program.cs
ldennington: I think the UI Helpers do need it since they name the thread they spawn `AppMain`, right?

ldennington: Ah, well the Linux/macOS ones do but Windows don't. Got it.

src/shared/Git-Credential-Manager/Program.cs
ldennington: That is incorrect and should be removed - thanks for catching!

src/shared/Git-Credential-Manager/Program.cs
ldennington: Nope, took care of that!

PR #1163, opened 2023-03-24
Update install.md

nickarino:
add Winget install info

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -176,6 +176,10 @@ There are two flavors of standalone installation on Windows:
 To install, double-click the desired installation package and follow the
 instructions presented.
 
+### Winget
+Winget is a package manager for Windows.  
+`winget install --id Microsoft.GitCredentialManagerCore`
+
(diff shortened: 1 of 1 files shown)


ldennington:
Hello there! Thank you for your contribution. We actually no longer publish new releases to winget, as we'd like to encourage users to install via the preferred method (i.e. the bundled version of GCM that ships with Git for Windows).

For reference, the most recent version published to winget was `2.0.785`, which is 5 versions behind the latest release (and also contains the `core` moniker on the end that we've since dropped from the project).

I'm not sure whether the capability to deprecate packages in `winget-pkgs` exists, but I went ahead and [opened an issue](https://github.com/microsoft/winget-pkgs/issues/100542) to ask about it.

nickarino:
Thank you, Lessley. 
So, the recommended command line way to install is 
`dotnet tool install -g git-credential-manager`

ldennington:
Yes, if you're planning to install without Git via the commandline, then that would be the way to go.

Issue #1165, opened 2023-03-27
fatal: Object reference not set to an instance of an object.


ThePlenkov:
### Version

2.0.973+784e7c794e

### Operating system

Linux

### OS version or distribution

Linux 5.15.90.1-microsoft-standard-WSL2 #1 SMP Fri Jan 27 02:56:13 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

### Git hosting provider(s)

GitLab

### (Azure DevOps only) What format is your remote URL?

None

### Can you access the remote repository directly in the browser?

Yes, I can access the repository

### Expected behavior

GCM should pick up credentials from the store. 

### Actual behavior

I still can input username and password and pull works - but it doesn't store the password.
I feel like may be the issue is more like with storing rather than with reading

Please review log

### Logs

WSLENV=$WSLENV:GCM_TRACE:GIT_TRACE GCM_TRACE=1 GIT_TRACE=1 git fetch
16:23:53.964529 git.c:439               trace: built-in: git fetch
16:23:53.964959 run-command.c:655       trace: run_command: GIT_DIR=.git git remote-https origin https://gitlab.myorg.com/finsys/docker/devops-docker-xs-cli.git
16:23:53.967078 git.c:725               trace: exec: git-remote-https origin https://gitlab.myorg.com/finsys/docker/devops-docker-xs-cli.git
16:23:53.967156 run-command.c:655       trace: run_command: git-remote-https origin https://gitlab.myorg.com/finsys/docker/devops-docker-xs-cli.git
16:23:54.108945 run-command.c:655       trace: run_command: '/usr/local/bin/git-credential-manager get'
16:23:54.252600 git.c:439               trace: built-in: git version
16:23:54.285050 git.c:439               trace: built-in: git config --null trace2.eventtarget
16:23:54.286986 git.c:439               trace: built-in: git config --null trace2.normaltarget
16:23:54.288267 git.c:439               trace: built-in: git config --null trace2.perftarget
16:23:54.310013 git.c:439               trace: built-in: git config --null credential.httpProxy
16:23:54.311355 git.c:439               trace: built-in: git config --null http.proxy
16:23:54.329163 git.c:439               trace: built-in: git config --null http.sslBackend
16:23:54.331603 git.c:439               trace: built-in: git config --null http.sslBackend
16:23:54.333684 git.c:439               trace: built-in: git config --null http.sslVerify
[...]

PR #1166, opened 2023-03-27
fixing fatal: Object reference not set to an instance of an object. #…

ThePlenkov:
Possible fix for #1165 . I have reviewed the code and was not able to find such an available method used in the original code.


ldennington:
Closing in favor of #1167

PR #1172, opened 2023-03-29
Cache Git configuration used in Settings for the lifetime of the process

mjcheetham:
When enumerating all Git configuration values in `Settings`, we currently cache a copy of all values for the lifetime of the `GetSettingValues()` method. This saves us some calls to Git, but we can do better.

Pull-up the cache of Git configuration entries to the lifetime of the `Settings` object, meaning all calls to `GetSettingValues()` over the lifetime of the process can benefit from the cache! The savings here can be significant as each `ISettings` property typically results in a call to `GetSettingValues()`.

Whilst in this space we can fix a couple of bugs:

1. We were not calling `git config --type=path` for `isPath = true` variables when resolving settings, and instead consulting the cache first. We always want to go to Git for these types of values.

2. We were not using the existing cache for unscoped entries!

In my limited testing this results in a 2-3x speedup in calls to `get && store`!

### DIFF
--- a/src/shared/Core/Settings.cs
+++ b/src/shared/Core/Settings.cs
@@ -429,7 +429,9 @@ namespace GitCredentialManager
                  *        property = value
                  *
                  */
-                if (config.TryGet($"{section}.{property}", isPath, out value))
+                string name = $"{section}.{property}";
+                if ((isPath && config.TryGet(name, isPath, out value)) ||
+                    configEntries.TryGetValue(name, out value))
                 {
                     yield return value;
                 }
--- a/src/shared/Core/Settings.cs
+++ b/src/shared/Core/Settings.cs
@@ -397,8 +397,8 @@ namespace GitCredentialManager
                         // Look for a scoped entry that includes the scheme "protocol://example.com" first as
                         // this is more specific. If `isPath` is true, then re-get the value from the
                         // `GitConfiguration` with `isPath` specified.
-                        if (configEntries.TryGetValue(queryName, out value) &&
-                            (!isPath || config.TryGet(queryName, isPath, out value)))
+                        if ((isPath && config.TryGet(queryName, isPath, out value)) ||
+                            configEntries.TryGetValue(queryName, out value))
                         {
                             yield return value;
                         }
@@ -408,8 +408,8 @@ namespace GitCredentialManager
                         // `isPath` specified.
                         string scopeWithoutScheme = scope.TrimUntilIndexOf(Uri.SchemeDelimiter);
                         string queryWithSchemeName = $"{section}.{scopeWithoutScheme}.{property}";
-                        if (configEntries.TryGetValue(queryWithSchemeName, out value) &&
-                            (!isPath || config.TryGet(queryWithSchemeName, isPath, out value)))
+                        if ((isPath && config.TryGet(queryWithSchemeName, isPath, out value)) ||
+                            configEntries.TryGetValue(queryWithSchemeName, out value))
                         {
                             yield return value;
                         }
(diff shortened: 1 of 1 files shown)


src/shared/Core/Settings.cs
```diff
@@ -334,6 +336,28 @@ public IEnumerable<string> GetSettingValues(string envarName, string section, st
             {
                 IGitConfiguration config = _git.GetConfiguration();
 
+                // Enumerate all configuration entries with the correct section and property name
+                // and make a local copy of them here to avoid needing to call `TryGetValue` on the
+                // IGitConfiguration object multiple times in a loop below.
+
+                // This is a performance optimisation to avoid calling `TryGet` on the IGitConfiguration
+                // object multiple times in a loop below, or each time this entire method is called.
+                // The assumption is that the configuration entries will not change between invocations
```
vdye, line 345:
```suggestion
                // The assumption is that the configuration entries will not during a single invocation
```

If I understand correctly, this caching is scoped to a single Git Credential Manager process (i.e., a single invocation). "Will not change between invocations" to me implies an expectation that a Git config won't change between two separate calls to GCM, which isn't a guarantee.

src/shared/Core/Settings.cs
```diff
@@ -334,6 +336,28 @@ public IEnumerable<string> GetSettingValues(string envarName, string section, st
...
+                // of Git Credential Manager, which is reasonable given process lifetime is typically less
+                // than a few seconds. For some entries (type=path), we still need to ask Git in order to
+                // expand the path correctly.
+                if (_configEntries is null)
+                {
+                    _configEntries = new Dictionary<string, string>(GitConfigurationKeyComparer.Instance);
```
vdye, line 351: Is there any chance of concurrent calls to `GetSettingValues()`? If not, this is fine; otherwise, this would probably need some kind of read/write locking setup.

src/shared/Core/Settings.cs
```diff
@@ -379,26 +403,14 @@ public IEnumerable<string> GetSettingValues(string envarName, string section, st
...
                     foreach (string scope in RemoteUri.GetGitConfigurationScopes())
                     {
                         string queryName = $"{section}.{scope}.{property}";
                         // Look for a scoped entry that includes the scheme "protocol://example.com" first as
                         // this is more specific. If `isPath` is true, then re-get the value from the
                         // `GitConfiguration` with `isPath` specified.
-                        if (configEntries.TryGetValue(queryName, out value) &&
-                            (!isPath || config.TryGet(queryName, isPath, out value)))
+                        if ((isPath && config.TryGet(queryName, true, out value)) ||
+                            _configEntries.TryGetValue(queryName, out value))
```
vdye, line 413: Nice optimization! I also find this condition much easier to parse 🙂

vdye (approved):
Other than one small comment about a comment & a question about concurrency, this looks great to me!

src/shared/Core/Settings.cs
mjcheetham: Right now almost nothing of the GCM shared components are thread-safe.

mjcheetham:
@vdye I found a bug in my changes as you reviewed (whoops!)

```diff
                 if (_configEntries is null)
                 {
                     _configEntries = new Dictionary<string, string>(GitConfigurationKeyComparer.Instance);
-                    config.Enumerate(section, property, entry =>
+                    config.Enumerate(entry =>

                     {
                         _configEntries[entry.Key] = entry.Value;
```

Had only been caching `section.property` values for the first requested setting!
Thank you tests :)

vdye (approved):

PR #1181, opened 2023-04-04
Fix app path name of Windows dropping file extension

mjcheetham:
During some refactoring we accidentially introduced an error in the 'app path' on the CommandContext`; rather than trimming the file extension off (for detection of the old executable name on Windows) on a local copy, we are now updating the app wide copy!

Only mutate a local copy.


ldennington (approved):
🐛 bash!!

PR #1190, opened 2023-04-07
Add support for GitHub enterprise-managed user accounts

mjcheetham:
Add support for GitHub [enterprise-manage users (EMU)](https://docs.github.com/en/enterprise-cloud@latest/admin/identity-and-access-management/using-enterprise-managed-users-for-iam/about-enterprise-managed-users) to the GitHub host provider.

Accounts in an 'EMU' enterprise/business are siloed from the regular, public GitHub.com accounts. EMU accounts are identified by the `_shortcode` suffix, where the `shortcode` is a moniker for the enterprise/business, for example `alice_contoso`.

When asked to recall credentials for the GitHub.com host we now attempt to filter stored accounts by the `shortcode`, given information provided in `WWW-Authenticate` headers from upcoming versions of Git that support these headers (as of https://github.com/gitgitgadget/git/commit/92c56da09683fa3331668adec073b6769da8f0b7).

The format of the header is:

```
WWW-Authenticate: Basic realm="GitHub" [domain_hint="X"] [enterprise_hint="Y"]
```

..where `X` is the shortcode, and `Y` is the enterprise name.

If multiple accounts are available for the given 'domain' then we present an account selection prompt. Users can avoid this prompt in the case of multiple user accounts by specifying the desired account in the remote URL (e.g. `https://alice@github.com/mona/test` to always use the `alice` account).

Note that GitHub.com does not yet return such `WWW-Authenticate` headers, except always `Basic realm="GitHub"`, so this may be subject to fixes later. In the case of `realm="GitHub"`, i.e., public accounts, there is no change.

### Testing

To test the new behaviour before GitHub.com returns such headers, it's possible to fake the server response using [`mitmproxy`](https://mitmproxy.org) and the following script:

```python
"""Add an HTTP header to each response."""

class AddHeader:
    # initialize a dict with shortcodes and paths
    def __init__(self):
        org1 = ("domain1", "enterprise1")
        org2 = ("domain2", "enterprise2")

        self.orgMap = {
            "org1" : enterprise1,
            "org2" : enterprise1,
            "org3" : enterprise2,
        }

    def response(self, flow):
        if flow.response.status_code == 401:
            # lookup the correct shortcode based on the org
            org = flow.request.path.split("/")[1]
            if org not in self.orgMap:
                return
            domain_hint = self.orgMap[org][0]
            enterprise_hint = self.orgMap[org][1]
            # build the header
            header = "Basic realm=\"GitHub\" enterprise_hint=\"" + enterprise_hint + "\" domain_hint=\"" + domain_hint + "\""
            # set the header
            flow.response.headers["WWW-Authenticate"] = header

addons = [
    AddHeader()
]
```

Replace `orgN` with the org names that are backed by an EMU Enterprise, and fill `domainN` for the shortcode, and `enterpriseN` for the enterprise slug/name.

Configure Git to use the proxy and run `mitmproxy` with the `--scripts` argument:

```shell
git config --global http.proxy 'http://127.0.0.1:8080'
mitmproxy --scripts <SCRIPT>
```

Now all Git interactions that touch `orgN` will include the `domain_hint` and `enterprise_hint`s as defined.

I use these two helpful aliases to quickly add and remove the local proxy from Git's config:

```shell
[alias]
	mitm = "!f(){ git config --global http.proxy 'http://127.0.0.1:8080'; }; f"
	unmitm = "!f(){ git config --global --unset http.proxy; }; f"
```

### DIFF
--- a/src/shared/GitHub/GitHubConstants.cs
+++ b/src/shared/GitHub/GitHubConstants.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 
 namespace GitHub
 {
@@ -18,6 +19,11 @@ namespace GitHub
         public static readonly Uri OAuthAuthorizationEndpointRelativeUri = new Uri("/login/oauth/authorize", UriKind.Relative);
         public static readonly Uri OAuthTokenEndpointRelativeUri = new Uri("/login/oauth/access_token", UriKind.Relative);
         public static readonly Uri OAuthDeviceEndpointRelativeUri = new Uri("/login/device/code", UriKind.Relative);
+
+        /// <summary>
+        /// GitHub user names that contain underscores but are not EMU logins.
+        /// </summary>
+        public static readonly IReadOnlyList<string> InvalidUnderscoreLogins = new[] { "pj_nitin", "up_the_irons" };
 
         /// <summary>
         /// The GitHub required HTTP accepts header value
@@ -59,6 +65,7 @@ namespace GitHub
             public const string DevOAuthClientId = "GCM_DEV_GITHUB_CLIENTID";
             public const string DevOAuthClientSecret = "GCM_DEV_GITHUB_CLIENTSECRET";
             public const string DevOAuthRedirectUri = "GCM_DEV_GITHUB_REDIRECTURI";
+            public const string AccountFiltering = "GCM_GITHUB_ACCOUNTFILTERING";
         }
 
         public static class GitConfiguration
@@ -70,6 +77,7 @@ namespace GitHub
                 public const string DevOAuthClientId = "gitHubDevClientId";
                 public const string DevOAuthClientSecret = "gitHubDevClientSecret";
                 public const string DevOAuthRedirectUri = "gitHubDevRedirectUri";
+                public const string AccountFiltering = "githubAccountFiltering";
             }
         }
     }
--- a/src/shared/GitHub/GitHubHostProvider.cs
+++ b/src/shared/GitHub/GitHubHostProvider.cs
@@ -135,6 +135,7 @@ namespace GitHub
             // are multiple options.
             string userName = input.UserName;
             bool addAccount = false;
+            bool filtered = false;
             if (string.IsNullOrWhiteSpace(userName))
             {
                 IList<string> accounts = _context.CredentialStore.GetAccounts(service);
@@ -143,6 +144,8 @@ namespace GitHub
                 {
                     _context.Trace.WriteLine($"  {account}");
                 }
+
+                filtered = FilterAccounts(remoteUri, input.WwwAuth, ref accounts);
 
                 switch (accounts.Count)
                 {
@@ -159,15 +162,15 @@ namespace GitHub
                 }
             }
 
-            // Always try and locate an existing credential in the OS credential store
-            // unless we're being told to explicitly add a new account. If the account lookup
-            // failed above we should still try to lookup an existing credential.
+            // Always try and locate an existing credential in the OS credential store unless we're being
+            // told to explicitly add a new account OR have specifically filtered out irrelevant accounts.
+            // If the account lookup failed for another reason we should still try to lookup an existing credential.
             ICredential credential = null;
             if (addAccount)
             {
                 _context.Trace.WriteLine("Adding a new account!");
             }
-            else
+            else if (!string.IsNullOrWhiteSpace(userName) || !filtered)
             {
                 _context.Trace.WriteLine($"Looking for existing credential in store with service={service} account={userName}...");
                 credential = _context.CredentialStore.Get(service, userName);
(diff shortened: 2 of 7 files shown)


src/shared/GitHub/GitHubConstants.cs
```diff
@@ -19,6 +20,11 @@ public static class GitHubConstants
         public static readonly Uri OAuthTokenEndpointRelativeUri = new Uri("/login/oauth/access_token", UriKind.Relative);
         public static readonly Uri OAuthDeviceEndpointRelativeUri = new Uri("/login/device/code", UriKind.Relative);
 
+        /// <summary>
+        /// GitHub user names that contain underscores but are not EMU logins.
+        /// </summary>
+        public static readonly IReadOnlyList<string> InvalidUnderscoreLogins = new[] { "pj_nitin", "up_the_irons" };
```
ldennington, line 26: Are these the only users on all of GH that have underscores in their names?

src/shared/GitHub/GitHubHostProvider.cs
```diff
@@ -145,24 +145,42 @@ public async Task<ICredential> GetCredentialAsync(InputArguments input)
...
-                GitHubAuthChallenge authChallenge = GitHubAuthChallenge.FromHeaders(input.WwwAuth).FirstOrDefault();
-                if (authChallenge is not null)
+                // Allow the user to disable account filtering until this feature stabilises.
+                // Default to enabled.
+                bool enableFiltering = !_context.Settings.TryGetSetting(
+                    GitHubConstants.EnvironmentVariables.AccountFiltering,
+                    Constants.GitConfiguration.Credential.SectionName,
+                    GitHubConstants.GitConfiguration.Credential.AccountFiltering,
+                    out string enableFilteringStr
+                ) || enableFilteringStr.ToBooleanyOrDefault(true);
```
ldennington, line 155: I'm a bit confused by this syntax. When will this `ToBooleanyOrDefault` only be invoked if `TryGetSetting` returns `true`?

ldennington (approved):
Looks great! So excited this is finally happening! Just had a couple minor questions.

src/shared/GitHub/GitHubHostProvider.cs
mjcheetham: If we expand this to an `if` statement, it maybe becomes a little clearer:

```csharp
bool enableFiltering;
if (_context.Settings.TryGetSetting(
                    GitHubConstants.EnvironmentVariables.AccountFiltering,
                    Constants.GitConfiguration.Credential.SectionName,
                    GitHubConstants.GitConfiguration.Credential.AccountFiltering,
                    out string enableFilteringStr))
{
    bool? enableBooleany = enableFilteringStr.ToBooleany();
    if (enableBooleany.HasValue)
    {
        // Take the valid setting value as the answer
        enableFiltering = enableBooleany.Value;
    }
    else
    {
        // Default to `true` if the setting existed but wasn't a boolean
        enableFiltering = true;
    }
}
else
{
    // Default to `true` if no setting is present
    enableFiltering = true;
}
```

An abridged version could also be represented like:
```csharp
bool enable = !TryGetSetting(.., out string val) || (val.ToBool?() ?? true)
```

Enabled if no settings, OR if the value wasn't a bool. Otherwise, use the value.

mjcheetham: This is a common pattern we're using in the `Settings` class. Example: https://github.com/git-ecosystem/git-credential-manager/blob/0ae7f478ef4e7871a9c15086d1051654c5ba14d2/src/shared/Core/Settings.cs#L605-L606

src/shared/GitHub/GitHubHostProvider.cs
```diff
@@ -183,6 +193,53 @@ public async Task<ICredential> GetCredentialAsync(InputArguments input)
             return credential;
         }
 
+        private bool FilterAccounts(Uri remoteUri, IEnumerable<string> wwwAuth, ref IList<string> accounts)
+        {
+            if (!IsGitHubDotCom(remoteUri))
+            {
+                _context.Trace.WriteLine("No account filtering outside of GitHub.com.");
```
tsvietOK, line 200: Should everything below be skipped if remoteUri is not a github.com by returning a false? Like, in case there is a link to the GHES.

matthewarkin:
Sorry to bump an old topic, but I'm curious if there's been any plans for Github.com to support the enterprise/domain hints that have been implemented here?

mjcheetham:
> Sorry to bump an old topic, but I'm curious if there's been any plans for Github.com to support the enterprise/domain hints that have been implemented here?

Hello! 👋 When youn say "plans for Github.com to support the .. hints" I'm not sure what you mean? GitHub.com will return `domain_hint` and `enterprise_hint` for EMU repos in the `WWW-Authenticate` header.

Do you mean hints for non-EMU/public repos?

Issue #1143, opened 2023-03-08
dependencies: update to latest versions


ldennington:
GCM has several key dependencies that have become outdated. Before our next
release (slated just ahead of Git's 2.41 release cycle), we should update these
dependencies to the latest stable versions available. These include:

- Avalonia
  - [Repo](https://github.com/AvaloniaUI/Avalonia)
  - [NuGet package](https://www.nuget.org/packages/Avalonia/)
- MSAL
  - [Repo](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet)
  - [NuGet packages](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet#nuget-package)

ldennington:
MSAL was updated with #1191 and Avalonia was updated with #1207

PR #1191, opened 2023-04-07
Update MSAL libraries and use new MSALRuntime-based broker implementation

mjcheetham:
Use the new Windows broker which is based on the MSALRuntime; an export wrapper around a native, cross-platform MSAL library.

In this new set up, we drop the `.Desktop` package in favour of the `.Broker` package that also means we drop the WebView2Loader.dll, which we didn't make use of anyway. There are a few new binaries to be distributed in the new model, including a P/Invoke layer, IdentityModel abstractions library, and the native `msalruntime_x86.dll`.

Note that GCM still only support x86 on Windows, and only supports broker use on Windows. For this reason we don't bother adding the broker package on non-.NET Framework builds to keep the sizes on Mac/Linux to a minimum.

Also update the MSAL extensions library whilst we are here to pick up various bug fixes, and drop workarounds that are no longer required for this new broker implementation!

### DIFF
--- a/src/shared/Core/Authentication/MicrosoftAuthentication.cs
+++ b/src/shared/Core/Authentication/MicrosoftAuthentication.cs
@@ -7,7 +7,7 @@ using Microsoft.Identity.Client;
 using Microsoft.Identity.Client.Extensions.Msal;
 
 #if NETFRAMEWORK
-using Microsoft.Identity.Client.Desktop;
+using Microsoft.Identity.Client.Broker;
 #endif
 
 namespace GitCredentialManager.Authentication
@@ -41,55 +41,6 @@ namespace GitCredentialManager.Authentication
             "live", "liveconnect", "liveid",
         };
 
-        #region Broker Initialization
-
-        public static bool IsBrokerInitialized { get; private set; }
-
-        public static void InitializeBroker()
-        {
-            if (IsBrokerInitialized)
-            {
-                return;
-            }
-
-            IsBrokerInitialized = true;
-
-            // Broker is only supported on Windows 10 and later
-            if (!PlatformUtils.IsWindowsBrokerSupported())
-            {
-                return;
-            }
-
-            // Nothing to do when not an elevated user
-            if (!PlatformUtils.IsElevatedUser())
-            {
-                return;
-            }
-
-            // Lower COM security so that MSAL can make the calls to WAM
-            int result = Interop.Windows.Native.Ole32.CoInitializeSecurity(
-                IntPtr.Zero,
-                -1,
-                IntPtr.Zero,
-                IntPtr.Zero,
-                Interop.Windows.Native.Ole32.RpcAuthnLevel.None,
-                Interop.Windows.Native.Ole32.RpcImpLevel.Impersonate,
-                IntPtr.Zero,
-                Interop.Windows.Native.Ole32.EoAuthnCap.None,
-                IntPtr.Zero
-            );
-
-            if (result != 0)
-            {
-                throw new Exception(
-                    $"Failed to set COM process security to allow Windows broker from an elevated process (0x{result:x})." +
-                    Environment.NewLine +
-                    $"See {Constants.HelpUrls.GcmWamComSecurity} for more information.");
-            }
-        }
-
-        #endregion
-
         public MicrosoftAuthentication(ICommandContext context)
             : base(context) { }
 
@@ -99,17 +50,10 @@ namespace GitCredentialManager.Authentication
             string authority, string clientId, Uri redirectUri, string[] scopes, string userName)
         {
             // Check if we can and should use OS broker authentication
-            bool useBroker = false;
-            if (CanUseBroker(Context))
-            {
-                // Can only use the broker if it has been initialized
-                useBroker = IsBrokerInitialized;
-
-                if (IsBrokerInitialized)
-                    Context.Trace.WriteLine("OS broker is available and enabled.");
-                else
-                    Context.Trace.WriteLine("OS broker has not been initialized and cannot not be used.");
-            }
+            bool useBroker = CanUseBroker();
+            Context.Trace.WriteLine(useBroker
+                ? "OS broker is available and enabled."
+                : "OS broker has not been initialized and cannot not be used.");
 
             // Create the public client application for authentication
             IPublicClientApplication app = await CreatePublicClientApplicationAsync(authority, clientId, redirectUri, useBroker);
@@ -287,12 +231,19 @@ namespace GitCredentialManager.Authentication
                 appBuilder.WithParentActivityOrWindow(() => new IntPtr(hWndInt));
             }
 
-            // On Windows 10+ & .NET Framework try and use the WAM broker
-            if (enableBroker && PlatformUtils.IsWindowsBrokerSupported())
+            // Configure the broker if enabled
+            // Currently only supported on Windows so only included in the .NET Framework builds
+            // to save on the distribution size of the .NET builds (no need for MSALRuntime bits).
+            if (enableBroker)
             {
 #if NETFRAMEWORK
-                appBuilder.WithExperimentalFeatures();
-                appBuilder.WithWindowsBroker();
+                appBuilder.WithBroker(
+                    new BrokerOptions(BrokerOptions.OperatingSystems.Windows)
+                    {
+                        Title = "Git Credential Manager",
+                        MsaPassthrough = true,
+                    }
+                );
 #endif
             }
 
@@ -458,11 +409,11 @@ namespace GitCredentialManager.Authentication
 
         #region Auth flow capability detection
 
-        public static bool CanUseBroker(ICommandContext context)
+        public bool CanUseBroker()
         {
 #if NETFRAMEWORK
             // We only support the broker on Windows 10+ and in an interactive session
-            if (!context.SessionManager.IsDesktopSession || !PlatformUtils.IsWindowsBrokerSupported())
+            if (!Context.SessionManager.IsDesktopSession || !PlatformUtils.IsWindowsBrokerSupported())
             {
                 return false;
             }
@@ -470,7 +421,7 @@ namespace GitCredentialManager.Authentication
             // Default to not using the OS broker
             const bool defaultValue = false;
 
-            if (context.Settings.TryGetSetting(Constants.EnvironmentVariables.MsAuthUseBroker,
+            if (Context.Settings.TryGetSetting(Constants.EnvironmentVariables.MsAuthUseBroker,
                     Constants.GitConfiguration.Credential.SectionName,
                     Constants.GitConfiguration.Credential.MsAuthUseBroker,
                     out string valueStr))
--- a/src/shared/Core/Authentication/MicrosoftAuthentication.cs
+++ b/src/shared/Core/Authentication/MicrosoftAuthentication.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Net.Http;
 using System.Threading.Tasks;
+using GitCredentialManager.Interop.Windows.Native;
 using Microsoft.Identity.Client;
 using Microsoft.Identity.Client.Extensions.Msal;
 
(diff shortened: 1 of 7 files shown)


mjcheetham:
@bgavrilMS could a member of the MSAL team have a quick glance over our usage of the new broker implementation on Windows, to make sure we're not doing something wrong? Thanks!

src/shared/Core/Authentication/MicrosoftAuthentication.cs
```diff
@@ -41,55 +42,6 @@ public class MicrosoftAuthentication : AuthenticationBase, IMicrosoftAuthenticat
             "live", "liveconnect", "liveid",
         };
 
-        #region Broker Initialization
```
bgavrilMS, line 44: deleting workounds, so satisfying!

src/shared/Core/Authentication/MicrosoftAuthentication.cs
```diff
@@ -279,20 +224,40 @@ private async Task<AuthenticationResult> GetAccessTokenSilentlyAsync(IPublicClie
...
+            // so that they are displayed as a child of the correct window.
+            if (PlatformUtils.IsWindows())
             {
-                appBuilder.WithParentActivityOrWindow(() => new IntPtr(hWndInt));
+                // If we have a parent window ID then use that, otherwise use the hosting terminal window.
+                IntPtr parentHandle;
+                if (!string.IsNullOrWhiteSpace(Context.Settings.ParentWindowId) &&
+                    int.TryParse(Context.Settings.ParentWindowId, out int hWndInt) && hWndInt > 0)
+                {
+                    parentHandle = new IntPtr(hWndInt);
```
bgavrilMS, line 236: How are you calculating the `hwndInt` ? For console apps it's a bit trickier - see https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/wam#parent-window-handles

src/shared/Core/Authentication/MicrosoftAuthentication.cs
```diff
@@ -279,20 +224,40 @@ private async Task<AuthenticationResult> GetAccessTokenSilentlyAsync(IPublicClie
...
+                }
+                else
+                {
+                    IntPtr consoleHandle = Kernel32.GetConsoleWindow();
+                    parentHandle = User32.GetAncestor(consoleHandle, GetAncestorFlags.GetRootOwner);
```
bgavrilMS, line 241: +1

bgavrilMS (approved):

src/shared/Core/Core.csproj
```diff
@@ -13,13 +13,13 @@
   <ItemGroup Condition="'$(TargetFramework)' == 'net472'">
     <Reference Include="System.Net.Http" />
     <Reference Include="System.Web" />
-    <PackageReference Include="Microsoft.Identity.Client.Desktop" Version="4.37.0" />
+    <PackageReference Include="Microsoft.Identity.Client.Desktop" Version="4.52.0" />
```
ldennington: This means one less thing to worry about with #1143. Thanks!

src/windows/Installer.Windows/Setup.iss
```diff
@@ -125,20 +125,19 @@ Source: "{#PayloadDir}\Microsoft.AzureRepos.dll";                       DestDir:
 Source: "{#PayloadDir}\gcmcore.dll";                                    DestDir: "{app}"; Flags: ignoreversion
 Source: "{#PayloadDir}\gcmcoreui.dll";                                  DestDir: "{app}"; Flags: ignoreversion
 Source: "{#PayloadDir}\gcmcoreuiwpf.dll";                               DestDir: "{app}"; Flags: ignoreversion
-Source: "{#PayloadDir}\Microsoft.Identity.Client.Desktop.dll";          DestDir: "{app}"; Flags: ignoreversion
+Source: "{#PayloadDir}\Microsoft.Identity.Client.Broker.dll";           DestDir: "{app}"; Flags: ignoreversion
```
ldennington, line 128: Nit: Is the extra space at the end on purpose?

src/windows/Installer.Windows/Setup.iss
```diff
@@ -125,20 +125,19 @@ Source: "{#PayloadDir}\Microsoft.AzureRepos.dll";                       DestDir:
...
 Source: "{#PayloadDir}\Microsoft.Identity.Client.dll";                  DestDir: "{app}"; Flags: ignoreversion
 Source: "{#PayloadDir}\Microsoft.Identity.Client.Extensions.Msal.dll";  DestDir: "{app}"; Flags: ignoreversion
-Source: "{#PayloadDir}\Microsoft.Web.WebView2.Core.dll";                DestDir: "{app}"; Flags: ignoreversion
-Source: "{#PayloadDir}\Microsoft.Web.WebView2.WinForms.dll";            DestDir: "{app}"; Flags: ignoreversion
-Source: "{#PayloadDir}\Microsoft.Web.WebView2.Wpf.dll";                 DestDir: "{app}"; Flags: ignoreversion
+Source: "{#PayloadDir}\Microsoft.Identity.Client.NativeInterop.dll";    DestDir: "{app}"; Flags: ignoreversion
+Source: "{#PayloadDir}\Microsoft.IdentityModel.Abstractions.dll";       DestDir: "{app}"; Flags: ignoreversion
+Source: "{#PayloadDir}\msalruntime_x86.dll";                            DestDir: "{app}"; Flags: ignoreversion
```
ldennington, line 133: Nit: extra spaces

src/shared/Core/Authentication/MicrosoftAuthentication.cs
```diff
@@ -223,12 +224,25 @@ private async Task<AuthenticationResult> GetAccessTokenSilentlyAsync(IPublicClie
...
```
ldennington, line 241: Do we need NULL checks here?

ldennington (approved):
Looks great! This allowed for a lot of nice cleanup.

src/windows/Installer.Windows/Setup.iss
mjcheetham: Yes, to keep the alignment of `DestDir: ...`.

src/windows/Installer.Windows/Setup.iss
mjcheetham: Needed to keep alignment.

ldennington: Ah I see - thanks for clarifying.

src/shared/Core/Authentication/MicrosoftAuthentication.cs
mjcheetham: `GetConsoleWindow` will [return `NULL`](https://learn.microsoft.com/en-us/windows/console/getconsolewindow#return-value) (`IntPtr.Zero`) if there is no console window available.

The [docs for `GetAncestor`](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getancestor) don't actually say what happens if you pass `NULL` as the `hwnd` argument. 🤔

However, trying it with `NULL` I get `NULL` back again (`IntPtr.Zero`). Using `IntPtr.Zero` with `WithParentActivityOrWindow` in MSAL results in an error since a window is absolutely required (for WAM).

Having no console window is totally possible, so I guess at this point all we can do is grab a handle to the desktop object?

cc: @bgavrilMS

mjcheetham: Added a new commit that, when the broker is enabled and we are unable to find a parent window, shows a 'dummy' top-level progress bar window, and uses that window handle as the parent.

wam_dummy_window

bgavrilMS: Ideally we should get this type of functionality into MSAL. Apps should need to reference Win.Forms and create windows. Will follow up offline.

bgavrilMS: Note that this solution is imperfect, for example if multiple monitors are used.

bgavrilMS: Another way of doing it is to fail yourself, i.e. if you are hosted in a UI application and the UI application does not initialize GCM correctly with its window handle - fail. If you are hosted in a console app, then the above logic should "just work"

src/shared/Core/Authentication/MicrosoftAuthentication.cs
```diff
@@ -229,20 +247,36 @@ private async Task<AuthenticationResult> GetAccessTokenSilentlyAsync(IPublicClie
...
+                    else if (enableBroker) // Only actually need to set a parent window when using the Windows broker
+                    {
+#if NETFRAMEWORK
+                        Context.Trace.WriteLine($"Using dummy parent window for MSAL authentication dialogs.");
+                        _dummyWindow = new DummyWindow();
+                        appBuilder.WithParentActivityOrWindow(_dummyWindow.ShowAndGetHandle);
+#else
+                        Context.Trace.WriteLine($"Using desktop parent window for MSAL authentication dialogs.");
+                        parentHandle = User32.GetDesktopWindow();
+#endif
```
ldennington, line 277: Is this alternative needed? Since we only need to do this if we're using the broker and the broker is only available on Windows, wouldn't we always be on .NET Framework?

I know there has been some talk with Bogdan about the macOS broker, though, so if you've added it in anticipation of that, then that's fine!

ldennington (approved):
Latest commit looks good, just a few small comments. Overall, though, I think this solution is all right until (hopefully) the issue is handled in MSAL.

src/shared/Core/Authentication/MicrosoftAuthentication.cs
mjcheetham: You are correct that in most cases you'd be using .NET Framework on Windows, however if you're using the .NET Tool on Windows then that's using .NET 6... but then again we'd never enable the broker.. so yes.. I'll remove this 😃

src/shared/Core/Authentication/MicrosoftAuthentication.cs
```diff
@@ -54,106 +62,116 @@ public MicrosoftAuthentication(ICommandContext context)
...
+                // On Windows 10 & .NET Framework, MSAL supports the Web Account Manager (WAM) broker - we try to use that if possible
+                // in the first instance.
+                //
+                // On .NET Framework MSAL supports the WinForms based 'embedded' webview UI. For Windows + .NET Framework this is the
+                // best and natural experience.
+                //
+                // On other runtimes (e.g., .NET Core) MSAL only supports the system webview flow (launch the user's browser),
+                // and the device-code flows.
+                //
+                //     Note: .NET Core 3 allows using WinForms when run on Windows but MSAL does not yet support this.
```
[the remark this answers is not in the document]
mjcheetham: Updated to drop this reference.

src/shared/Core/Authentication/MicrosoftAuthentication.cs
```diff
@@ -54,106 +62,116 @@ public MicrosoftAuthentication(ICommandContext context)
...
+                // we need to prompt the user for credentials.
+                //
+                // If the user has expressed a preference in how the want to perform the interactive authentication flows then we respect that.
+                // Otherwise, depending on the current platform and session type we try to show the most appropriate authentication interface:
+                //
+                // On Windows 10 & .NET Framework, MSAL supports the Web Account Manager (WAM) broker - we try to use that if possible
+                // in the first instance.
+                //
+                // On .NET Framework MSAL supports the WinForms based 'embedded' webview UI. For Windows + .NET Framework this is the
+                // best and natural experience.
```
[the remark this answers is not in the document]
mjcheetham: We did, before broker (when enabled). Reworded this entire block to be clearer.

src/shared/Core/Authentication/MicrosoftAuthentication.cs
mjcheetham: Since this scenario (no console window and not being told the parent window explicitly) is likely the default for most versions of Visual Studio and tools like SourceTree etc, we're going to need to support this out-of-the box.

I think implementing the 'dummy' progress dialog window should suffice.

mjcheetham:
@ldennington I've updated that outdated block comment in _MicrosoftAuthentication.cs`; did you want to have another :eyes: over it? Thanks!

ldennington (approved):
Comment updates look good - thanks for doing that!

PR #1196, opened 2023-04-11
Catch OAuth refresh token errors for generic provider

mjcheetham:
In case we have an existing refresh token, but it has expired or been revoked, we should gracefully fall back to performing an interactive OAuth flow.


ldennington (approved):
Looks good!

PR #1197, opened 2023-04-11
Add support for using the current Windows user for WAM on DevBox

### LINKED ISSUE #917: Enable SSO capabilities with the Windows signed in user while on Dev Box
As part of broad push towards optimizing the developer experience for the new [Microsoft Dev Box ](https://azure.microsoft.com/en-us/services/dev-box/#overview)service, we'd like to make sure the Azure Dev Ops (ADO) identity experience for GCM is optimal while on Dev Box. From that standpoint, we'd like to work with you to minimize ADO sign in prompts by potentially enabling SSO capabilities with the Windows signed in user while on Dev Box.

For example, today whenever a user attempts to clone an ADO repo from the Visual Studio (VS) UI, GCM prompts for credentials for each and every repo regardless of the identities available on the VS keychain or even selected from the clone's dialog account picker.

While the VS team already has feedback about the overall UX model, right now we'd like to primarily focus on the Dev Box experience as that scenario has potentially overall less friction and could be streamlined based on the signed in Windows account via the WAM capabilities currently available for GCM.

Happy to share more context as needed and/or schedule a meeting to discuss further.

mjcheetham:
Add the ability to configure MSAL to use the default OS account when the broker is enabled.
Also detect when we are in a Microsoft Dev Box environment, and if we are, then default to enabling the new setting (and enable WAM).

Show a confirmation prompt before continuing to use the current OS account, which is similar to how Microsoft Teams operates.

windows-defaultaccount

Left: Avalonia UI, Right: fallback WPF window

Fixes #917

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -543,7 +543,10 @@ git config --global credential.msauthFlow devicecode
 
 Use the operating system account manager where available.
 
-Defaults to `false`. This default is subject to change in the future.
+Defaults to `false`. In certain cloud hosted environments when using a work or
+school account, such as [Microsoft DevBox][devbox], the default is `true`.
+
+These defaults are subject to change in the future.
 
 _**Note:** before you enable this option on Windows, please review the
 [Windows Broker][wam] details for what this means to your local Windows user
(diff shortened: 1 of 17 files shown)


ldennington (approved):
This is awesome, Matthew! Aside from the initial wait time, the Dev Box stuff is pretty cool, and it's very slick that the broker/default account functionality are used automatically there. I tested various GUI and terminal scenarios, and all seems to be in order.

nyila (commented):
Fixes #

PR #1228, opened 2023-04-27
Add Git configuration options for trace & debug

mjcheetham:
Add options for enabling tracing and debugging in Git configuration. Today the below options were only supported via environment variable:

 - `GCM_TRACE`
 - `GCM_TRACE_SECRETS`
 - `GCM_TRACE_MSAUTH`
 - `GCM_DEBUG`

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -76,6 +76,86 @@ Defaults to enabled.
 
 ---
 
+### credential.trace
+
+Enables trace logging of all activities.
+Configuring Git and GCM to trace to the same location is often desirable, and
+GCM is compatible and cooperative with `GIT_TRACE`.
+
+#### Example
+
+```shell
+git config --global credential.trace /tmp/git.log
+```
+
+If the value of `credential.trace` is a full path to a file in an existing
+directory, logs are appended to the file.
+
+If the value of `credential.trace` is `true` or `1`, logs are written to
+standard error.
+
+Defaults to disabled.
+
+**Also see: [GCM_TRACE][gcm-trace]**
+
+---
+
+### credential.traceSecrets
+
+Enables tracing of secret and sensitive information, which is by default masked
+in trace output. Requires that `credential.trace` is also enabled.
+
+#### Example
+
+```shell
+git config --global credential.traceSecrets true
+```
+
+If the value of `credential.traceSecrets` is `true` or `1`, trace logs will include
+secret information.
+
+Defaults to disabled.
+
+**Also see: [GCM_TRACE_SECRETS][gcm-trace-secrets]**
+
+---
+
+### credential.traceMsAuth
+
+Enables inclusion of Microsoft Authentication library (MSAL) logs in GCM trace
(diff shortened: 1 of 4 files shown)


docs/configuration.md
```diff
@@ -76,6 +76,86 @@ Defaults to enabled.
...
+git config --global credential.trace /tmp/git.log
+```
+
+If the value of `credential.trace` is a full path to a file in an existing
+directory, logs are appended to the file.
+
+If the value of `credential.trace` is `true` or `1`, logs are written to
+standard error.
+
+Defaults to tracing disabled.
```
ldennington, line 97: Nit: maybe just say this to align with the other updates:

```suggestion
Defaults to disabled.
```

docs/configuration.md
```diff
@@ -76,6 +76,86 @@ Defaults to enabled.
...
+
+**Also see: [GCM_TRACE][gcm-trace]**
```
ldennington, line 99: Links are resulting in 404s. Will be resolved once merged?

src/shared/Core/Settings.cs
```diff
@@ -536,7 +540,11 @@ public bool IsInteractionAllowed
             }
         }
 
-        public bool GetTracingEnabled(out string value) => _environment.Variables.TryGetValue(KnownEnvars.GcmTrace, out value) && !value.IsFalsey();
+        public bool GetTracingEnabled(out string value) =>
+            TryGetSetting(KnownEnvars.GcmTrace,
+                KnownGitCfg.Credential.SectionName,
+                KnownGitCfg.Credential.Trace,
+                out value) && !value.IsFalsey();
```
ldennington, line 547: Maybe make this `IsTruthy()` to align with the `IsDebuggingEnabled` check above and the `IsSecretTracingEnabled`/`IsMsalTracingEnabled` checks below.

ldennington (approved):
Looks good! Just a few nitpicky suggestions - the only main concern is the 404s on the links, but I suspect that's just an issue with GitHub's PR Files view.

src/shared/Core/Settings.cs
mjcheetham: We need to keep this to `!IsFalsey` because we can set this variable to something like a file path which is considered "enabled".

Value|`TryGetSetting`|`IsTruthy`|`IsFalsey`|`TryGetSetting && !IsFalsey`
-|-|-|-|-
`null`|❌|❌|❌|❌
`"/tmp/file.txt"`|✅|❌|❌|✅
`true`/`1`/`"yes"`/`"on"`|✅|✅|❌|✅
`false`/`0`/`"no"`/`"off"`|✅|❌|✅|❌

docs/configuration.md
mjcheetham: Hmm.. they are correct.. so I assume some issue with the PR view.

src/shared/Core/Settings.cs
ldennington: Lovely truth table - explains very nicely, thank you!

PR #1238, opened 2023-05-03
Fix several UI bugs in 2.1 pre-release

mjcheetham:
Fix a few bugs that were spotted during testing of the 2.1 pre-release:

- Missing Atlassian logo resource from the compiled binaries
- Remote URL is shown even for Bitbucket cloud with in-proc GUI
- In-proc basic auth GUI user name field should only be set if non-null or whitespace

Also bump the VERSION to 2.1.1 in the expectation for this to replace 2.1.0.


ldennington (approved):
Thanks for the fixes! 🌟

PR #1239, opened 2023-05-03
Lazily initalise the HttpClient for the Bitbucket OAuth2ClientRegistry

mjcheetham:
Lazily acquire the `HttpClient` instance from the factory inside of the `OAuth2ClientRegistry`. There is a relatively large cost to creating a new `HttpClient` instance (especially the first; due to config reading) that is being paid each time we just register the entire Bitbucket host provider.

Whilst we are here also cleanup some of the styling/field naming in this class to be consistent with the rest of GCM.

This saves approximately 120ms of startup time.

### DIFF
--- a/src/shared/Atlassian.Bitbucket/OAuth2ClientRegistry.cs
+++ b/src/shared/Atlassian.Bitbucket/OAuth2ClientRegistry.cs
@@ -3,21 +3,18 @@ using GitCredentialManager;
 
 namespace Atlassian.Bitbucket
 {
-    public class OAuth2ClientRegistry : IRegistry<BitbucketOAuth2Client>
+    public class OAuth2ClientRegistry : DisposableObject, IRegistry<BitbucketOAuth2Client>
     {
-        private readonly HttpClient http;
-        private ISettings settings;
-        private readonly ITrace trace;
-        private readonly ITrace2 trace2;
-        private Cloud.BitbucketOAuth2Client cloudClient;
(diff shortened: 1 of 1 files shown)


src/shared/Atlassian.Bitbucket/OAuth2ClientRegistry.cs
```diff
@@ -30,15 +27,20 @@ public BitbucketOAuth2Client Get(InputArguments input)
...
         }
 
-        public void Dispose()
+        protected override void ReleaseManagedResources()
         {
-            http.Dispose();
-            settings.Dispose();
-            cloudClient = null;
-            dataCenterClient = null;
+            _httpClient.Dispose();
```
vdye, line 32: Will this cause an exception if `_httpClient` is `null`?

vdye (approved):
Looks good! I had one question about `ReleaseManagedResources`, hopefully easy to fix or explain either way. 🚀

src/shared/Atlassian.Bitbucket/OAuth2ClientRegistry.cs
mjcheetham: Yes! Thanks for catching. Fixed in next push.

vdye (approved):

mminns:
👍

mminns (commented):
Looks good to me, thanks.

PR #1242, opened 2023-05-04
Add support for Linux ARM

juur:
Add support for Linux builds on ARM64/aarch64. Possibly requires dpkg-architecture installed.
I had to install dpkg-cross on an aarch64 installation of Ubuntu, but after that it compiled working binaries.


ldennington:
Hey @juur! Thanks for your contribution. This is actually something the maintainers are in the process of implementing - we'll be sure to tag you on the full PR when it's available.

Issue #1160, opened 2023-03-23
Support `httpcookieFile`


miya789:
### Feature description

Could you please make GCM read cookie files in libcurl format as well as the Git command[^git-cookie]?
When I issue an OAuth2 token by GCM, the communication to the path /oauth/ is not working well because it is not Git that GCM wraps, but GCM that does not support cookies accesses it.
I want to use this because I need to access the Git server behind the Reverse Proxy with Authenticator for a special reason.

I thought about implementing it myself, but Git is calling curl, while GCM is an HTTPClient connection using C#, and it seemed difficult to implement, so I would appreciate it if you could help me.

[^git-cookie]: https://git-scm.com/docs/git-config#Documentation/git-config.txt-httpcookieFile

ldennington:
Implementing this will require parsing the cookie file format (see [here](https://github.com/curl/curl/blob/master/docs/HTTP-COOKIES.md#cookie-file-format)) and setting the cookies in the `CookieContainer` on the `HttpClient`. We would welcome a contribution from the community though! Hopefully the `help-wanted` label will attract some interest.

mjcheetham:
Closing this issue as the support was added in #1251, and will be available in the next release! :-)

PR #1251, opened 2023-05-13
Add support for `httpCookieFile`

miya789:
## Contents
#1160 
I added code to support [`httpCookieFile`](https://git-scm.com/docs/git-config#Documentation/git-config.txt-httpcookieFile).
If `httpCookieFile` is set in `gitconfig`, gcm will add cookie header to request headers such as requests about OAuth2.

## Note
- This does not support [`http.saveCookies`](https://git-scm.com/docs/git-config#Documentation/git-config.txt-httpsaveCookies) because when it is set, built-in git will update cookie and gcm need not to update.
- If you want git via reverse proxy authentication, you can use this (https://github.com/miya789/git-via-reverse-proxy-authn). The structure image is like this:
  - https://stackoverflow.com/questions/64731848/git-cli-verification-via-saml-sso

### DIFF
--- a/src/shared/Core.Tests/HttpClientFactoryTests.cs
+++ b/src/shared/Core.Tests/HttpClientFactoryTests.cs
@@ -311,6 +311,36 @@ namespace GitCredentialManager.Tests
             fileSystemMock.Verify(fs => fs.FileExists(It.IsAny<string>()), expectBundleChecked ? Times.Once : Times.Never);
         }
 
+        [Theory]
+        [InlineData(null, false, null)]
+        [InlineData("~/.git-cookie", true, "# Netscape HTTP Cookie File\n" +
+                          "# https://curl.haxx.se/rfc/cookie_spec.html\n" +
+                          "# This is a generated file! Do not edit.\n" +
+                          "\n" +
+                          ".example.com\tTRUE\t/\tTRUE\t0\tcookie1\tvalue1\n" +
+                          ".example.com\tTRUE\t/\tTRUE\t0\tcookie2\tvalue2\n" +
+                          "#HttpOnly_.example.com\tTRUE\t/\tTRUE\t0\tcookie3\tvalue3\n")]
+        public void HttpClientFactory_GetClient_SetCookieOnlyIfEnabled(string cookieFilePath, bool expectCookieChecked, string cookieFileContent)
+        {
+            var fileSystemMock = new Mock<IFileSystem>();
+            fileSystemMock.Setup(fs => fs.FileExists(It.IsAny<string>())).Returns(true);
+            if (!string.IsNullOrWhiteSpace(cookieFileContent))
+            {
+                fileSystemMock.Setup(fs => fs.ReadAllText(cookieFilePath)).Returns(cookieFileContent);
+            }
+
+            var settings = new TestSettings()
+            {
+                CustomCookieFilePath = cookieFilePath
+            };
+
+            var factory = new HttpClientFactory(fileSystemMock.Object, Mock.Of<ITrace>(), Mock.Of<ITrace2>(), settings, new TestStandardStreams());
+
+            HttpClient client = factory.CreateClient();
+
+            fileSystemMock.Verify(fs => fs.FileExists(It.IsAny<string>()), expectCookieChecked ? Times.AtLeastOnce : Times.Never);
+        }
+
         private static void AssertDefaultCredentials(ICredentials credentials)
         {
             var netCred = (NetworkCredential) credentials;
--- a/src/shared/Core/CurlCookie.cs
+++ b/src/shared/Core/CurlCookie.cs
@@ -3,7 +3,7 @@ using System.Collections.Generic;
 using System.Net;
 using GitCredentialManager;
 
-namespace CurlCookie
+namespace GitCredentialManager
 {
     public class CurlCookieParser
     {
--- a/src/shared/Core/CurlCookie.cs
+++ b/src/shared/Core/CurlCookie.cs
@@ -38,7 +38,7 @@ namespace GitCredentialManager
                     {
                         domain = domain.TrimStart('.');
                     }
-                    var path = parts[2] == "" ? "/" : parts[2];
+                    var path = string.IsNullOrWhiteSpace(parts[2]) ? "/" : parts[2];
                     var secureOnly = parts[3].Equals("TRUE", StringComparison.OrdinalIgnoreCase);
                     var expires = ParseExpires(parts[4]);
                     var name = parts[5];
--- a/src/shared/Core/CurlCookie.cs
+++ b/src/shared/Core/CurlCookie.cs
@@ -33,7 +33,7 @@ namespace GitCredentialManager
                 if (parts.Length >= 7 && (!parts[0].StartsWith("#") || parts[0].StartsWith(HttpOnlyPrefix)))
                 {
                     var domain = parts[0].StartsWith(HttpOnlyPrefix) ? parts[0].Substring(HttpOnlyPrefix.Length) : parts[0];
-                    var includeSubdomains = parts[1] == "TRUE";
+                    var includeSubdomains = StringComparer.OrdinalIgnoreCase.Equals(parts[1], "TRUE");
                     if (!includeSubdomains)
                     {
                         domain = domain.TrimStart('.');
--- a/src/shared/Core/CurlCookie.cs
+++ b/src/shared/Core/CurlCookie.cs
@@ -1,0 +1,83 @@
+using System;
+using System.Collections.Generic;
+using System.Net;
+using GitCredentialManager;
+
+namespace CurlCookie
+{
+    public class CurlCookieParser
+    {
+        private readonly ITrace _trace;
+
+        public CurlCookieParser(ITrace trace)
+        {
+            _trace = trace;
+        }
+
+        public IList<Cookie> Parse(string content)
+        {
+            if (string.IsNullOrWhiteSpace(content))
+            {
+                return Array.Empty<Cookie>();
+            }
+
+            const string HttpOnlyPrefix = "#HttpOnly_";
+
+            var cookies = new List<Cookie>();
+
+            // Parse the cookie file content
+            var lines = content.Split(new[] { '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries);
+            foreach (var line in lines)
+            {
+                var parts = line.Split(new[] { '\t' }, StringSplitOptions.None);
+                if (parts.Length >= 7 && (!parts[0].StartsWith("#") || parts[0].StartsWith(HttpOnlyPrefix)))
+                {
+                    var domain = parts[0].StartsWith(HttpOnlyPrefix) ? parts[0].Substring(HttpOnlyPrefix.Length) : parts[0];
+                    var includeSubdomains = parts[1] == "TRUE";
+                    if (!includeSubdomains)
+                    {
+                        domain = domain.TrimStart('.');
+                    }
+                    var path = parts[2] == "" ? "/" : parts[2];
+                    var secureOnly = parts[3].Equals("TRUE", StringComparison.OrdinalIgnoreCase);
+                    var expires = ParseExpires(parts[4]);
+                    var name = parts[5];
+                    var value = parts[6];
+
+                    cookies.Add(new Cookie()
+                    {
+                        Domain = domain,
+                        Path = path,
+                        Expires = expires,
+                        HttpOnly = true,
+                        Secure = secureOnly,
+                        Name = name,
+                        Value = value,
+                    });
+                }
+                else
+                {
+                    _trace.WriteLine($"Invalid cookie line: {line}");
+                }
+            }
+
+            return cookies;
+        }
+
+        private static DateTime ParseExpires(string expires)
+        {
+#if NETFRAMEWORK
+            DateTime epoch = new DateTime(1970, 01, 01, 0, 0, 0, DateTimeKind.Utc);
+#else
+            DateTime epoch = DateTime.UnixEpoch;
+#endif
+
+            if (long.TryParse(expires, out long i))
+            {
+                return epoch.AddSeconds(i);
+            }
+
+            return epoch;
+        }
+    }
+}
--- a/src/shared/Core/HttpClientFactory.cs
+++ b/src/shared/Core/HttpClientFactory.cs
@@ -197,32 +197,25 @@ namespace GitCredentialManager
 #endif
             }
 
-            _trace.WriteLine($"Custom cookie file has been enabled with {_settings.CustomCookieFilePath}");
             // If IsUsingCookieEnabled is enabled, set Cookie header from cookie file, which is written by libcurl
-            if (!string.IsNullOrWhiteSpace(_settings.CustomCookieFilePath))
+            if (!string.IsNullOrWhiteSpace(_settings.CustomCookieFilePath) && _fileSystem.FileExists(_settings.CustomCookieFilePath))
             {
                 // get the filename from gitconfig
                 string cookieFilePath = _settings.CustomCookieFilePath;
                 _trace.WriteLine($"Custom cookie file has been enabled with {cookieFilePath}");
-                
-                // Throw exception if cookie file not found
-                if (!_fileSystem.FileExists(cookieFilePath) || string.IsNullOrEmpty(_fileSystem.ReadAllText(cookieFilePath)))
-                {
-                    var format = "Custom cookie file not found at path: {0}";
-                    var message = string.Format(format, cookieFilePath);
-                    throw new Trace2FileNotFoundException(_trace2, message, format, cookieFilePath);
-                }
 
                 // get cookie from cookie file
+                string cookieFileContents = _fileSystem.ReadAllText(cookieFilePath);
+
                 var cookieParser = new CurlCookieParser(_trace);
-                var cookies = cookieParser.Parse(_fileSystem.ReadAllText(cookieFilePath));
+                var cookies = cookieParser.Parse(cookieFileContents);
 
                 // Set the cookie
                 var cookieContainer = new CookieContainer();
                 foreach (var cookie in cookies)
                 {
                     var schema = cookie.Secure ? "https" : "http";
-                    var uri = new UriBuilder(schema, cookie.Domain).Uri;
+                    var uri = new UriBuilder(schema, cookie.Domain.TrimStart('.')).Uri;
                     cookieContainer.Add(uri, new Cookie(cookie.Name, cookie.Value));
                 }
                 handler.CookieContainer = cookieContainer;
@@ -347,35 +340,48 @@ namespace GitCredentialManager
 
             public IList<Cookie> Parse(string content)
             {
+                if (string.IsNullOrWhiteSpace(content))
+                {
+                    return Array.Empty<Cookie>();
+                }
+
+                const string HttpOnlyPrefix = "#HttpOnly_";
+
                 var cookies = new List<Cookie>();
 
                 // Parse the cookie file content
                 var lines = content.Split(new[] { '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries);
-                for (int i = 3; i < lines.Length; i++)
+                foreach (var line in lines)
                 {
-                    var parts = lines[i].Split(new[] { '\t' }, StringSplitOptions.RemoveEmptyEntries);
-                    if (parts.Length >= 7)
+                    var parts = line.Split(new[] { '\t' }, StringSplitOptions.None);
+                    if (parts.Length >= 7 && (!parts[0].StartsWith("#") || parts[0].StartsWith(HttpOnlyPrefix)))
                     {
-                        var domain = parts[0];
-                        domain = domain.StartsWith("#HttpOnly_") ? parts[0].Substring("#HttpOnly_".Length) : parts[0];
-                        domain = domain.TrimStart('.');
+                        var domain = parts[0].StartsWith(HttpOnlyPrefix) ? parts[0].Substring(HttpOnlyPrefix.Length) : parts[0];
                         var includeSubdomains = parts[1] == "TRUE";
-                        var path = parts[2];
-                        var secureOnly = parts[3] == "TRUE";
-                        var expires = parts[4];
+                        if (!includeSubdomains)
+                        {
+                            domain = domain.TrimStart('.');
+                        }
+                        var path = parts[2] == "" ? "/" : parts[2];
+                        var secureOnly = parts[3].Equals("TRUE", StringComparison.OrdinalIgnoreCase);
+                        var expires = ParseExpires(parts[4]);
                         var name = parts[5];
                         var value = parts[6];
 
-                        cookies.Add(new Cookie(name, value, path, domain)
+                        cookies.Add(new Cookie()
                         {
-                            Expires = ParseExpires(expires),
+                            Domain = domain,
+                            Path = path,
+                            Expires = expires,
+                            HttpOnly = true,
                             Secure = secureOnly,
-                            HttpOnly = true
+                            Name = name,
+                            Value = value,
                         });
                     }
                     else
                     {
-                        _trace.WriteLine($"Invalid cookie line: {lines[i]}");
+                        _trace.WriteLine($"Invalid cookie line: {line}");
                     }
                 }
 
@@ -384,18 +390,18 @@ namespace GitCredentialManager
 
             private static DateTime ParseExpires(string expires)
             {
-                if (expires.Equals("0", StringComparison.OrdinalIgnoreCase))
+    #if NETFRAMEWORK
+                DateTime epoch = new DateTime(1970, 01, 01, 0, 0, 0, DateTimeKind.Utc);
+    #else
+                DateTime epoch = DateTime.UnixEpoch;
+    #endif
+
+                if (long.TryParse(expires, out long i))
                 {
-                    return DateTime.MinValue;
+                    return epoch.AddSeconds(i);
                 }
-                else if (DateTime.TryParse(expires, out var result))
-                {
-                    return result;
-                }
-                else
-                {
-                    return DateTime.MaxValue;
-                }
+
+                return epoch;
             }
         }
     }
--- a/src/shared/Core/HttpClientFactory.cs
+++ b/src/shared/Core/HttpClientFactory.cs
@@ -220,6 +220,7 @@ namespace GitCredentialManager
                     cookieContainer.Add(uri, new Cookie(cookie.Name, cookie.Value));
                 }
                 handler.CookieContainer = cookieContainer;
+                handler.UseCookies = true;
                 
                 _trace.WriteLine("Configured to automatically send cookie header.");
 
--- a/src/shared/Core/HttpClientFactory.cs
+++ b/src/shared/Core/HttpClientFactory.cs
@@ -7,6 +7,7 @@ using System.Net.Http;
 using System.Net.Http.Headers;
 using System.Net.Security;
 using System.Security.Cryptography.X509Certificates;
+using CurlCookie;
 
 namespace GitCredentialManager
 {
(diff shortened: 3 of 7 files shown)


ldennington (commented):
Looking good overall! I just added a few minor suggestions.

src/shared/Core/HttpClientFactory.cs
```diff
@@ -301,5 +335,68 @@ public bool TryCreateProxy(out IWebProxy proxy)
...
+                        var path = parts[2];
+                        var secureOnly = parts[3] == "TRUE";
+                        var expires = parts[4];
+                        var name = parts[5];
+                        var value = parts[6];
+
+                        cookies.Add(new Cookie(name, value, path, domain)
+                        {
+                            Expires = ParseExpires(expires),
+                            Secure = secureOnly,
```
mjcheetham, line 372: This function doens't look correct to me. The epoch for `DateTime` is `0001-01-01T00:00:00Z`, **not** the UNIX time epoch of `1970-01-01T00:00:00Z`, and `DateTime.TryParse(string, out DateTime)` would, when given an integer, take this as a number of ticks since the .NET time epoch.

I would suggest an implementation similar to below, where we parse the string as a number of seconds since the UNIX epoch:

```csharp
            private static DateTime ParseExpires(string expires)
            {
#if NETFRAMEWORK
                DateTime epoch = new DateTime(1970, 01, 01, 0, 0, 0, DateTimeKind.Utc);
#else
                DateTime epoch = DateTime.UnixEpoch;
#endif

                if (!long.TryParse(expires, out long i))
                {
                    return epoch.AddSeconds(i);
                }

                return epoch;
			}
```

Sadly since we also target .NET Framework, and `DateTime.UnixEpoch` is [only defined as of .NET Core 2.1](https://learn.microsoft.com/en-us/dotnet/api/system.datetime.unixepoch?view=net-7.0) we need to manually create the epoch for .NET Framework.

src/shared/Core.Tests/HttpClientFactoryTests.cs
```diff
@@ -311,6 +311,36 @@ public void HttpClientFactory_TryCreateProxy_ProxyEmptyUserAndEmptyPass_ReturnsT
...
+        [Theory]
+        [InlineData(null, false, null)]
+        [InlineData("~/.git-cookie", true, "# Netscape HTTP Cookie File\n" +
+                          "# https://curl.haxx.se/rfc/cookie_spec.html\n" +
+                          "# This is a generated file! Do not edit.\n" +
+                          "\n" +
+                          ".example.com\tTRUE\t/\tTRUE\t0\tcookie1\tvalue1\n" +
+                          ".example.com\tTRUE\t/\tTRUE\t0\tcookie2\tvalue2\n" +
+                          "#HttpOnly_.example.com\tTRUE\t/\tTRUE\t0\tcookie3\tvalue3\n")]
+        public void HttpClientFactory_GetClient_SetCookieOnlyIfEnabled(string cookieFilePath, bool expectCookieChecked, string cookieFileContent)
```
mjcheetham, line 323: I'd love to also see some tests of the `CookieFileParser` class, given a file containing various valid and invalid fields lines.

src/shared/Core/HttpClientFactory.cs
```diff
@@ -197,6 +197,40 @@ public HttpClient CreateClient()
...
 
+            _trace.WriteLine($"Custom cookie file has been enabled with {_settings.CustomCookieFilePath}");
+            // If IsUsingCookieEnabled is enabled, set Cookie header from cookie file, which is written by libcurl
+            if (!string.IsNullOrWhiteSpace(_settings.CustomCookieFilePath))
+            {
+                // get the filename from gitconfig
+                string cookieFilePath = _settings.CustomCookieFilePath;
+                _trace.WriteLine($"Custom cookie file has been enabled with {cookieFilePath}");
+                
+                // Throw exception if cookie file not found
```
mjcheetham, line 208: Does Git also `die`/error if the configured cookie file is not found? Or just warn? We should do the same.

src/shared/Core/HttpClientFactory.cs
```diff
@@ -197,6 +197,40 @@ public HttpClient CreateClient()
...
+                if (!_fileSystem.FileExists(cookieFilePath) || string.IsNullOrEmpty(_fileSystem.ReadAllText(cookieFilePath)))
```
mjcheetham, line 209: We are reading the cookie file contents in to a string twice! Once here for a `IsNullOrEmpty` check, and once in the `cookieParser.Parse(..)` call.

Consider changing this to something like:

```csharp
if (!_fileSystem.FileExists(cookieFilePath))
{
	// exception or trace warning.. whatever Git does in this case
	...
}

string cookieFileContents = _fileSystem.ReadAllText(cookieFilePath);

var cookieParser = new CurlCookieParser(_trace);
var cookies = cookieParser.Parse(cookieFileContents);
```

The `CurlCookieParser` should be robust to `null` or empty contents.

src/shared/Core/HttpClientFactory.cs
```diff
@@ -301,5 +335,68 @@ public bool TryCreateProxy(out IWebProxy proxy)
             proxy = null;
             return false;
         }
+
+        public class CurlCookieParser
+        {
+            private readonly ITrace _trace;
+
+            public CurlCookieParser(ITrace trace)
```
mjcheetham, line 343: We should add a quick check if `content` is null or whitespace here and just return an empty collection.

```csharp
public IList<Cookie> Parse(string content)
{
    if (string.IsNullOrWhiteSpace(content))
    {
        return Array.Empty<Cookie>();
    }

	// rest of method
}
```

mjcheetham (changes requested):
Thanks for the contribution here! This is great work to see :-)

I've left some comments that will need addressing before we can take the submission however, specifically around more tests for the `CurlCookieParser` and making it more robust to invalid or different-cased lines.

src/shared/Core/HttpClientFactory.cs
```diff
@@ -301,5 +335,68 @@ public bool TryCreateProxy(out IWebProxy proxy)
...
+                        domain = domain.StartsWith("#HttpOnly_") ? parts[0].Substring("#HttpOnly_".Length) : parts[0];
+                        domain = domain.TrimStart('.');
+                        var includeSubdomains = parts[1] == "TRUE";
+                        var path = parts[2];
+                        var secureOnly = parts[3] == "TRUE";
+                        var expires = parts[4];
+                        var name = parts[5];
+                        var value = parts[6];
+
+                        cookies.Add(new Cookie(name, value, path, domain)
```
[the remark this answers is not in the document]
miya789: Is it correct that you mean this?
```C#
...
var name = parts[5];
var value = parts[6];

cookies.Add(new Cookie()
{
    Domain = domain,
    Path = path,
    Expiration = ParseExpiration(expiration),
    HttpOnly = true,
    Secure = secureOnly,
    Name = name,
    Value = value,
});
```

src/shared/Core/HttpClientFactory.cs
```diff
@@ -301,5 +335,68 @@ public bool TryCreateProxy(out IWebProxy proxy)
...
+
+                // Parse the cookie file content
+                var lines = content.Split(new[] { '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries);
+                for (int i = 3; i < lines.Length; i++)
+                {
+                    var parts = lines[i].Split(new[] { '\t' }, StringSplitOptions.RemoveEmptyEntries);
+                    if (parts.Length >= 7)
+                    {
+                        var domain = parts[0];
+                        domain = domain.StartsWith("#HttpOnly_") ? parts[0].Substring("#HttpOnly_".Length) : parts[0];
```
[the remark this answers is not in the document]
miya789: I'll declare the `const string` as this.
```diff c#
            public IList<Cookie> Parse(string content)
            {
+               const string HttpOnlyPrefix = "#HttpOnly_";

                var cookies = new List<Cookie>();
```

src/shared/Core/HttpClientFactory.cs
```diff
@@ -197,6 +197,40 @@ public HttpClient CreateClient()
 #endif
             }
 
+            _trace.WriteLine($"Custom cookie file has been enabled with {_settings.CustomCookieFilePath}");
+            // If IsUsingCookieEnabled is enabled, set Cookie header from cookie file, which is written by libcurl
+            if (!string.IsNullOrWhiteSpace(_settings.CustomCookieFilePath))
```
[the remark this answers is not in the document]
miya789: I think that `_settings.CustomCookieFilePath`, which I added ([b841181f](https://github.com/miya789/git-credential-manager/blob/b841181f122691ead2fec506a3637cc5c1763846/src/shared/Core/Settings.cs#L634-L635
)), used `TryGetPathSetting`.

So I think the current situation is sufficient, but should I use `TryGetPathSetting` again?

src/shared/Core/HttpClientFactory.cs
```diff
@@ -197,6 +197,40 @@ public HttpClient CreateClient()
...
+                // get the filename from gitconfig
+                string cookieFilePath = _settings.CustomCookieFilePath;
+                _trace.WriteLine($"Custom cookie file has been enabled with {cookieFilePath}");
+                
+                // Throw exception if cookie file not found
+                if (!_fileSystem.FileExists(cookieFilePath) || string.IsNullOrEmpty(_fileSystem.ReadAllText(cookieFilePath)))
+                {
+                    var format = "Custom cookie file not found at path: {0}";
+                    var message = string.Format(format, cookieFilePath);
+                    throw new Trace2FileNotFoundException(_trace2, message, format, cookieFilePath);
```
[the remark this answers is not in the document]
miya789: I'll rewrite as this.
```c#
throw new FileNotFoundException(message, format, cookieFilePath);
```

src/shared/Core.Tests/HttpClientFactoryTests.cs
miya789: Excuse me, but may I ask you to tell me where and how to write a test for `CookieFileParser`, as I am new to writing C# tests and could not find a sample test to write a test for this.

src/shared/Core/HttpClientFactory.cs
```diff
@@ -301,5 +335,68 @@ public bool TryCreateProxy(out IWebProxy proxy)
...
+            {
+                _trace = trace;
+            }
+
+            public IList<Cookie> Parse(string content)
+            {
+                var cookies = new List<Cookie>();
+
+                // Parse the cookie file content
+                var lines = content.Split(new[] { '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries);
```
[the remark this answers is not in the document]
miya789: When I tried on Windows or WSL, it worked with the access to OAuth2 endpoint using cookie.
(For some reasons, I used two different PC.)

```bash
// Windows
.NET SDK (reflecting any global.json):
 Version:   7.0.105
 Commit:    e1bc5e001c

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.19044
 OS Platform: Windows
 RID:         win10-x86
 Base Path:   C:\Program Files (x86)\dotnet\sdk\7.0.105\

Host:
  Version:      7.0.5
  Architecture: x64
  Commit:       8042d61b17

.NET SDKs installed:
  7.0.105 [C:\Program Files (x86)\dotnet\sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 6.0.11 [/usr/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 7.0.0 [/usr/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 7.0.5 [/usr/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 5.0.17 [/usr/share/dotnet/shared/Microsoft.NETCore.App]
  Microsoft.NETCore.App 6.0.11 [/usr/share/dotnet/shared/Microsoft.NETCore.App]
  Microsoft.NETCore.App 7.0.0 [/usr/share/dotnet/shared/Microsoft.NETCore.App]
  Microsoft.NETCore.App 7.0.5 [/usr/share/dotnet/shared/Microsoft.NETCore.App]
  Microsoft.WindowsDesktop.App 5.0.17 [/usr/share/dotnet/shared/Microsoft.NETCore.App]
  Microsoft.WindowsDesktop.App 6.0.11 [/usr/share/dotnet/shared/Microsoft.NETCore.App]
  Microsoft.WindowsDesktop.App 7.0.0 [/usr/share/dotnet/shared/Microsoft.NETCore.App]
  Microsoft.WindowsDesktop.App 7.0.5 [/usr/share/dotnet/shared/Microsoft.NETCore.App]

Other architectures found:
  x86   [C:\Program Files\dotnet]
    registered at [HKLM\SOFTWARE\dotnet\Setup\InstalledVersions\x64\InstallLocation]

Environment variables:
  Not set

global.json file:
  Not found

Learn more:
  https://aka.ms/dotnet/info

Download .NET:
  https://aka.ms/dotnet/download


// WSL
.NET SDK (reflecting any global.json):
 Version:   6.0.405
 Commit:    27ab36058b

Runtime Environment:
 OS Name:     debian
 OS Version:  11
 OS Platform: Linux
 RID:         debian.11-x64
 Base Path:   /usr/share/dotnet/sdk/6.0.405/

global.json file:
  Not found

Host:
  Version:      6.0.13
  Architecture: x64
  Commit:       1af80ba017

.NET SDKs installed:
  6.0.405 [/usr/share/dotnet/sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 6.0.13 [/usr/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 6.0.13 [/usr/share/dotnet/shared/Microsoft.NETCore.App]

Download .NET:
  https://aka.ms/dotnet-download

Learn about .NET Runtimes and SDKs:
  https://aka.ms/dotnet/runtimes-sdk-info
```

src/shared/Core/HttpClientFactory.cs
```diff
@@ -301,5 +335,68 @@ public bool TryCreateProxy(out IWebProxy proxy)
...
+                        var includeSubdomains = parts[1] == "TRUE";
+                        var path = parts[2];
+                        var secureOnly = parts[3] == "TRUE";
+                        var expires = parts[4];
+                        var name = parts[5];
+                        var value = parts[6];
+
+                        cookies.Add(new Cookie(name, value, path, domain)
+                        {
+                            Expires = ParseExpires(expires),
```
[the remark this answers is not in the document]
miya789: According to the function of curl ([remove_expired](https://github.com/curl/curl/blob/0773d2a956f4dd637d68dbc637d762109b3efa9d/lib/cookie.c#LL384C17-L384C17)), the name `expires` is used. Therefore, I think there is no need to change the name. what do you think?

src/shared/Core/HttpClientFactory.cs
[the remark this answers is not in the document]
ldennington: Nope, this is fine. I meant to delete this comment before publishing the review but forgot. Sorry about that!

src/shared/Core/HttpClientFactory.cs
```diff
@@ -301,5 +335,68 @@ public bool TryCreateProxy(out IWebProxy proxy)
...
+
+            public IList<Cookie> Parse(string content)
+            {
+                var cookies = new List<Cookie>();
+
+                // Parse the cookie file content
+                var lines = content.Split(new[] { '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries);
+                for (int i = 3; i < lines.Length; i++)
+                {
+                    var parts = lines[i].Split(new[] { '\t' }, StringSplitOptions.RemoveEmptyEntries);
```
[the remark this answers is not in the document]
mjcheetham: > But should I rewrite this, too?
> 
> ```diff
> // Parse the cookie file content
> - var lines = content.Split(new[] { '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries);
> + var lines = content.Split(new[] { '\r', '\n' }, StringSplitOptions.None));
> ```

I think we can continue to skip over empty entries for splitting on the _lines_ in the file. It's only the fields in each line we should not remove empty parts.

src/shared/Core.Tests/HttpClientFactoryTests.cs
mjcheetham: Here's an example of some unit tests that exercise the `CurlCookieParser`. There are five tests that cover various type of input:

1. Empty cookie file - should return no cookies
2. Cookie line with missing fields - should skip bad lines
3. Cookie lines with empty fields - should use default values
4. Several cookies with various cases and values - should parse all of them OK
5. Cookie file with comments '#' - these lines should be skipped, but also careful to also check that `#HttpOnly_` lines are NOT treated as comments

```csharp
using System;
using System.Collections.Generic;
using System.Net;
using GitCredentialManager;
using GitCredentialManager.Tests.Objects;
using Xunit;

namespace Core.Tests;

public class CurlCookieParserTests
{
    [Fact]
    public void CurlCookieParser_EmptyFile_ReturnsNoCookies()
    {
        const string content = "";

        var trace = new NullTrace();
        var parser = new CurlCookieParser(trace);

        IList<Cookie> actual = parser.Parse(content);

        Assert.Empty(actual);
    }

    [Fact]
    public void CurlCookieParser_Parse_MissingFields_SkipsInvalidLines()
    {
        const string content =
            // Valid cookie
            ".example.com\tTRUE\t/path/here\tTRUE\t0\tcookie1\tvalue1\n" +

            // Missing several fields - not a valid cookie so should be skipped
            ".example.com\tTRUE\tTRUE\tcookie1\tvalue1\n";

        var trace = new NullTrace();
        var parser = new CurlCookieParser(trace);

        IList<Cookie> actual = parser.Parse(content);

        Assert.Equal(1, actual.Count);
        AssertCookie(actual[0], ".example.com", "/path/here", true, 0, "cookie1", "value1");
    }

    [Fact]
    public void CurlCookieParser_Parse_MissingFields_ReturnsValidCookiesWithDefaults()
    {
        const string content =
            // Empty path field (default "/")
            ".example.com\tTRUE\t\tTRUE\t852852\tcookie1\tvalue1\n"+

            // Empty expiration field (default 0)
            ".example.com\tTRUE\t/path/here\tTRUE\t\tcookie1\tvalue1";

        var trace = new NullTrace();
        var parser = new CurlCookieParser(trace);

        IList<Cookie> actual = parser.Parse(content);

        Assert.Equal(2, actual.Count);
        AssertCookie(actual[0], ".example.com", "/", true, 852852, "cookie1", "value1");
        AssertCookie(actual[1], ".example.com", "/path/here", true, 0, "cookie1", "value1");
    }

    [Fact]
    public void CurlCookieParser_Parse_ValidFields_ReturnsValidCookies()
    {
        const string content =
            ".example.com\tTRUE\t/path\tTRUE\t0\tcookie1\tvalue1\n" +
            ".example.com\tfAlSe\t/path\ttRuE\t0\tcookie1\tvalue1\n" +
            ".example.com\tTRUE\t/path\tTRUE\t0\tcookie1 with spaces\tvalue1 with spaces\n" +
            ".example.com\tFALSE\t/path\tTRUE\t0\tcookie1\tvalue1\n" +
            "example.com\tFALSE\t/path\tTRUE\t0\tcookie1\tvalue1\n" +
            "example.com\tTRUE\t/path\tTRUE\t0\tcookie1\tvalue1\n" +
            ".example.com\tTRUE\t/path\tFALSE\t0\tcookie1\tvalue1\n" +
            ".example.com\tTRUE\t/path\tFALSE\t401654\tcookie1\tvalue1\n";

        var trace = new NullTrace();
        var parser = new CurlCookieParser(trace);

        IList<Cookie> actual = parser.Parse(content);

        Assert.Equal(8, actual.Count);
        AssertCookie(actual[0], ".example.com", "/path", true, 0, "cookie1", "value1");
        AssertCookie(actual[1], "example.com", "/path", true, 0, "cookie1", "value1");
        AssertCookie(actual[2], ".example.com", "/path", true, 0, "cookie1 with spaces", "value1 with spaces");
        AssertCookie(actual[3], "example.com", "/path", true, 0, "cookie1", "value1");
        AssertCookie(actual[4], "example.com", "/path", true, 0, "cookie1", "value1");
        AssertCookie(actual[5], "example.com", "/path", true, 0, "cookie1", "value1");
        AssertCookie(actual[6], ".example.com", "/path", false, 0, "cookie1", "value1");
        AssertCookie(actual[7], ".example.com", "/path", false, 401654, "cookie1", "value1");
    }

    [Fact]
    public void CurlCookieParser_Parse_Comments_ReturnsCookies()
    {
        const string content =
            // Comment block
            "# This is a cookie file with various comments!\n" +
            "# Lines starting with # are comments, except those that\n" +
            "# start with exactly '#HttpOnly_'.. two #s is a comment still!\n" +

            // This is still a comment!
            "##HttpOnly_ <-- this is a comment still!\n" +

            // Valid line
            ".example.com\tTRUE\t/\tTRUE\t0\tcookie1\tvalue1\n" +

            // Commented out cookie line
            "#.example.com\tTRUE\t/\tTRUE\t0\tcookie1\tvalue1\n" +

            // Valid cookie but HTTP only
            "#HttpOnly_.example.com\tTRUE\t/\tTRUE\t0\tcookie1\tvalue1\n";

        var trace = new NullTrace();
        var parser = new CurlCookieParser(trace);

        IList<Cookie> actual = parser.Parse(content);

        Assert.Equal(2, actual.Count);
        AssertCookie(actual[0], ".example.com", "/", true, 0, "cookie1", "value1");
        AssertCookie(actual[1], ".example.com", "/", true, 0, "cookie1", "value1");
    }

    private static void AssertCookie(Cookie cookie, string domain, string path, bool secureOnly, long expires, string name, string value)
    {
        Assert.Equal(name, cookie.Name);
        Assert.Equal(value, cookie.Value);
        Assert.Equal(domain, cookie.Domain);
        Assert.Equal(path, cookie.Path);
        Assert.Equal(secureOnly, cookie.Secure);
        Assert.Equal(expires, cookie.Expires.Subtract(DateTime.UnixEpoch).TotalSeconds);
    }
}

```

Some of these tests fail with the current implementation which points to some bugs! I think most of them are already pointed out in previous review comments.

I hope this helps!

src/shared/Core/HttpClientFactory.cs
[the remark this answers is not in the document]
ldennington: Yep! That looks good.

src/shared/Core/HttpClientFactory.cs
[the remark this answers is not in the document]
ldennington: Fair enough, it's fine to follow `curl`'s lead and leave as-is.

src/shared/Core/HttpClientFactory.cs
```diff
@@ -301,5 +335,68 @@ public bool TryCreateProxy(out IWebProxy proxy)
...
+                        _trace.WriteLine($"Invalid cookie line: {lines[i]}");
+                    }
+                }
+
+                return cookies;
+            }
+
+            private static DateTime ParseExpires(string expires)
+            {
+                if (expires.Equals("0", StringComparison.OrdinalIgnoreCase))
```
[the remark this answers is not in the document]
miya789: I'll set the default value with the following code:
```c#
DateTime epoch = new DateTime(1970, 01, 01, 0, 0, 0, DateTimeKind.Utc);
```

src/shared/Core/HttpClientFactory.cs
```diff
@@ -301,5 +335,68 @@ public bool TryCreateProxy(out IWebProxy proxy)
...
+                _trace = trace;
+            }
+
+            public IList<Cookie> Parse(string content)
+            {
+                var cookies = new List<Cookie>();
+
+                // Parse the cookie file content
+                var lines = content.Split(new[] { '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries);
+                for (int i = 3; i < lines.Length; i++)
```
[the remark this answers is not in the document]
miya789: I see.
I'll rewrite the loop and the condition like this. Is there any problem?
```diff c#
      // Parse the cookie file content
      var lines = content.Split(new[] { '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries);
-     for (int i = 3; i < lines.Length; i++)
+     foreach (var line in lines)
      {
        var parts = line.Split(new[] { '\t' }, StringSplitOptions.None);
-       if (parts.Length >= 7)
+       if (parts.Length >= 7 && (!parts[0].StartsWith("#") || parts[0].StartsWith(HttpOnlyPrefix)))
```

src/shared/Core.Tests/HttpClientFactoryTests.cs
miya789: Thank you very much for kindly including the example tests.
Thanks to you, I can now modify the parse function more easily.

src/shared/Core/HttpClientFactory.cs
miya789: As far as I could confirm by operating it, Git did not die.
Therefore, if the configured cookie file is not found, I will change it so that only operations related to cookies are not performed.

src/shared/Core/HttpClientFactory.cs
```diff
@@ -301,5 +335,68 @@ public bool TryCreateProxy(out IWebProxy proxy)
...
+                {
+                    var parts = lines[i].Split(new[] { '\t' }, StringSplitOptions.RemoveEmptyEntries);
+                    if (parts.Length >= 7)
+                    {
+                        var domain = parts[0];
+                        domain = domain.StartsWith("#HttpOnly_") ? parts[0].Substring("#HttpOnly_".Length) : parts[0];
+                        domain = domain.TrimStart('.');
+                        var includeSubdomains = parts[1] == "TRUE";
```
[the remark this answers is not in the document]
miya789: Shoud I rewrite like this?
```c#
          var domain = parts[0].StartsWith(HttpOnlyPrefix) ? parts[0].Substring(HttpOnlyPrefix.Length) : parts[0];
          var includeSubdomains = parts[1] == "TRUE";
          if (!includeSubdomains)
          {
            domain = domain.TrimStart('.');
          }
```

miya789:
I modified the code based on your review.
However, I did not understand [`curl/cookie.c`](https://github.com/curl/curl/blob/0773d2a956f4dd637d68dbc637d762109b3efa9d/lib/cookie.c) at all, so I have only fixed the parts mentioned in your review.

src/shared/Core/CurlCookie.cs
```diff
@@ -0,0 +1,83 @@
+using System;
+using System.Collections.Generic;
+using System.Net;
+using GitCredentialManager;
+
+namespace CurlCookie
```
mjcheetham, line 6: The namespace should be `GitCredentialManager`:

```suggestion
namespace GitCredentialManager
```

src/shared/Core/CurlCookie.cs
```diff
@@ -0,0 +1,83 @@
...
+
+            // Parse the cookie file content
+            var lines = content.Split(new[] { '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries);
+            foreach (var line in lines)
+            {
+                var parts = line.Split(new[] { '\t' }, StringSplitOptions.None);
+                if (parts.Length >= 7 && (!parts[0].StartsWith("#") || parts[0].StartsWith(HttpOnlyPrefix)))
+                {
+                    var domain = parts[0].StartsWith(HttpOnlyPrefix) ? parts[0].Substring(HttpOnlyPrefix.Length) : parts[0];
+                    var includeSubdomains = parts[1] == "TRUE";
```
mjcheetham, line 36: Let's do case insensitive comparison for all boolean fields.

```suggestion
                    var includeSubdomains = StringComparer.OrdinalIgnoreCase.Equals(parts[1], "TRUE");
```

src/shared/Core/CurlCookie.cs
```diff
@@ -0,0 +1,83 @@
...
+                    if (!includeSubdomains)
+                    {
+                        domain = domain.TrimStart('.');
+                    }
+                    var path = parts[2] == "" ? "/" : parts[2];
```
mjcheetham, line 41: Let's cover all whitespace or empty/null possibilities:

```suggestion
                    var path = string.IsNullOrWhiteSpace(parts[2]) ? "/" : parts[2];
```

src/shared/Core/HttpClientFactory.cs
```diff
@@ -7,6 +7,7 @@
 using System.Net.Http.Headers;
 using System.Net.Security;
 using System.Security.Cryptography.X509Certificates;
+using CurlCookie;
```
mjcheetham, line 10: Once we name the namespace in the `CurlCookie.cs` file to `GitCredentialManager`, we can drop this `using` statement.

```suggestion
```

src/shared/Core/HttpClientFactory.cs
```diff
@@ -197,6 +198,33 @@ public HttpClient CreateClient()
...
+
+                // Set the cookie
+                var cookieContainer = new CookieContainer();
+                foreach (var cookie in cookies)
+                {
+                    var schema = cookie.Secure ? "https" : "http";
+                    var uri = new UriBuilder(schema, cookie.Domain.TrimStart('.')).Uri;
+                    cookieContainer.Add(uri, new Cookie(cookie.Name, cookie.Value));
+                }
+                handler.CookieContainer = cookieContainer;
```
mjcheetham, line 222: We may also need to set `UseCookies` to `true`?

```suggestion
                handler.CookieContainer = cookieContainer;
                handler.UseCookies = true;
```

mjcheetham (commented):
Great stuff @miya789! Just a few minor suggestions, and I think we will be good to merge! 🎉

miya789:
Thank you for comments, @mjcheetham -san!
Should I use rebase to clean up the commits?
Please let me know if there is a format specification.

mjcheetham (approved):
This is great work @miya789! Thank you again for the contribution.
We can merge this now.

PR #1263, opened 2023-05-18
release: remove .NET tool deployment from release workflow

ldennington:
Move deployment of .NET tool to nuget.org out of release workflow. With these changes, the .NET tool is published only after a release has been published (rather than with the creation of the release). This emulates what we currently do with Homebrew and ensures we only release the .NET tool when we are confident enough in a release's stability to publish it.

A test of this new deployment workflow targeting [NuGet's QA Gallery](https://int.nugettest.org/packages/git-credential-manager/) can be found [here](https://github.com/ldennington/git-credential-manager/actions/runs/5018996340/jobs/8999029309).

### DIFF
--- a/.github/workflows/release-dotnet-tool.yaml
+++ b/.github/workflows/release-dotnet-tool.yaml
@@ -1,0 +1,22 @@
+name: release-dotnet-tool
+on:
+  release:
+    types: [released]
+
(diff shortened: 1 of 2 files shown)


mjcheetham (approved):
🎉

PR #1265, opened 2023-05-19
Update `System.CommandLine` to latest version

mjcheetham:
Update `System.CommandLine` to the latest version. There have been some changes to the API since the last version around binding a command to a handler, and the removal of the model binding.

We now explicitly specify the options when setting the handlers. This means we should be able to avoid the use of reflection inside of the package, making it trimable in the future!


ldennington (approved):
LGTM!

PR #1267, opened 2023-05-22
github: add multi-account prompt and login/logout commands

mjcheetham:
Add multi-account support for the GitHub host provider.

Previously it was possible to use multiple accounts with GitHub, but this required prior knowledge about including the username in the remote URL (e.g. `https://mona_alt@github.com/mona/test`), and without this the incorrect account may be selected, and subsequently erased due to insufficient permissions.

This PR adds a few different features and prompts to improve discoverability and experience using multiple GitHub accounts with GCM:

1. Add `login`, `logout` and `list` commands for the GitHub provider
2. Add UI and TTY prompts to select between GitHub user accounts
3. Abridged documentation and links to quickly discover the 'user-in-remote-URL' functionality
4. (bonus!) Ability to suppress GUI prompts over text-based ones via the command-line using `--no-ui`

One thing that is absent compared to the Azure Repos provider's multi-account support is the lack of account 'binding'. This is due to lack of knowledge, by default, of the full remote URL from Git to be able to discern between different repositories or organisations on the same host.

**Command-line usage:**

```console
% git-credential-manager github --help    
Description:
  Commands for interacting with the GitHub host provider

Usage:
  git-credential-manager github [command] [options]

Options:
  --no-ui         Do not use graphical user interface prompts
  -?, -h, --help  Show help and usage information

Commands:
  list              List all known GitHub accounts.
  login             Add a GitHub account.
  logout <account>  Remove a GitHub account.
```
```console
% git-credential-manager github login --help
Description:
  Add a GitHub account.

Usage:
  git-credential-manager github login [options]

Options:
  --url <url>             URL of the GitHub instance to target, otherwise use GitHub.com
  --username <username>   User name to authenticate with
  --device                Use device flow to authenticate
  --browser, --web        Use a web browser to authenticate
  --pat, --token <token>  Use personal access token to authenticate
  --no-ui                 Do not use graphical user interface prompts
  -?, -h, --help          Show help and usage information
```
```console
% git-credential-manager github logout --help 
Description:
  Remove a GitHub account.

Usage:
  git-credential-manager github logout <account> [options]

Arguments:
  <account>  Account to remove

Options:
  --url <url>     URL of the GitHub instance to target, otherwise use GitHub.com
  --no-ui         Do not use graphical user interface prompts
  -?, -h, --help  Show help and usage information
```
```console
% git-credential-manager github list --help  
Description:
  List all known GitHub accounts.

Usage:
  git-credential-manager github list [options]

Options:
  --url <url>     URL of the GitHub instance to target, otherwise use GitHub.com
  --no-ui         Do not use graphical user interface prompts
  -?, -h, --help  Show help and usage information
```

**Select account UI prompts:**

Avalonia|Windows legacy helper (WPF)
-|-
image|image

### DIFF
--- a/docs/multiple-users.md
+++ b/docs/multiple-users.md
@@ -79,3 +79,11 @@ git remote set-url origin https://employee9999@example.com/big-company/secret-re
 which you should also be aware of if you're using it.
 
 [azure-access-tokens]: azrepos-users-and-tokens.md
+
+## GitHub
+
+You can use the `github [list | login | logout]` commands to manage your GitHub
+accounts. These commands are documented in the [command-line usage][cli-usage]
+or by running `git-credential-manager github --help`.
+
+[cli-usage]: usage.md
--- a/docs/multiple-users.md
+++ b/docs/multiple-users.md
@@ -4,6 +4,15 @@ If you work with multiple different identities on a single Git hosting service,
 you may be wondering if Git Credential Manager (GCM) supports this workflow. The
 answer is yes, with a bit of complexity due to how it interoperates with Git.
 
+---
+
+**Prompted to select an account?**
+
+Read the [**TL;DR** section][tldr] below for a quick summary of how to make GCM
+remember which account to use for which repository.
+
+---
+
 ## Foundations: Git and Git hosts
 
 Git itself doesn't have a single, strong concept of "user". There's the
@@ -86,4 +95,28 @@ You can use the `github [list | login | logout]` commands to manage your GitHub
 accounts. These commands are documented in the [command-line usage][cli-usage]
 or by running `git-credential-manager github --help`.
 
+## TL;DR: Tell GCM to remember which account to use
+
+The easiest way to have GCM remember which account to use for which repository
+is to include the account name in the remote URL. If you're using HTTPS remotes,
+you can include the account name in the URL by inserting it before the `@` sign
+in the domain name.
+
+For example, if you want to always use the `alice` account for the `mona/test`
+GitHub repository, you can clone it using the `alice` account by running:
+
+```shell
+git clone https://alice@github.com/mona/test
+```
+
+To update an existing clone, you can run `git remote set-url` to update the URL:
+
+```shell
+git remote set-url origin https://alice@github.com/mona/test
+```
+
+If your account name includes an `@` then remember to escape this character
+using `%40`: `https://alice%40contoso.com@example.com/test`.
+
+[tldr]: #tldr-tell-gcm-to-remember-which-account-to-use
 [cli-usage]: usage.md
--- a/src/shared/Core/Authentication/AuthenticationBase.cs
+++ b/src/shared/Core/Authentication/AuthenticationBase.cs
@@ -21,13 +21,13 @@ namespace GitCredentialManager.Authentication
         }
 
         protected Task<IDictionary<string, string>> InvokeHelperAsync(string path, string args,
-            IDictionary<string, string> standardInput = null)
+            StreamReader standardInput = null)
         {
-            return InvokeHelperAsync(path, args, null, CancellationToken.None);
+            return InvokeHelperAsync(path, args, standardInput, CancellationToken.None);
         }
 
         protected internal virtual async Task<IDictionary<string, string>> InvokeHelperAsync(string path, string args,
-            IDictionary<string, string> standardInput, CancellationToken ct)
+            StreamReader standardInput, CancellationToken ct)
         {
             var procStartInfo = new ProcessStartInfo(path)
             {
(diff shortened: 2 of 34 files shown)


src/shared/Core/Authentication/AuthenticationBase.cs
```diff
@@ -56,9 +56,15 @@ protected AuthenticationBase(ICommandContext context)
             // Kill the process upon a cancellation request
             ct.Register(() => process.Kill());
 
-            if (!(standardInput is null))
+            // Write the standard input to the process if we have any to write
+            if (standardInput is not null)
```
ldennington, line 60: Nice cleanup.

src/shared/GitHub/UI/Views/SelectAccountView.axaml.cs
```diff
@@ -0,0 +1,18 @@
...
+    public SelectAccountView()
+    {
+        InitializeComponent();
+    }
+
+    private void InitializeComponent()
+    {
+        AvaloniaXamlLoader.Load(this);
+    }
+}
```
ldennington, line 18: Nit: add newline.

docs/multiple-users.md
```diff
@@ -86,4 +95,28 @@ You can use the `github [list | login | logout]` commands to manage your GitHub
 accounts. These commands are documented in the [command-line usage][cli-usage]
 or by running `git-credential-manager github --help`.
 
+## TL;DR: Tell GCM to remember which account to use
```
ldennington, line 98: Great explanation!

ldennington (approved):
Awesome work! 🚀 I did notice something in my testing that may or may not be by design. Running the below forced a prompt, even though I had GitHub creds for ldennington in my KeyChain.

```
Users/ldennington/repos/git-credential-manager/out/shared/Git-Credential-Manager/bin/Debug/net6.0/git-credential-manager github login --username ldennington
```
I also had to consent:

Screenshot 2023-05-23 at 1 54 01 PM

I then tried Device code and the same thing happened:
```
Users/ldennington/repos/git-credential-manager/out/shared/Git-Credential-Manager/bin/Debug/net6.0/git-credential-manager github login --username ldennington
--device
```
Screenshot 2023-05-23 at 1 55 36 PM

If this is by design, carry on! But wanted to call it out just in case.

mjcheetham:
@ldennington:
> I did notice something in my testing that may or may not be by design. Running the below forced a prompt, even though I had GitHub creds for ldennington in my KeyChain.

This is expected, I guess. We don't check for existing credentials during `github login`. The idea was/is to _add_ a new login, or replace an existing credential for the same account.

The GitHub CLI takes the opposite approach, and say "you're already logged in, login again?":

```console
% gh auth login
? What account do you want to log into? GitHub.com
? You're already logged into github.com. Do you want to re-authenticate? (y/N) 
```

I guess we could modify our `login` command here to do the following, if you think this makes more sense?

- `login`: show the login prompt - we don't know if the user is going to pick a different account from what may/may-not already exist.
- `login --username <username>`: check for an existing credential for `<username>` and if it already exists, then bail saying "credential for user already exists" or similar
- `login --username <username> --force` or `--add`: explicitly add/replace an already existing credential.. the user may wish to renew the token used for some reason

Thoughts?

src/shared/Core/Authentication/MicrosoftAuthentication.cs
```diff
@@ -198,6 +198,8 @@ private async Task<bool> UseDefaultAccountAsync(string userName)
         {
             ThrowIfUserInteractionDisabled();
 
+            userName = "user@contoso.com";
```
mjcheetham, line 201: Oops.. missed this! Will remove 😅

ldennington:
> Thoughts?

I like this approach - it aligns better with what I expected to happen when I ran the command.

src/shared/GitHub/GitHubHostProvider.Commands.cs
```diff
@@ -63,15 +69,37 @@ private void ListAccounts(Uri url)
...
+        userName ??= url.GetUserName();
+
         string service = GetServiceName(url);
 
-        ICredential credential = await GenerateCredentialAsync(url, userName);
+        ICredential credential;
+        if (token is not null)
+        {
+            // Resolve the GitHub user handle if the user didn't supply one
+            if (string.IsNullOrEmpty(userName))
```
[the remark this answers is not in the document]
mjcheetham: Yes. We're doing `userName = userName ?? url.GetUserName()` and so `userName` could still be null or empty if there was no `--username` option passed _and_ no user in the remote URL.

`git-credential-manager github login --token $TOKEN` is a complete valid command, and `--username` is not required.

src/shared/GitHub/GitHubHostProvider.Commands.cs
```diff
@@ -24,11 +24,17 @@ public ProviderCommand CreateCommand()
         listCmd.SetHandler(ListAccounts, urlOpt);
 
         //
-        // login [--url <url>]
+        // login [--url <url>] [--username <username>] [--device | --browser/--web | --pat/--token <token>]
         //
         var loginCmd = new Command("login", "Add a GitHub account.");
+        var userNameOpt = new Option<string>("--username", "User name to authenticate with");
```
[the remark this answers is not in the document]
mjcheetham: Elsewhere in GCM public facing strings, we've used "user name" as two words rather than "username".

mjcheetham:
> I like this approach - it aligns better with what I expected to happen when I ran the command.

Sure! I can work on this.

mjcheetham:
Added! cc @ldennington 

```console
% gcm-dev github login --username mjcheetham
Account 'mjcheetham' already has credentials for https://github.com; use --force to re-authenticate
```

ldennington:
> Added! cc @ldennington
> 
Awesome! Took a look at the corresponding changes and they all look good to me.

mjcheetham:
@ldennington just pushed one more commit (https://github.com/git-ecosystem/git-credential-manager/pull/1267/commits/b94ade0780815170ff48f9dd60ad5fc44db63c18) on top to allow double-clicking the account from the list, to accept and continue... so you don't have to select the account from the list and then press the 'Continue' button.

PR #1274, opened 2023-05-26
dependencies: migrate from json.net to `system.text.json`

ldennington:
# Overview

Migrate GCM from JSON.NET to `System.Text.Json`.

The first commit updates to .NET 7.0 to take advantage of certain features like `JsonRequired`. The second updates `Microsoft.Identity.Client` so that the workaround for [this issue](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/4108 ) is no longer necessary. The third updates the current global Json.NET dependency to `System.Text.Json` and ensures it is only used for .NET Framework. The fourth adds a custom converter to handle Bitbucket's use of the  non-standard 'scopes' property to provide token endpoint results. The fifth through ninth commits update `Trace2Message`, the Bitbucket REST API, the GitHub REST API, the Azure DevOps API tests, and OAuth-specific code to use `System.Text.Json` instead of Json.NET for JSON serialization/deserialization.

# Performance comparison
Comparison runs were done using [`dotnet-trace`](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/dotnet-trace) with the command `printf "protocol=https\nhost=github.com" | git-credential-manager get`. The machine used was a Mac with a 2.4 GHz 8-Core Intel Core i9 processor running Ventura 13.4. Times are in milliseconds.

## Individual Run Times
| **Newtonsoft.json** | **System.Text.Json** |
|---------------------|----------------------|
| 364.08              | 304.92               |
| 366.11              | 305.23               |
| 377.31              | 313.09               |
| 381.3               | 313.51               |
| 373.25              | 297.71               |
| 373.98              | 312.78               |
| 370.84              | 317.86               |
| 368.25              | 311.78               |
| 369.79              | 307.69               |

## Run Summary
|             | **Newtonsoft.json** | **System.Text.Json** |
|-------------|---------------------|----------------------|
| **Average** | 371.66          | 309.40          |
| **Median**  | 370.84              | 311.78               |

## Analysis
Per the above, the transition to `System.Text.Json` and .NET 7.0 decreased end-to-end execution times by about 17% on average.


Directory.Build.props
```diff
@@ -28,8 +28,8 @@
 
   <ItemGroup>
     <!-- Workaround https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/4108  -->
-    <PackageReference Include="Newtonsoft.Json">
-      <Version>13.0.1</Version>
+    <PackageReference Include="System.Text.Json">
+      <Version>7.0.2</Version>
     </PackageReference>
```
mjcheetham, line 33: I think the MSAL bug was fixed in 4.54.0!

src/shared/Atlassian.Bitbucket/BitbucketTokenEndpointResponseJson.cs
```diff
@@ -1,12 +1,57 @@
...
+                }
+
+                if (reader.TokenType == JsonTokenType.PropertyName)
+                {
+                    var propertyName = reader.GetString();
+                    reader.Read();
+                    if (propertyName != null)
+                    {
+                        // Bitbucket uses "scopes" for the scopes property name rather than the standard "scope" name
+                        if (propertyName.Equals("scopes", StringComparison.OrdinalIgnoreCase))
```
mjcheetham, line 42: How are the other properties on the JSON response object populated by this converter?

src/shared/Core.Tests/TokenEndpointResponseJsonTest.cs
```diff
@@ -0,0 +1,25 @@
...
+        var scopesString = "x,y,z";
+        var json = $"{{\"access_token\": \"\", \"token_type\": \"\", \"expires_in\": \"1000\", \"scopes\": \"a,b,c\", \"scope\": \"{scopesString}\"}}";
+
+        var result = JsonSerializer.Deserialize<TokenEndpointResponseJson>(json,
+            new JsonSerializerOptions
+            {
+                PropertyNameCaseInsensitive = true,
+            });
+
+        Assert.Equal(scopesString, result.Scope);
```
mjcheetham, line 22: Can we verify the other properties are correctly deserialised too?

src/shared/Core/Trace2Message.cs
```diff
@@ -175,19 +179,21 @@ private static string BuildSpan(PerformanceFormatSpan component, string data)
...
     {
-        return JsonConvert.SerializeObject(this,
-                new StringEnumConverter(typeof(SnakeCaseNamingStrategy)),
-            new IsoDateTimeConverter()
+        return JsonSerializer.Serialize(this,
+            new JsonSerializerOptions
             {
-                DateTimeFormat = TimeFormat
+                PropertyNameCaseInsensitive = true,
+                Converters = { new JsonStringEnumConverter(new SnakeCaseNamingPolicy()) }
```
mjcheetham, line 196: Could we make the `JsonSerializerOptions` a `protected static readonly` instance on the abstract `Trace2Message`, to avoid needing to construct a new one for each message type?

mjcheetham (commented):
Looks good! Happy to see even using reflection we're seeing a 10% improvement 🎉 

I notice there's a warning from the build in _OAuthClient.cs_:
> The using directive for 'System.Text.Json' appeared previously in this namespace

src/shared/Core.Tests/TokenEndpointResponseJsonTest.cs
ldennington: Good catch. Made the necessary corrections so that all properties are correctly deserialized.

src/shared/Core/Trace2Message.cs
```diff
@@ -13,28 +12,33 @@ public abstract class Trace2Message
     private const string NormalPerfTimeFormat = "HH:mm:ss.ffffff";
 
     protected const string EmptyPerformanceSpan =  "|     |           |           |             ";
-    protected const string TimeFormat = "yyyy'-'MM'-'dd'T'HH':'mm':'ss'.'ffffff'Z'";
 
-    [JsonProperty("event")]
+    [JsonPropertyName("event")]
+    [JsonPropertyOrder(1)]
```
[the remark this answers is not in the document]
ldennington: It's not - I added it to ensure ordering for testing purposes and put a note about it in the commit message.

ldennington:
@mjcheetham - I decided that it would be a better idea to run performance comparisons against `HEAD` rather than the latest release. I went ahead and also re-ran the `System.Text.Json` numbers as well and I've updated the results in the PR description. I'm thinking that the jump from the 10% decrease to the 17% decrease was brought on by the move to .NET 7.0 and have noted that as well.

mjcheetham (approved):

PR #1279, opened 2023-06-01
Upload artifacts from CI builds

mjcheetham:
We currently don't capture or upload artifacts from our CI builds.. we should do this! This will help with testing PRs and the tip of `main`.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -1,20 +1,19 @@
-name: GCM
+name: ci
 
 on:
   workflow_dispatch:
   push:
-    branches: [ main, linux ]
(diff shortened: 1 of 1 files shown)


derrickstolee (approved):
I was initially surprised that you abandoned the matrix, but then realized that the artifacts are special on a per-platform basis, so the matrix wasn't really helping us.

mjcheetham:
You can see an example with artifacts from this PR run:
https://github.com/git-ecosystem/git-credential-manager/actions/runs/5148298999?pr=1279#artifacts

mjcheetham:
Merging with bypass as the required build checks have changed as a result of this PR!
Will update the PR merge requirements post this merge.

Issue #594, opened 2021-12-31
Change redirect URI from localhost to 127.0.0.1


hickford:
The OAuth spec https://datatracker.ietf.org/doc/html/rfc8252#section-8.3 recommends to use 127.0.0.1 instead of localhost:

>  While redirect URIs using localhost (i.e., "http://localhost:{port}/{path}") function similarly to loopback IP redirects described in Section 7.3, the use of localhost is NOT RECOMMENDED.  Specifying a redirect URI with the loopback IP literal rather than localhost avoids inadvertently listening on network interfaces other than the loopback interface.  It is also less susceptible to client-side firewalls and misconfigured host name resolution on the user's device.

Examples in code -- BitBucket, GitHub, Azure https://github.com/GitCredentialManager/git-credential-manager/search?q=localhost

hickford:
NB. This would require changing the redirect URI in the OAuth app settings on GitHub etc. https://docs.github.com/en/developers/apps/building-oauth-apps/authorizing-oauth-apps#redirect-urls

Since GitHub allows at most one URI, I can't see how to do this in a backwards compatible way.

mislav:
For GitHub, instead of redirecting back to either `127.0.0.1` or `localhost`, GCM should switch to implementing the [Device Authorization Flow](https://docs.github.com/en/developers/apps/building-oauth-apps/authorizing-oauth-apps#device-flow).

The benefits are that the completed browser flow does not have to redirect to any particular URL (and thus GCM doesn't have to spawn a web server at localhost), and the user is able to complete the browser flow on a different machine than the one where GCM is running as a process.

hickford:
GCM already supports OAuth Device Authorization Grant for GitHub . You can configure a preference with [credential.gitHubAuthModes](https://github.com/GitCredentialManager/git-credential-manager/blob/main/docs/configuration.md#credentialgithubauthmodes).

mislav:
@hickford Thanks for the info! I was able to opt into Device flow by doing this:
```
git config --global credential.gitHubAuthModes device
```
My next question is, why hasn't GCM auto-detected this when authorizing a git operation from `https://github.com`, but instead it defaulted to OAuth web app flow? I would argue that the Device flow should be the default and that it's straightforward to feature-detect.

hickford:
[...]

PR #1281, opened 2023-06-06
Preserve exact original redirect URL in OAuth client

mjcheetham:
The OAuth 2.0 spec requires that redirect URLs be matched _exactly_ if specified, including matching trailing slashes.

Since the .NET `Uri` type's `.ToString()` method will append a trailing slash to the end of path-less URLs (e.g., "http://foo" => "http://foo/") we need to use the `.OriginalString` property instead.

Shoring up this area in anticipation for changes to support multiple GitHub redirect URLs with #594

### DIFF
--- a/src/shared/Core.Tests/Authentication/OAuth2ClientTests.cs
+++ b/src/shared/Core.Tests/Authentication/OAuth2ClientTests.cs
@@ -44,6 +44,55 @@ namespace GitCredentialManager.Tests.Authentication
             Assert.Equal(expectedAuthCode, result.Code);
         }
 
+        [Theory]
+        [InlineData("http://localhost")]
+        [InlineData("http://localhost/")]
+        [InlineData("http://localhost/oauth-callback")]
+        [InlineData("http://localhost/oauth-callback/")]
+        [InlineData("http://127.0.0.1")]
+        [InlineData("http://127.0.0.1/")]
+        [InlineData("http://127.0.0.1/oauth-callback")]
+        [InlineData("http://127.0.0.1/oauth-callback/")]
+        public async Task OAuth2Client_GetAuthorizationCodeAsync_RedirectUrlOriginalStringPreserved(string expectedRedirectUrl)
+        {
+            var baseUri = new Uri("https://example.com");
+            OAuth2ServerEndpoints endpoints = CreateEndpoints(baseUri);
+
+            var httpHandler = new TestHttpMessageHandler {ThrowOnUnexpectedRequest = true};
+
+            OAuth2Application app = new OAuth2Application(TestClientId)
+            {
+                Secret = TestClientSecret,
+                RedirectUris = new[] {new Uri(expectedRedirectUrl)}
+            };
+
+            var server = new TestOAuth2Server(endpoints);
+            server.RegisterApplication(app);
+            server.Bind(httpHandler);
+            server.TokenGenerator.AuthCodes.Add("unused");
+            server.AuthorizationEndpointInvoked += (_, request) =>
+            {
+                IDictionary<string, string> actualParams = request.RequestUri.GetQueryParameters();
+                Assert.True(actualParams.TryGetValue(OAuth2Constants.RedirectUriParameter, out string actualRedirectUri));
+                Assert.Equal(expectedRedirectUrl, actualRedirectUri);
+            };
+
+            IOAuth2WebBrowser browser = new TestOAuth2WebBrowser(httpHandler);
+
+            var redirectUri = new Uri(expectedRedirectUrl);
+
+            var trace2 = new NullTrace2();
+            OAuth2Client client =  new OAuth2Client(
+                new HttpClient(httpHandler),
+                endpoints,
+                TestClientId,
+                trace2,
+                redirectUri,
+                TestClientSecret);
+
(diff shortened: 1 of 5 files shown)


src/shared/Core.Tests/Authentication/OAuth2ClientTests.cs
```diff
@@ -44,6 +44,55 @@ public async Task OAuth2Client_GetAuthorizationCodeAsync()
...
+            var trace2 = new NullTrace2();
+            OAuth2Client client =  new OAuth2Client(
+                new HttpClient(httpHandler),
+                endpoints,
+                TestClientId,
+                trace2,
+                redirectUri,
+                TestClientSecret);
+
+            await client.GetAuthorizationCodeAsync(new[] { "unused" }, browser, null, CancellationToken.None);
```
ldennington, line 93: We don't seem to be testing any outcomes here. Are we just validating that this works with these redirect URLs?

src/shared/TestInfrastructure/Objects/TestOAuth2Server.cs
```diff
@@ -527,23 +527,25 @@ public TokenEndpointResponseJson CreateTokenByDeviceCodeGrant(TestOAuth2ServerTo
...
-                if (redirectUri == uri)
+                // We only accept exact matches, including trailing slashes and case sensitivity
+                if (StringComparer.Ordinal.Equals(redirectUri.OriginalString, url))
                 {
                     return true;
                 }
 
-                // For localhost we ignore the port number
-                if (redirectUri.IsLoopback && uri.IsLoopback)
+                // For loopback URLs _only_ we ignore the port number
```
ldennington, line 540: I'm curious - why do we ignore ports for loopback urls?

src/shared/TestInfrastructure/Objects/TestOAuth2Server.cs
```diff
@@ -527,23 +527,25 @@ public TokenEndpointResponseJson CreateTokenByDeviceCodeGrant(TestOAuth2ServerTo
...
                 {
-                    var cmp = StringComparer.OrdinalIgnoreCase;
+                    // *Case-sensitive* comparison of scheme, host and path
+                    var cmp = StringComparer.Ordinal;
 
-                    // Uri::Authority does not include port, whereas Uri::Host does
+                    // Uri::Authority includes port, whereas Uri::Host does not
                     return cmp.Equals(redirectUri.Scheme, uri.Scheme) &&
-                           cmp.Equals(redirectUri.Authority, uri.Authority) &&
+                           cmp.Equals(redirectUri.Host, uri.Host) &&
```
ldennington, line 548: Was this incorrect before?

ldennington (approved):
LGTM! Just had a couple minor questions.

src/shared/TestInfrastructure/Objects/TestOAuth2Server.cs
mjcheetham: Yes. I wanted to compare just the 'domain' part, not ports (as the comment said). But had `Authority` and `Host` backwards! 🤦

src/shared/TestInfrastructure/Objects/TestOAuth2Server.cs
mjcheetham: This is to mimic real-world OAuth2 authorisation servers, which typically treat loopback addresses (`localhost`, `127.0.0.1`, `::1`) as special.

Redirect URLs are normally supposed to match exactly, including port number. This is the location where the authorisation server will redirect the user agent to, including the authorisation code in the query params for the client to consume.

For public client applications (ones that run on an end-user device), we need to get the authorisation code back to the client application from a user agent _somehow_. You can achieve this in two ways:

1. Using an application-specific URL like (`myapp://oauth`) that is registered with the operating system, or
2. Using a local loopback URL, and have the client application open a port listening for that HTTP request.

For option 2, you can either explicitly define a port number to use for the redirect, or leave it blank. Leaving it blank, the client will pick any random free port before it makes the initial authorisation request. Since the port may not be known ahead of time, the authorization server cannot restrict ahead of time.

Why not just pick an explicit port? Well.. which one should we use? 😜
Obviously one of the ephemeral ports (49152–65535), but there's of course a non-zero (although small) chance we pick a port that's in use by something else on the system...

src/shared/Core.Tests/Authentication/OAuth2ClientTests.cs
mjcheetham: Look a little higher up at line 73:

```csharp
server.AuthorizationEndpointInvoked += (_, request) =>
{
    IDictionary<string, string> actualParams = request.RequestUri.GetQueryParameters();
    Assert.True(actualParams.TryGetValue(OAuth2Constants.RedirectUriParameter, out string actualRedirectUri));
    Assert.Equal(expectedRedirectUrl, actualRedirectUri);
};
```

We are setting an event handler on the `AuthorizationEndpointInvoked` `event`, and performing the validation of the `redirect_uri` in there.

ldennington: Got it! Thanks for clarifying.

PR #1282, opened 2023-06-06
dotnet: update install from src script to target .NET 7

mjcheetham:
In #1274 we updated to target .NET 7, but didn't update the install from source script for Linux to use SDK 7.0.


ldennington (approved):
Teamwork makes the dream work.

PR #1285, opened 2023-06-08
Add the `user` scope to OAuth token requests for GitHub

mjcheetham:
Add the `user` scope to OAuth token requests for GitHub, to better align with other tooling that uses the token generated by GCM, including Visual Studio.

The next time a token needs to be generated, the user will be asked to re-consent to include the new scope. This will not invalidate any existing tokens.


ldennington (approved):

mjcheetham:
Closing PR as the request was dropped.

PR #1288, opened 2023-06-09
Read new `wwwauth[]` input from Git 2.41

mjcheetham:
Add the new `wwwauth[]` credential property to the input arguments surfaced to providers and commands. This optional property is available as of Git 2.41.

```
Git v2.41 Release Notes
=======================

UI, Workflows & Features

 * Allow information carried on the WWW-Authenticate header to be
   passed to the credential helpers.
```

https://lore.kernel.org/git/xmqqleh3a3wm.fsf@gitster.g/

This new property may contain multiple values, as identified by a trailing `[]` on the property name. Extend our `Dictionary/Stream(Reader|Writer)` methods to allow for these multi-valued properties and add a convenience property to the `InputArguments` to get `wwwauth[]`.

### DIFF
--- a/src/shared/Core.Tests/StreamExtensionsTests.cs
+++ b/src/shared/Core.Tests/StreamExtensionsTests.cs
@@ -33,9 +33,9 @@ namespace GitCredentialManager.Tests
 
             Assert.NotNull(output);
             Assert.Equal(3, output.Count);
-            Assert.Contains(KeyValuePair.Create("a", "1"), output);
-            Assert.Contains(KeyValuePair.Create("b", "2"), output);
-            Assert.Contains(KeyValuePair.Create("c", "3"), output);
+            AssertDictionary("1", "a", output);
+            AssertDictionary("2", "b", output);
+            AssertDictionary("3", "c", output);
         }
 
         [Fact]
@@ -47,9 +47,9 @@ namespace GitCredentialManager.Tests
 
             Assert.NotNull(output);
             Assert.Equal(3, output.Count);
-            Assert.Contains(KeyValuePair.Create("a", "1"), output);
-            Assert.Contains(KeyValuePair.Create("b", "2"), output);
-            Assert.Contains(KeyValuePair.Create("c", "3"), output);
+            AssertDictionary("1", "a", output);
+            AssertDictionary("2", "b", output);
+            AssertDictionary("3", "c", output);
         }
 
         [Fact]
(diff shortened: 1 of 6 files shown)


src/shared/Core/Trace.cs
```diff
@@ -215,13 +216,25 @@ public void Flush()
             {
                 bool isSecretEntry = !(secretKeys is null) &&
                                      secretKeys.Contains(entry.Key, keyComparer ?? EqualityComparer<TKey>.Default);
-                if (isSecretEntry && !this.IsSecretTracingEnabled)
+
+                void WriteSecretLine(object value)
```
ldennington: I wasn't aware you could do this. Interesting!

src/shared/Core/InputArguments.cs
```diff
@@ -15,17 +15,24 @@ namespace GitCredentialManager
...
         {
-            if (dict == null)
-            {
-                throw new ArgumentNullException(nameof(dict));
-            }
+            EnsureArgument.NotNull(dict, nameof(dict));
+
+            // Transform input from 1:1 to 1:n and store as readonly
+            _dict = new ReadOnlyDictionary<string, IList<string>>(
+                dict.ToDictionary(x => x.Key, x => (IList<string>)new[] { x.Value })
```
ldennington, line 26: Elegant!

src/shared/Core.Tests/StreamExtensionsTests.cs
```diff
@@ -207,6 +402,20 @@ private static string WriteStringStream(IDictionary<string, string> input, Actio
             return output.ToString();
         }
 
+        private static void AssertDictionary(string expectedValue, string key, IDictionary<string, string> dict)
```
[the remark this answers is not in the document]
ldennington: Ok, I'm fine with that.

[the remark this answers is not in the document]
mjcheetham: Pushed a new commit on top that now uses the `AssertDictionary` helper method.

src/shared/Core/Trace.cs
mjcheetham: Do you mean the use of [local functions](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/local-functions)?

ldennington: Yep!

PR #1289, opened 2023-06-09
Prefer authority from `wwwauth[]` if present for Azure Repos

mjcheetham:
Now that Git can forward on the `WWW-Authenticate` headers from the remote, we no longer need to make a second, redundant, `HEAD` query to the organisation URL to rediscover the header values.

If we have been provided such a header from Git, use it in the first instance! Otherwise continue to make a `HEAD` query (and cache it).

Note this only affects the OAuth mode, just as the authority caching only works here.


ldennington (approved):
Great stuff! 🚀

PR #1291, opened 2023-06-13
install from source: designate arch linux as unsupported

ldennington:
Arch Linux was incorrectly added to the install-from-source script and
accompanying validation as a .NET-supported distribution, when this is not
actually the case. A list of supported distributions can be found here:
  
https://learn.microsoft.com/en-us/dotnet/core/install/linux
  
Instead of removing unsupported distros, we plan to print out a message
notifying users that, while they can be installed, they are not officially
validated or supported by the GCM project.

### DIFF
--- a/src/linux/Packaging.Linux/install-from-source.sh
+++ b/src/linux/Packaging.Linux/install-from-source.sh
@@ -114,6 +114,14 @@ apt_install() {
     $sudo_cmd apt install $pkg_name -y 2>/dev/null
 }
 
+print_unsupported_distro() {
+    prefix=$1
+    distro=$2
+
+    echo "$prefix: $distro is not officially supported by the GCM project."
+    echo "See https://gh.io/gcm/linux for details."
+}
+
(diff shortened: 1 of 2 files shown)


src/linux/Packaging.Linux/install-from-source.sh
```diff
@@ -114,6 +114,14 @@ apt_install() {
     $sudo_cmd apt install $pkg_name -y 2>/dev/null
 }
 
+print_unsupported_distro() {
+    prefix=$1
+    distro=$2
+
+    echo "$prefix: $distro is not officially supported by the GCM Project."
```
mjcheetham, line 121: Lowercase project 'p'?

```suggestion
    echo "$prefix: $distro is not officially supported by the GCM project."
```

mjcheetham (changes requested):
Just a nit and suggestion to use a shortlink.

mjcheetham (approved):

PR #1299, opened 2023-06-20
Clarify to install GCM for Windows for WSL

philsttr:
Previously it wasn't clear whether to install GCM for Linux or Windows when using WSL until later in the document.

Now the first step explicitly says GCM for Windows.


mjcheetham (approved):

mjcheetham:
Thanks for this update / clarification @philsttr !

Issue #1298, opened 2023-06-19
Installation script resulted in unintended Fedora upgrade


JWShaw:
Running the `linux-install-source.sh` script resulted in `dnf update -y` being executed, which caused an undesired upgrade from Fedora 36 to Fedora 37 on my machine.  Can we omit the `-y` flag to prevent similar occurrences in the future?

It's also worth noting that, according to the `dnf` manpages, `update` is a deprecated alias for `upgrade`

JWShaw:
I was mistaken about the system upgrade: I was already running Fedora 37, and my packages were just several weeks out of date.

Nonetheless, is the non-interactive flag required/desirable?

ldennington:
Hi @JWShaw - glad to hear that the system upgrade wasn't caused by the script, as that would certainly be unexpected! We purposefully use the non-interactive flag after printing a message at the beginning of the script indicating that we're going to install things and asking the user to accept:

```
This script will download, compile, and install Git Credential Manager to:

    /usr/local/bin

Git Credential Manager is licensed under the MIT License: https://aka.ms/gcm/license
Do you want to continue? [Y/n] 
```

If you have thoughts for improvement on the message or would like to help transition us away from the deprecated `upgrade` command, we'd welcome a PR! You can find the script [here](https://github.com/git-ecosystem/git-credential-manager/blob/main/src/linux/Packaging.Linux/install-from-source.sh) if you're interested.

PR #1300, opened 2023-06-20
Replace DNF's deprecated 'dnf update' alias with 'dnf upgrade'

JWShaw:
Addresses #1298.  According to the [DNF manpages](https://man7.org/linux/man-pages/man8/dnf.8.html), `update` is a deprecated alias.

### DIFF
--- a/src/linux/Packaging.Linux/install-from-source.sh
+++ b/src/linux/Packaging.Linux/install-from-source.sh
@@ -163,7 +163,7 @@ case "$distribution" in
         fi
     ;;
     fedora | centos | rhel)
-        $sudo_cmd dnf update -y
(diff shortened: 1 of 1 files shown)


ldennington:
Thank you for your contribution! To better understand the scope of this change, can you provide the version of RHEL/Fedora in which the `update` command was initially deprecated?

JWShaw:
I've not been successful in finding the exact version of RHEL/Fedora in which `update` was initially deprecated, but I found by perusing the `dnf` repository that `dnf update` was deprecated by a [commit on November 12, 2012](https://github.com/rpm-software-management/dnf/commit/91a92b01172ce03ff2cc8836622842c48534d905).

ldennington (approved):

Issue #1297, opened 2023-06-18
Error authenticating to Azure DevOps with OAuth "Application ... is configured for use by Azure Active Directory users only"


hickford:
I tried to authenticate to Azure DevOps with `credential.azreposCredentialType=oauth`. I'm using GCM 2.1.2 on Linux (installed with dotnet tool):

    echo protocol=https\nhost=dev.azure.com\npath=matthickford0459/testdevazuredotcom/_git/testdevazuredotcom\nusername=matthickford0459\nwwwauth[]=Basic realm=\"https://tfsproduks1.visualstudio.com/\" | git -c credential.azreposCredentialType=oauth credential-manager get

gave me error

> fatal: AADSTS9002332: Application '872cd9fa-d31f-45e0-9eab-6e460a02d1f1'(Visual Studio - Legacy) is configured for use by Azure Active Directory users only. Please do not use the /consumers endpoint to serve this request.

GCM trace:

<details>
<pre>
16:04:33.468133 ...re/Application.cs:95 trace: [RunInternalAsync] Version: 2.1.2.0
16:04:33.525134 ...re/Application.cs:96 trace: [RunInternalAsync] Runtime: .NET 6.0.16
16:04:33.525284 ...re/Application.cs:97 trace: [RunInternalAsync] Platform: Linux (x86-64)
16:04:33.525326 ...re/Application.cs:98 trace: [RunInternalAsync] OSVersion: Linux penguin 5.15.112-19373-g62629859bf76 #1 SMP PREEMPT Sun Jun 11 18:47:30 PDT 2023 x86_64 GNU/Linux
16:04:33.525474 ...re/Application.cs:99 trace: [RunInternalAsync] AppPath: /home/matthickford/.dotnet/tools/git-credential-manager
16:04:33.525783 ...e/Application.cs:100 trace: [RunInternalAsync] InstallDir: /home/matthickford/.dotnet/tools/.store/git-credential-manager/2.1.2/git-credential-manager/2.1.2/tools/net6.0/any/
16:04:33.525965 ...e/Application.cs:101 trace: [RunInternalAsync] Arguments: get
16:04:33.893875 ...GitCommandBase.cs:33 trace: [ExecuteAsync] Start 'get' command...
16:04:33.978260 ...GitCommandBase.cs:47 trace: [ExecuteAsync] Detecting host provider for input:
16:04:33.998439 ...GitCommandBase.cs:48 trace: [ExecuteAsync] 	protocol=https
[...]

PR #1321, opened 2023-07-10
Workaround MSAL.NET issue with MSA-PT account silent auth

mjcheetham:
When we have a Microsoft Account (MSA) in the cache and attempt to do a silent authentication, if we're an MSA-PT app we need to specify the special MSA transfer tenant ID to make sure we get the a token silently, correctly.

See the [issue](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3077) in the MSAL repo for more information.

Fixes: #1297

### DIFF
--- a/src/shared/Core/Authentication/MicrosoftAuthentication.cs
+++ b/src/shared/Core/Authentication/MicrosoftAuthentication.cs
@@ -23,7 +23,7 @@ namespace GitCredentialManager.Authentication
     public interface IMicrosoftAuthentication
     {
         Task<IMicrosoftAuthenticationResult> GetTokenAsync(string authority, string clientId, Uri redirectUri,
-            string[] scopes, string userName);
+            string[] scopes, string userName, bool msaPt = false);
     }
 
     public interface IMicrosoftAuthenticationResult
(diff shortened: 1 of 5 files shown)


src/shared/Core/Authentication/MicrosoftAuthentication.cs
```diff
@@ -295,9 +297,27 @@ private async Task<AuthenticationResult> GetAccessTokenSilentlyAsync(IPublicClie
...
+                    if (account is null)
+                    {
+                        Context.Trace.WriteLine($"No cached account found for user '{userName}'...");
+                        return null;
+                    }
+
+                    var atsBuilder = app.AcquireTokenSilent(scopes, account);
+
+                    // Is we are operating with an MSA passthrough app we need to ensure that we target the
+                    // special MSA 'transfer' tenant explicitly. This is a workaround for MSAL issue:
```
ldennington, line 312: Super nit: I'm wondering whether there's any public documentation available about the transfer tenant that we can link here for those who would like to learn more.

ldennington (approved):
Nice! ✨

src/shared/Core/Authentication/MicrosoftAuthentication.cs
mjcheetham: Unfortunately MSA passthrough is an internal (and legacy) feature of the Microsoft Identity platform, and the only docs are for Microsoft employees only: https://review.learn.microsoft.com/en-us/identity/microsoft-identity-platform/stack-choices?branch=main#msa-passthrough

mohamadfaqani:
***@***.***

Hello, I don't know why, but I don't receive emails in my Microsoft
account, no one knows how to fix it

gmail.android

ldennington (approved):

PR #1327, opened 2023-07-11
rename: remove git-credential-manager-core symlinks

### LINKED ISSUE #1322: rename: remove symlinks
Remove symlinks created to minimize user disruption during the rename of the `git-credential-manager-core` exe to `git-credential-manager`. This is in keeping with the [timeline outlined in the rename announcement](https://github.com/git-ecosystem/git-credential-manager/blob/main/docs/rename.md#rename-transition).

- [x] Remove symlinks and related code from codebase
- [x] Update `rename.md` with version in which symlinks were removed.

ldennington:
Since 2 versions of Git have released since the rename of the executable from `git-credential-manager-core` to `git-credential-manager`, remove the associated symlinks and warnings, (as outlined in `docs/rename.md`).

### DIFF
--- a/src/shared/Git-Credential-Manager/Program.cs
+++ b/src/shared/Git-Credential-Manager/Program.cs
@@ -55,36 +55,6 @@ namespace GitCredentialManager
                 // Write the start and version events
                 context.Trace2.Start(context.ApplicationPath, args);
 
-                //
-                // Git Credential Manager's executable used to be named "git-credential-manager-core" before
-                // dropping the "-core" suffix. In order to prevent "helper not found" errors for users who
-                // haven't updated their configuration, we include either a 'shim' or symlink with the old name
-                // that print warning messages about using the old name, and then continue execution of GCM.
-                //
-                // On Windows the shim is an exact copy of the main "git-credential-manager.exe" executable
-                // with the old name. We inspect argv[0] to see which executable we are launched as.
-                //
-                // On UNIX systems we do the same check, except instead of a copy we use a symlink.
-                //
-                string appPath = context.ApplicationPath;
-                if (!string.IsNullOrWhiteSpace(appPath))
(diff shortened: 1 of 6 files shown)


src/shared/Git-Credential-Manager/Program.cs
```diff
@@ -76,13 +76,6 @@ private static void AppMain(object o)
                     {
                         appPath = appPath.Substring(0, appPath.Length - 4);
                     }
```
mjcheetham, line 78: We can also delete the containing block and commentary above.

mjcheetham (changes requested):
Looks good! I think we can remove some more in _Program.cs_ too.
Also, I wonder if we should update the rename doc to say that the old symlink/name was retired in 2.3.x.

src/shared/Git-Credential-Manager/Program.cs
ldennington: Oops. That's so clear in Rider, but was less so in VS Code.

ldennington:
>I wonder if we should update the rename doc to say that the old symlink/name was retired in 2.3.x.

I don't want to be in the same situation we were before where we have a placeholder hanging around. Instead, how about I leave the issue open (I've removed the **Fixes** from the PR description) until this is released with an outstanding item to update `rename.md` with the correct version?

mjcheetham:
> > I wonder if we should update the rename doc to say that the old symlink/name was retired in 2.3.x.
> 
> @mjcheetham - I don't want to be in the same situation we were before where we have a placeholder hanging around. Instead, how about I leave the issue open (I've removed the **Fixes** from the PR description) until this is released with an outstanding item to update `rename.md` with the correct version?

Sure, that works too!

mjcheetham (approved):

PR #1328, opened 2023-07-11
Only match credential entries with correct namespace in the Windows Credential Manager

### LINKED ISSUE #1325: Select an account prompt showing duplicate "Personal Access Token" options when Visual Studio is also signed-in with GitHub
### Feature description

I have multiple Git accounts.
When the popup Select Account shows up... I have the choice between "Personal Access token" and "Personal Access token".
There is no way to make any distinctions.

See screenshot

image

mjcheetham:
GCM by default creates entries in the Windows Credential Manager on Windows, and prefixes the 'target name' of the entry with "git:". This 'namespace' prefix is configurable, but is not often changed in practice outside of tests.

Visual Studio, when adding GitHub accounts (either natively or by the older GitHub extension for VS), it creates three credential entries:

1. GitHub for Visual Studio - https://github.com
2. git:https://github.com
3. https://github.com

Entry 1 is used by VS for it's own purposes. Entry 2 is created for the benefit for GCM, so that we are 'primed'. It is unknown what entry 3 is for at this time.

There is an error in our existing logic for enumerating credentials that is also matching entry 3 as well as the expected entry 2.

Modify and fix the matching logic to ensure that the namespace prefix matches, rather than just stripping it and matching (even if it doesn't exist!).

Fixes #1325

---

**Bug repro instructions:**

1. Open Visual Studio
2. File > Account Settings
3. Add a GitHub account
4. Open a terminal (inside or outside of VS) and attempt to clone/fetch/push to or from a private GitHub repository.

At this point a window should appear asking you to select between two "Personal Access Token" accounts.

After installing [the bits from this PR build (artifacts > win-x86)](https://github.com/git-ecosystem/git-credential-manager/pull/1328/checks), attempting step 4 should **no longer** result in a prompt to select between two "Personal Access Token" accounts.

### DIFF
--- a/src/shared/Core.Tests/Interop/Windows/WindowsCredentialManagerTests.cs
+++ b/src/shared/Core.Tests/Interop/Windows/WindowsCredentialManagerTests.cs
@@ -270,6 +270,70 @@ namespace GitCredentialManager.Tests.Interop.Windows
             Assert.Equal(expected, actual);
         }
 
+        [PlatformFact(Platforms.Windows)]
+        public void WindowsCredentialManager_IsMatch_NoNamespace_NotMatched()
+        {
+            var win32Cred = new Win32Credential
+            {
+                UserName = "test",
+                TargetName = $"{WindowsCredentialManager.TargetNameLegacyGenericPrefix}https://example.com"
+            };
+
+            var credManager = new WindowsCredentialManager(TestNamespace);
+
+            bool result = credManager.IsMatch("https://example.com", null, win32Cred);
+
+            Assert.False(result);
+        }
+
+        [PlatformFact(Platforms.Windows)]
+        public void WindowsCredentialManager_IsMatch_DifferentNamespace_NotMatched()
+        {
+            var win32Cred = new Win32Credential
+            {
+                UserName = "test",
+                TargetName = $"{WindowsCredentialManager.TargetNameLegacyGenericPrefix}:random-namespace:https://example.com"
+            };
+
+            var credManager = new WindowsCredentialManager(TestNamespace);
+
+            bool result = credManager.IsMatch("https://example.com", null, win32Cred);
+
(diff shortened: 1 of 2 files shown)


dscho (commented):
In https://github.com/git-ecosystem/git-credential-manager/pull/1328/commits/de7b93ba5d2066c0c1980ea163c95f3239dc7cf3#diff-c4c58b08f112c67b1d49742bdc348cbfc3dd410fb2e46ecfee795175d3cd5e4bL268, the modified method is marked with the comment `/* for testing */`, which looks a bit funny to me: is this comment obsolete? Or is this method really used only for testing?

Speaking of testing, is there a way to ensure that this does the right thing by adding a test?

mjcheetham:
> the modified method is marked with the comment `/* for testing */`, which looks a bit funny to me: is this comment obsolete? Or is this method really used only for testing?

The method is not used _for_ testing, but the **`internal` visibility modifier** is because this method is [called from unit tests](https://github.com/git-ecosystem/git-credential-manager/blob/a0a7d948a1c5abd022b8cbd35549c56297e2f7ae/src/shared/Core.Tests/Interop/Windows/WindowsCredentialManagerTests.cs#L238-L271).

> Speaking of testing, is there a way to ensure that this does the right thing by adding a test?

Although we have tests for all sorts of matching, we never tested the matching of no-namespace credential entries.. largely because we cannot be sure the test environment is clean and devoid of credentials. Local developer machines may be less 'clean' than CI, so failures may be more limited than expected. I can add a test for `https://example.com` and `git-test:https://example.com`.

mjcheetham:
@dscho I have now added extra tests around this area.

derrickstolee (approved):

ldennington (approved):
New tests are great, and I got the fix working on my machine.

PR #1330, opened 2023-07-12
Use localhost redirect URI for GHES instances

### LINKED ISSUE #1329: GitHub Enterprise redirects to `localhost:80` instead of `127.0.0.1:60167`
### Version

2.2.1

### Operating system

Windows

### OS version or distribution

Windows 11

### Git hosting provider(s)

GitHub Enterprise Server

### Other hosting provider

GitHub Enterprise Server 3.8.2

### (Azure DevOps only) What format is your remote URL?

None

### Can you access the remote repository directly in the browser?

Yes, I can access the repository

### Expected behavior

1. User clears their authentication settings in Windows Credential Manager (due to [this issue](https://github.com/git-ecosystem/git-credential-manager/issues/1325#issuecomment-1631642589)).
2. User runs any git command that requires auth to our GHES instance.
3. User receives the expected "Sign in with Browser" popup and enters our GHES instance information, clicking the link.
4. The link takes the user through the OAuth operation and the user is signed in.

### Actual behavior

On step 2 the initial GET URL will look like

```
GET https://OUR_GHES_SERVER_URL/login/oauth/authorize?client_id=0120e057bd645470c1ed&code_challenge=2NKS...&code_challenge_method=S256&redirect_uri=http://127.0.0.1:60167/&response_type=code&scope=repo gist workflow&state=36BUNCHOALPHANUMERICS9dd
```

Note the redirect URL of `http://127.0.0.1:60167` [as described in this change](https://github.com/git-ecosystem/git-credential-manager/issues/594). The server will then respond with 

```
302 Found
Location: http://localhost/?error=redirect_uri_mismatch&error_description=The+redirect_uri+MUST+match+the+registered+callback+URL+for+this+application.&error_uri=https%3A%2F%2Fdocs.github.com%2Fenterprise-server%403.8%2Fapps%2Fmanaging-oauth-apps%2Ftroubleshooting-authorization-request-errors%2F%23redirect-uri-mismatch&state=36BUNCHOALPHANUMERICS9dd
```

The user is unable to log in using OAuth. 

### Logs

Our current workaround for this is to suggest `Log in with Code` instead.

We are currently working to correlate affected users with possible machine configuration issues (many of these folks are web developers) that may be altering local URLs. This issue began this week as folks are updating to the latest version of Git for Windows.

Running `diagnose` and `GCM_TRACE=1` did not reveal interesting information, just the tool correctly determining the GHE version and the `Pat, OAuth` schemes.

mjcheetham:
For GitHub.com we've updated the redirect URI to 127.0.0.1, whilst also keeping the localhost variant around for backwards compatibility with older GCM clients.

However, since GHES has not been updated with the new 127.0.0.1 redirect, and older GHES servers will be stuck with the old redirect we must continue to use the localhost redirect on the client for non-dotcom targets.

Fixes #1329

### DIFF
--- a/src/shared/GitHub/GitHubConstants.cs
+++ b/src/shared/GitHub/GitHubConstants.cs
@@ -16,6 +16,7 @@ namespace GitHub
         // [SuppressMessage("Microsoft.Security", "CS002:SecretInNextLine", Justification="OAuth2 public client application 'secrets' are required and permitted to be public")]
         public const string OAuthClientSecret = "18867509d956965542b521a529a79bb883344c90";
         public static readonly Uri OAuthRedirectUri = new Uri("http://127.0.0.1/"); // Note that the trailing slash is important!
+        public static readonly Uri OAuthLegacyRedirectUri = new Uri("http://localhost/"); // Note that the trailing slash is important!
         public static readonly Uri OAuthAuthorizationEndpointRelativeUri = new Uri("/login/oauth/authorize", UriKind.Relative);
         public static readonly Uri OAuthTokenEndpointRelativeUri = new Uri("/login/oauth/access_token", UriKind.Relative);
         public static readonly Uri OAuthDeviceEndpointRelativeUri = new Uri("/login/device/code", UriKind.Relative);
--- a/src/shared/GitHub/GitHubOAuth2Client.cs
+++ b/src/shared/GitHub/GitHubOAuth2Client.cs
@@ -9,7 +9,7 @@ namespace GitHub
     {
         public GitHubOAuth2Client(HttpClient httpClient, ISettings settings, Uri baseUri, ITrace2 trace2)
             : base(httpClient, CreateEndpoints(baseUri),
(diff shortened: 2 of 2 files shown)


jeffhostetler (approved):

vdye (approved):
The commit message + comments explain the issue and implementation clearly. Everything looks good to me! 🚢

vdye (commented):
Actually, one other note - is `GitHubOAuth2Client` unit tested? If not, could/should it be?

mjcheetham:
> is `GitHubOAuth2Client` unit tested? If not, could/should it be?

`GitHubOAuth2Client` is not unit tested, but the base class `OAuth2Client` is heavily unit tested.

Unit testing would not have _necessarily_ caught this sort of issue. The oversight here was that there are different "GitHub" environments (dotcom, AE, ES) and each of them may have different server-side configuration (and with GHES, different versions have different configuration over time). We forgot to do a real-world test against the latest GHES, and haven't yet had a new release of GHES that updates the redirect URI.

End-to-end testing against a real GHES instance, latest plus N older versions, would be ideal and have caught this.

PR #1363, opened 2023-08-01
Update rename doc to mention symlink removal

mjcheetham:
In GCM 2.3 we dropped the symlinks.
Let's update our docs on the issue to reflect this.

Rendered: https://github.com/git-ecosystem/git-credential-manager/blob/rename-docs-update/docs/rename.md


ldennington (approved):
:shipit:

PR #1372, opened 2023-08-07
Support managed identities and service principals

### LINKED ISSUE #1313: Support Azure managed identity authentication with AzureDevOps repos
### Feature description

AzureDevOps released in public preview a feature to enable [authentication with Azure Active Directory service principals & managed identities](https://learn.microsoft.com/en-us/azure/devops/integrate/get-started/authentication/service-principal-managed-identity?view=azure-devops) and they are encouraging people to use for automation instead of PATs.

It would be useful to have support for managed identities authentication in git-credential-manager, in the same way it already supports AAD authentication with users.
The managed identity client id could be provided in the configuration.

This would greatly simplify setting up a repo for automation, and everything would seemlessly after setting up the git-credential-manager configuration.

mjcheetham:
Add support for both managed identities and service principals for Azure Repos authentication.

Service principals are the "service account" equivalent to normal "users" (user principals). Rather than a password+MFA, service principals (SP) authenticate either with a 'secret' or a certificate - we support both.

Managed identities (MI) are an evolution of service principals whereby the secret/certificate are hidden and managed by Azure. There are two types of managed identities: system-assigned and user-assigned. System-assigned are tied to a particular VM or resource, whereas user-assigned can be shared between VMs/resources.

Azure DevOps recently learned to support SPs and MIs as identities that can be members of orgs and projects. Note that this only applies to AAD-connected Azure DevOps orgs.

This functionality is most compelling for automated scenarios, like CI machines, that need access to Azure Repos.

Fixes #1313

### DIFF
--- a/src/shared/Core/Authentication/MicrosoftAuthentication.cs
+++ b/src/shared/Core/Authentication/MicrosoftAuthentication.cs
@@ -13,6 +13,7 @@ using System.Threading;
 using GitCredentialManager.UI;
 using GitCredentialManager.UI.ViewModels;
 using GitCredentialManager.UI.Views;
+using Microsoft.Identity.Client.AppConfig;
 
 #if NETFRAMEWORK
 using System.Drawing;
@@ -44,6 +45,25 @@ namespace GitCredentialManager.Authentication
         /// <param name="scopes">Scopes to request.</param>
         /// <returns>Authentication result.</returns>
         Task<IMicrosoftAuthenticationResult> GetTokenForServicePrincipalAsync(ServicePrincipalIdentity sp, string[] scopes);
+
+        /// <summary>
+        /// Acquire a token using the managed identity in the current environment.
+        /// </summary>
+        /// <param name="managedIdentity">Managed identity to use.</param>
+        /// <param name="resource">Resource to obtain an access token for.</param>
+        /// <returns>Authentication result including access token.</returns>
+        /// <remarks>
+        /// There are several formats for the <paramref name="managedIdentity"/> parameter:
+        /// <para/>
+        ///  - <c>"system"</c> - Use the system-assigned managed identity.
+        /// <para/>
+        ///  - <c>"{guid}"</c> - Use the user-assigned managed identity with client ID <c>{guid}</c>.
+        /// <para/>
+        ///  - <c>"id://{guid}"</c> - Use the user-assigned managed identity with client ID <c>{guid}</c>.
+        /// <para/>
+        ///  - <c>"resource://{guid}"</c> - Use the user-assigned managed identity with resource ID <c>{guid}</c>.
+        /// </remarks>
+        Task<IMicrosoftAuthenticationResult> GetTokenForManagedIdentityAsync(string managedIdentity, string resource);
     }
 
     public class ServicePrincipalIdentity
@@ -260,6 +280,31 @@ namespace GitCredentialManager.Authentication
             catch (Exception ex)
             {
                 Context.Trace.WriteLine($"Failed to acquire token for service principal '{sp.TenantId}/{sp.TenantId}'.");
+                Context.Trace.WriteException(ex);
+                throw;
+            }
+        }
+
+        public async Task<IMicrosoftAuthenticationResult> GetTokenForManagedIdentityAsync(string managedIdentity, string resource)
+        {
+            var httpFactoryAdaptor = new MsalHttpClientFactoryAdaptor(Context.HttpClientFactory);
+
+            ManagedIdentityId mid = GetManagedIdentity(managedIdentity);
+
+            IManagedIdentityApplication app = ManagedIdentityApplicationBuilder.Create(mid)
+                .WithHttpClientFactory(httpFactoryAdaptor)
+                .Build();
+
+            try
+            {
+                AuthenticationResult result = await app.AcquireTokenForManagedIdentity(resource).ExecuteAsync();
+                return new MsalResult(result);
+            }
+            catch (Exception ex)
+            {
+                Context.Trace.WriteLine(mid == ManagedIdentityId.SystemAssigned
+                    ? "Failed to acquire token for system managed identity."
+                    : $"Failed to acquire token for user managed identity '{managedIdentity:D}'.");
                 Context.Trace.WriteException(ex);
                 throw;
             }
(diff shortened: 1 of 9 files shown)


mjcheetham:
@bgavrilMS would you or someone in your team be able to have a brief look over the Service Principal and Managed Identity parts in _MicrosoftAuthentication.cs_? Thank you!

src/shared/Core/Authentication/MicrosoftAuthentication.cs
```diff
@@ -23,8 +25,66 @@ namespace GitCredentialManager.Authentication
...
+        /// <returns>Authentication result including access token.</returns>
+        Task<IMicrosoftAuthenticationResult> GetTokenForManagedIdentityAsync(Guid managedIdentity, string resource);
+    }
+
+    public class ServicePrincipalIdentity
+    {
+        /// <summary>
+        /// Client ID of the service principal.
+        /// </summary>
+        public string Id { get; set; }
```
bgavrilMS, line 66: Hmm, not sure why this is needed?

src/shared/Core/Authentication/MicrosoftAuthentication.cs
```diff
@@ -23,8 +25,66 @@ namespace GitCredentialManager.Authentication
...
+        /// <param name="sp">Service principal identity.</param>
+        /// <param name="scopes">Scopes to request.</param>
+        /// <returns>Authentication result.</returns>
+        Task<IMicrosoftAuthenticationResult> GetTokenForServicePrincipalAsync(ServicePrincipalIdentity sp, string[] scopes);
+
+        /// <summary>
+        /// Acquire a token using the managed identity in the current environment.
+        /// </summary>
+        /// <param name="managedIdentity">
+        /// Client ID of the user-assigned managed identity.
```
bgavrilMS, line 53: Excellent abstraction :)

bgavrilMS: FYI - In MSAL we went with a more verbose API because you can specify the MI in different ways: 

- by client id
- by resource id
- by object id
- by SP id

There are scenarios where you want to use resource id instead of client id. But anyway, you can address the scenarios later.

src/shared/Core/Authentication/MicrosoftAuthentication.cs
```diff
@@ -418,11 +522,42 @@ internal MicrosoftAuthenticationFlowType GetFlowType()
...
+
+            return app;
+        }
+
         #endregion
 
         #region Helpers
 
-        private async Task RegisterTokenCacheAsync(IPublicClientApplication app, ITrace2 trace2)
+        private async Task RegisterTokenCacheAsync(IClientApplicationBase app, ITrace2 trace2)
```
bgavrilMS, line 560: For client_credentials, the tokens are placed in `app.AppTokenCache`. We recommend not mixing AppTokenCache and UserTokenCache, i.e. just use another file for app token cache.

src/shared/Core/X509Utils.cs
```diff
@@ -0,0 +1,23 @@
+using System.Security.Cryptography.X509Certificates;
+
+namespace GitCredentialManager;
+
+public static class X509Utils
+{
+    public static X509Certificate2 GetCertificateByThumbprint(string thumbprint)
```
bgavrilMS, line 7: Here's a slight more battle tested way to load certs. I think this is fine for loading by thumbprint, but if loading by name you should exclude expired certs and load the most fresh.

https://github.com/AzureAD/microsoft-identity-web/blob/96323e40bc6e6610192461f7b1dfaf4c2605ff21/src/Microsoft.Identity.Web.Certificate/CertificateLoaderHelper.cs#L55

bgavrilMS (changes requested):
Ensure that you serialize `cca.AppTokenCache` for client_credentials.

src/shared/Core/Authentication/MicrosoftAuthentication.cs
mjcheetham: From my understanding, a MI is really just an SP under the hood. Is this correct? It's how it appears when you go digging in the AAD portal.

Also it's my understanding that:

- Client ID = ID of the "application"
- SP ID = ID of the materialised SP in a particular tenant

The application (or MI) owns the credentials (secret or cert).

What is the resource ID? Could this be, for example, the object ID of the Virtual Machine resource that has a system-assigned MI? I can certainly see how this may be useful (since you may not know what the system MI OID is for a resource, like a VM).

How does that work for user-assigned MIs?

The MSAL API only shows me two options:

- `WithUserAssignedClientId`
- `WithUserAssignedResourceId`

src/shared/Core/X509Utils.cs
mjcheetham: Is there a compelling argument for allowing people to specify something other than a thumbprint (like the cert name, or file path)? Are there common situations where the thumbprint would not be known?

I guess at certificate renewal a new thumbprint would be created, but the certificate would have the same name?

Right now we only have `credential.azreposServicePrincipalCertificateThumbprint` / `GCM_AZREPOS_SP_CERT_THUMBPRINT` to specify a thumbprint.

We could add other options like `..CertificatePath` / `.._CERT_PATH` and `..CertificateName` / `.._CERT_NAME` in the future (unless you think it's useful from the get go!) :)

mjcheetham:
> Ensure that you serialize `cca.AppTokenCache` for client_credentials.

Thanks for spotting this! I've pushed an update that should now correctly serialise the `AppTokenCache` and `UserTokenCache`s separately. Had a question on if there's any known shared AppTokensCaches, akin to the VS/dev-tools one for user tokens.

Eye69wives (commented):
Full access 

> Editing

src/shared/Microsoft.AzureRepos/AzureReposHostProvider.cs
```diff
@@ -461,6 +493,89 @@ private bool UsePersonalAccessTokens()
...
+            {
+                sp = null;
+                return false;
+            }
+
+            string[] split = spStr.Split(new[] { '/' }, count: 2);
+
+            if (split.Length < 2 || string.IsNullOrWhiteSpace(split[0]))
+            {
+                _context.Streams.Error.WriteLine("error: unable to use configured service principal - missing tenant ID in configuration");
```
ldennington, line 512: Nit: if `split.Length` is less than 1, do we know for sure that the tenant ID is missing? Couldn't the client ID be the piece that is missing?

mjcheetham: True! Added guards and specific messages for missing tenant or client ID.

bayu889900:
> Add support for both managed identities and service principals for Azure Repos authentication.
> 
> Service principals are the "service account" equivalent to normal "users" (user principals). Rather than a password+MFA, service principals (SP) authenticate either with a 'secret' or a certificate - we support both.
> 
> Managed identities (MI) are an evolution of service principals whereby the secret/certificate are hidden and managed by Azure. There are two types of managed identities: system-assigned and user-assigned. System-assigned are tied to a particular VM or resource, whereas user-assigned can be shared between VMs/resources.
> 
> Azure DevOps recently learned to support SPs and MIs as identities that can be members of orgs and projects. Note that this only applies to AAD-connected Azure DevOps orgs.
> 
> This functionality is most compelling for automated scenarios, like CI machines, that need access to Azure Repos.
> 
> Fixes #1313

PR #1383, opened 2023-08-17
dependencies: update to Avalonia 11.0.4

de1acr0ix:
Update from Avalonia 11.0.0-preview6 to 11.0.4 and fix the codes using APIs removed since 11.0.0 release.

### DIFF
--- a/src/shared/Core/Core.csproj
+++ b/src/shared/Core/Core.csproj
@@ -14,24 +14,24 @@
     <Reference Include="System.Net.Http" />
     <Reference Include="System.Web" />
     <PackageReference Include="Microsoft.Identity.Client.Broker" Version="4.54.0" />
-    <PackageReference Include="Avalonia.Win32" Version="11.0.0-preview6" />
+    <PackageReference Include="Avalonia.Win32" Version="11.0.3" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' != 'net472'">
-    <PackageReference Include="Avalonia.Desktop" Version="11.0.0-preview6" />
+    <PackageReference Include="Avalonia.Desktop" Version="11.0.3" />
   </ItemGroup>
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Identity.Client" Version="4.54.0" />
     <PackageReference Include="Microsoft.Identity.Client.Extensions.Msal" Version="2.28.0" />
     <PackageReference Include="System.CommandLine" Version="2.0.0-beta4.22272.1" />
-    <PackageReference Include="Avalonia" Version="11.0.0-preview6" />
-    <PackageReference Include="Avalonia.Skia" Version="11.0.0-preview6" />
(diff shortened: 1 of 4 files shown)


de1acr0ix:
Please don't approve this. I found GCM actually does not exit after GitHub OAuth workflow is complete.

mjcheetham:
Thanks for your contribution here!

> Please don't approve this. I found GCM actually does not exit after GitHub OAuth workflow is complete.

You are correct. A while ago I had been testing Avalonia using the 11.0 non-preview released bits and found the same issue. For some reason the signal to terminate the platform runloop is not working correctly, although I had confirmed it was being called. I suspect some deadlock, but never had the time to investigate much further.

We would ideally like some advice from Avalonia maintainers, on how to use Avalonia in GCM's somewhat strange scenario:

Starting as a CLI app, at some point during execution GCM _may_ wish to start up Avalonia and a UI thread. On macOS the UI thread _must_ be the entry thread, so we need to reserve that thread for Avalonia's dispatcher.

mjcheetham (changes requested):
Reviewing with 'Request changes' to block merge.

de1acr0ix:
> Thanks for your contribution here!
> 
> > Please don't approve this. I found GCM actually does not exit after GitHub OAuth workflow is complete.
> 
> You are correct. A while ago I had been testing Avalonia using the 11.0 non-preview released bits and found the same issue. For some reason the signal to terminate the platform runloop is not working correctly, although I had confirmed it was being called. I suspect some deadlock, but never had the time to investigate much further.
> 
> We would ideally like some advice from Avalonia maintainers, on how to use Avalonia in GCM's somewhat strange scenario:
> 
> Starting as a CLI app, at some point during execution GCM _may_ wish to start up Avalonia and a UI thread. On macOS the UI thread _must_ be the entry thread, so we need to reserve that thread for Avalonia's dispatcher.

I did some investigation and I now believe it is a bug of Avalonia. I have created a [PR](https://github.com/AvaloniaUI/Avalonia/pull/12579) for Avalonia and it shall fix this issue (I have verified locally on my Macbook). Once it is merged and a new release of Avalonia available, I will bump the version in this PR.

de1acr0ix:
@mjcheetham Avalonia team really is responsive. My changes got merged and new release 11.0.4 is available. I have bumped the version in this PR and tested locally.

mjcheetham (approved):
@de1acr0ix thank you for fixing the Avalonia issue and updating GCM! I've just also confirmed 11.0.4 works perfectly 🎉

de1acr0ix:
> @de1acr0ix thank you for fixing the Avalonia issue and updating GCM! I've just also confirmed 11.0.4 works perfectly 🎉

There is a change in Avalonia 11 I didn't adopt in this PR. Avalonia 11 can generate `InitializeComponent()` for `.axaml.cs` and fields for named controls. I would rather do it separately in a different PR.

PR #1409, opened 2023-09-12
Support linux install path from src

### LINKED ISSUE #1304: how to change the installation directory when installing from source
I don't have the root permission, so I tried to install from source. I used the command sh install-from-source.sh. 

However, it always install to the directory /usr/local/bin, which I cannot since I don't have the permission to write file into the directory.

Is there a way, like --prefix= to change the installation directory to my home directory?

m4ss1m0g:
This PR fix #1304.

I found that the add func [add_to_PATH](https://github.com/git-ecosystem/git-credential-manager/blob/afcb6b87302b34d1809a0a057f24c7ba7a3ff5db/src/linux/Packaging.Linux/install-from-source.sh#L94) doesn't persist the path to the $PATH env variable. 

To persist in the current shell is necessary to use a dot or `source` [see doc](https://ss64.com/bash/source.html)

_When a script is run using source it runs within the existing shell, any variables created or modified by the script will remain available after the script completes. In contrast if the script is run just as filename, then a separate subshell (with a completely separate set of variables) would be spawned to run the script._

To persist the PATH among the reboot/new shell, it is necessary to update the _.profile_ with `echo 'PATH="$PATH:$installPath' >>~/.profile`.

I don't know if this is the correct behavior, Prior to this commit, it was working because the `/usr/local/bin` is already in the PATH.

### DIFF
--- a/src/linux/Packaging.Linux/install-from-source.sh
+++ b/src/linux/Packaging.Linux/install-from-source.sh
@@ -6,6 +6,7 @@
 # for additional details.
 set -e
 
+installLocation=/usr/local
 is_ci=
 for i in "$@"; do
     case "$i" in
@@ -13,15 +14,26 @@ for i in "$@"; do
         is_ci=true
         shift # Past argument=value
         ;;
+        --install-location=*)
+        installLocation="${i#*=}"
+        shift # past argument=value
+        ;;
     esac
 done
+
+
+# If pass the install-location check if it exists
+if [! -d "$installLocation" ]; then
+    echo "The folder $installLocation do not exists"
+    exit
+fi
 
 # In non-ci scenarios, advertise what we will be doing and
 # give user the option to exit.
 if [ -z $is_ci ]; then
     echo "This script will download, compile, and install Git Credential Manager to:
 
-    /usr/local/bin
+    $installLocation/bin
 
 Git Credential Manager is licensed under the MIT License: https://aka.ms/gcm/license"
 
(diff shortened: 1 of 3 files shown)


ldennington:
@m4ss1m0g - thank you for your contributions! I made a few tweaks an submitted them [here](https://github.com/m4ss1m0g/git-credential-manager/pull/1). Once you've merged my PR, I think we'll be good to merge here.

src/linux/Packaging.Linux/install-from-source.sh
```diff
@@ -225,5 +240,5 @@ if [ -z "$DOTNET_ROOT" ]; then
 fi
 
 cd "$toplevel_path"
-$sudo_cmd env "PATH=$PATH" $DOTNET_ROOT/dotnet build ./src/linux/Packaging.Linux/Packaging.Linux.csproj -c Release -p:InstallFromSource=true
-add_to_PATH "/usr/local/bin"
+$sudo_cmd env "PATH=$PATH" $DOTNET_ROOT/dotnet build ./src/linux/Packaging.Linux/Packaging.Linux.csproj -c Release -p:InstallFromSource=true -p:InstallLocation=$installLocation
+add_to_PATH $installLocation
```
mjcheetham, line 244: `$installLocation` will be something like `/usr/local` but we want to add `/usr/local/bin` to the `$PATH`.. we should therefore be using something like this instead here:

```shell
add_to_PATH "${installLocation}/bin"
```

mjcheetham (changes requested):
Small issue with adding the correct directory to the path.

Also, we may wish to consider changing "install location" something like "prefix" or "install prefix", since this is a piece of common terminology in the realm of build/configure scripts etc.

src/linux/Packaging.Linux/install-from-source.sh
ldennington: Good catch! I also agree with the "prefix" name change.

ldennington: @m4ss1m0g - I opened https://github.com/m4ss1m0g/git-credential-manager/pull/2 to address these comments. Once merged, I think we will _actually_ be good to go here 😀.

m4ss1m0g: Thanks

ldennington (approved):

PR #1416, opened 2023-09-18
Update dev.azure.com product page URLs

mjcheetham:
Update the URL for the Azure DevOps product page. This used to be just https://dev.azure.com, but this now redirects you to https://azure.microsoft.com/en-us/products/devops instead.

The markdown link linter times out following the redirect, so let's just update the original link.


ldennington (approved):
Thank you!

PR #1419, opened 2023-09-20
Enable assembly trimming and single-file publish on all platforms

### LINKED ISSUE #113: Migrate from .NET Framework on Windows
We are currently not actually using the .NET Core on Windows. We should actually use the .NET Core runtime!

The reason we are currently using .NET Framework instead is because MSAL does (didn't) support embedded browser on Windows when targeting .NET Core or .NET Standard. We can use the `UseCustomWebUi` extension point to restore this functionality.

mjcheetham:
*Depends on: https://github.com/git-ecosystem/git-credential-manager/pull/1417 and https://github.com/git-ecosystem/git-credential-manager/pull/1418*

Now that we no longer have to target .NET Framework on Windows, and we are using the latest .NET (.NET 8) and CoreCLR, we can take advantage of the mature and stable assembly trimming logic in the SDK.

This will allow us to reduce the number and sizes of binaries that we produce and ship. In addition we enable single-file publishing on all platforms that links all CIL code in to the same entry executable. Native dependencies remain separate files, such as the MSAL runtime, and Avalonia graphics library dependencies.

In order to enable trimming we had to:
- Use source generation in System.Text.Json
- Enable compiled bindings in Avalonia components

** TODO: size comparisons **

Fixes: https://github.com/git-ecosystem/git-credential-manager/issues/113

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -29,6 +29,11 @@ jobs:
       with:
         dotnet-version: 8.0.x
 
+    - name: Setup .NET
+      uses: actions/setup-dotnet@v3.2.0
+      with:
+        dotnet-version: 8.0.x
+
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
       uses: github/codeql-action/init@v4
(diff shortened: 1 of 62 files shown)


mjcheetham:
Already completed.

PR #1441, opened 2023-10-19
recognise BitBucket hosts by WWW-Authenticate header

hickford:
Tested with https://bitbucket.hl7.org

### DIFF
--- a/src/shared/Atlassian.Bitbucket/BitbucketHostProvider.cs
+++ b/src/shared/Atlassian.Bitbucket/BitbucketHostProvider.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Net.Http;
 using System.Threading.Tasks;
 using Atlassian.Bitbucket.Cloud;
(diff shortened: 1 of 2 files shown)


src/shared/Atlassian.Bitbucket/BitbucketHostProvider.cs
```diff
@@ -43,6 +44,11 @@ public bool IsSupported(InputArguments input)
                 return false;
             }
 
+            if (input.WwwAuth.Any(x => x.Contains("realm=\"Atlassian Bitbucket\"")))
```
mjcheetham, line 47: I wonder if we want to be a little more broad in string matching and ignore the case?

mjcheetham (changes requested):
Great addition! One comment on ignoring case, but also please could we add a simple test to check that `IsSupported` returns `true` if the header matches? Thanks!

src/shared/Atlassian.Bitbucket/BitbucketHostProvider.cs
hickford: Good idea, done.

mjcheetham (approved):

PR #1444, opened 2023-10-20
default OAuth redirect URI to http://127.0.0.1

hickford:


mjcheetham (commented):
Good idea to set a sensible default! Just remember to also update the documentation around this, clarifying that it's now optional to set redirect URI (and that the default is 127.0.0.1).

Docs are in `docs/generic-oauth.md`

mjcheetham (approved):

ldennington (approved):

PR #1445, opened 2023-10-20
Add software rendering flag for Windows

mjcheetham:
Add ability to disable hardware acceleration when drawing Avalonia-based GUI prompts on Windows. This is a useful workaround for an issue on certain ARM64 devices where the windows appear empty.

https://github.com/AvaloniaUI/Avalonia/issues/10405
https://developercommunity.visualstudio.com/t/Git-authentication-dialog-is-invisible-o/10467795

### DIFF
--- a/docs/environment.md
+++ b/docs/environment.md
@@ -294,6 +294,10 @@ export GCM_GUI_SOFTWARE_RENDERING=1
 
 Defaults to false (use hardware acceleration where available).
 
+> [!NOTE]
+> Windows on ARM devices defaults to using software rendering to work around a
+> known Avalonia issue: <https://github.com/AvaloniaUI/Avalonia/issues/10405>
+
 **Also see: [credential.guiSoftwareRendering][credential-guisoftwarerendering]**
 
 ---
--- a/docs/environment.md
+++ b/docs/environment.md
@@ -272,6 +272,32 @@ Defaults to enabled.
 
 ---
 
+### GCM_GUI_SOFTWARE_RENDERING
+
+Force the use of software rendering for GUI prompts.
+
+This is currently only applicable on Windows.
+
+#### Example
+
+##### Windows
+
+```batch
+SET GCM_GUI_SOFTWARE_RENDERING=1
+```
+
+##### macOS/Linux
+
+```bash
+export GCM_GUI_SOFTWARE_RENDERING=1
+```
+
+Defaults to false (use hardware acceleration where available).
+
+**Also see: [credential.guiSoftwareRendering][credential-guisoftwarerendering]**
+
+---
+
 ### GCM_AUTODETECT_TIMEOUT
 
 Set the maximum length of time, in milliseconds, that GCM should wait for a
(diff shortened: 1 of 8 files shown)


shango420 (commented):
docs/configuration.md

src/shared/Core/Settings.cs
```diff
@@ -559,6 +564,22 @@ public bool IsInteractionAllowed
...
+            {
+                // WORKAROUND: Some Windows ARM devices have a graphics driver issue that causes transparent windows
+                // when using hardware rendering. Until this is fixed, we will default to software rendering on these
+                // devices. Users can always override this setting back to HW-accelerated rendering if they wish.
+                bool defaultValue = PlatformUtils.IsWindows() && PlatformUtils.IsArm();
+
+                return TryGetSetting(KnownEnvars.GcmGuiSoftwareRendering,
+                    KnownGitCfg.Credential.SectionName,
+                    KnownGitCfg.Credential.GuiSoftwareRendering,
+                        out string str) && str.ToBooleanyOrDefault(defaultValue);
```
ldennington, line 579: Does this `&&` mean this will be restricted only to devices where the `defaultValue` is `true` (i.e. Windows ARM devices)?

docs/environment.md
```diff
@@ -272,6 +272,36 @@ Defaults to enabled.
...
+##### macOS/Linux
+
+```bash
+export GCM_GUI_SOFTWARE_RENDERING=1
+```
+
+Defaults to false (use hardware acceleration where available).
+
+> [!NOTE]
+> Windows on ARM devices default to using software rendering to workaround a
```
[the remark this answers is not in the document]
rimrul: The wording here should probably be consistent with `docs/configuration.md`, so either both should say "default to" or both should say "defaults to".

src/shared/Core/Settings.cs
rimrul: This Code is certainly not ideal. It seems like `defaultValue` is intended to be used when no value (or a bogus value) is explicitly configured. But when no value is explicitly configured, `TryGetSetting()` will return `false` and `str` will be `null`, so the `&&` will short circuit and return `false`, despite `defaultValue` potentially being `true`.

mjcheetham: > It seems like defaultValue is intended to be used when no value (or a bogus value) is explicitly configured.

Correct. This was the _intent_.. but also correct that this isn't what it's doing! 🤦‍♂️ 

Updated the logic to return the default value either when there's no setting (`TryGet` returns false) or a bad value (`ToBooleany` returns the default).

ldennington (approved):

shango420:
access_token$sandbox$6c5mhjz4hhzks7bk$498becf30ca25c3d655465583bbca1bb

mjcheetham:
@shango420 please do not use this repository to practice using GitHub. Please also be careful what information you are sharing. The last comment you added (that I have now deleted) seemed to contain an sensitive access token.

PR #1447, opened 2023-10-25
platformutils: drop unnecessary netfx #ifdefs

mjcheetham:
Since we're targetting .NET Framework 4.7.2 there are some unnecessary conditional compliation `#if NETFRAMEWORK` blocks. Namely around CLR and runtime/OS detection.


ldennington (approved):
Lovely cleanup!

PR #1464, opened 2023-11-03
Store PasswordExpiry and OAuthRefreshToken

### LINKED ISSUE #268: Add basic validation of stored credentials
Right now GCM Core always returns any stored credentials for a request without any validation.

We cannot ever 100% validate that a credential is "good" for the particular Git command/request because the remote Git server can reject the credentials based on the content of the pack during a push, for example.

We can however validate if a stored a credential is expired or not, such as a PAT or OAuth JWT token that have expiration dates.

To do this we need to extend the credential store API and platform implementations to support reading/writing expiry metadata.

In the event the metadata is missing or the credentials have no expiry we should optimistically return them - assume they are good. This would mean we only reject definitely bad credentials (expired) and avoid false negatives.

### LINKED ISSUE #1463: Store oauth_refresh_token
It would be useful to store Git credential attribute [oauth_refresh_token](https://git-scm.com/docs/git-credential#Documentation/git-credential.txt-codeoauthrefreshtokencode) 

Currently GitLabHostProvider, BitBucketHostProvider and GenericHostProvider use a hacky workaround to save oauth_refresh_token as the password for a separate service:

https://github.com/git-ecosystem/git-credential-manager/blob/2365cac1a6a92106e9a3edef6048ecc788a9521d/src/shared/GitLab/GitLabHostProvider.cs#L218-L219

### LINKED ISSUE #2059: Credential protocol enhancement: OAuth token expiration and refresh tokens
Starting in Git 2.40, the protocol learned a few new keys including around storing and retrieving an OAuth token's expiration date. This will be helpful to support such that tokens stored in credential stores can be used or discarded without an extra network call to determine if they are still within the validity period. 

Likewise, the protocol from v2.41 now includes an optional OAuth refresh token key that allows helpers to more easily communicate both an access and refresh token as part of a `get + store/erase` request and response pair. Today, host providers have had to preemptively store the refresh token manually, in a separate credential store entry, even if the incorrect user account was selected.

Author: `@mjcheetham`

hickford:
Add properties ICredential.PasswordExpiry and ICredential.OAuthRefreshToken. These correspond to Git credential attributes password_expiry_utc and oauth_refresh_token, see https://git-scm.com/docs/git-credential#IOFMT. Previously these attributes were silently disarded.

Plumb these properties from input to host provider to credential store to output.

Credential store support for these attributes is optional, marked by new properties CredentialStore.CanStorePasswordExpiry and ICredentialStore.CanStoreOAuthRefreshToken. Implement support in CredentialCacheStore, SecretServiceCollection and WindowsCredentialManager.

Add method IHostProvider.ValidateCredentialAsync. The default implementation simply checks expiry. Other implementations might query a server.

Improve implementations of GenericHostProvider and GitLabHostProvider. Previously, GetCredentialAsync saved credentials as a side effect. This is no longer necessary. The workaround to store OAuth refresh tokens under a separate service is no longer necessary assuming CredentialStore.CanStoreOAuthRefreshToken. Querying GitLab to check token expiration is no longer necessary assuming CredentialStore.CanStorePasswordExpiry.

Fixes #1463
Fixes #268
Fixes #2059


hickford:
Windows build ought to be fixed by https://github.com/git-ecosystem/git-credential-manager/pull/1418

src/shared/Core/Credential.cs
```diff
@@ -15,21 +18,57 @@ public interface ICredential
         /// Password.
         /// </summary>
         string Password { get; }
+
+        /// <summary>
+        /// The expiry date of the password. This is Git's password_expiry_utc
+        /// attribute. https://git-scm.com/docs/git-credential#Documentation/git-credential.txt-codepasswordexpiryutccode
+        /// </summary>
+        DateTimeOffset? PasswordExpiry { get => null; }
```
ldennington, line 26: This may need to be updated to avoid default interface methods given the current situation with .NET 8 on Windows.

src/shared/Core/Credential.cs
```diff
@@ -15,21 +18,57 @@ public interface ICredential
...
+
+        /// <summary>
+        /// An OAuth refresh token. This is Git's oauth_refresh_token
+        /// attribute. https://git-scm.com/docs/git-credential#Documentation/git-credential.txt-codeoauthrefreshtokencode
+        /// </summary>
+        string OAuthRefreshToken { get => null; }
```
ldennington, line 32: Ditto here 😞.

src/shared/Core/Credential.cs
```diff
@@ -15,21 +18,57 @@ public interface ICredential
...
     }
 
     /// <summary>
     /// Represents a credential (username/password pair) that Git can use to authenticate to a remote repository.
     /// </summary>
-    public class GitCredential : ICredential
+    public record GitCredential : ICredential
```
ldennington, line 38: What was the reason for changing this from `class` to `record`?

src/shared/Core/Credential.cs
```diff
@@ -15,21 +18,57 @@ public interface ICredential
...
+        }
+
+        public string Account { get; init; }
+
+        public string Password { get; init; }
 
-        public string Password { get; }
+        public DateTimeOffset? PasswordExpiry { get; init; }
+        
+        public string OAuthRefreshToken { get; init; }
```
ldennington, line 72: The `init`s may need to be removed given the current situation with .NET 8 on Windows.

src/shared/Core/CredentialCacheStore.cs
```diff
@@ -42,9 +45,10 @@ public IList<string> GetAccounts(string service)
             return Array.Empty<string>();
         }
 
+
         public ICredential Get(string service, string account)
         {
-            var input = MakeGitCredentialsEntry(service, account);
+            var input = MakeGitCredentialsEntry(service, new GitCredential(account, null));
```
ldennington, line 51: Can/should we add a `GitCredential` constructor that just takes an `account`?

src/shared/Core/CredentialStore.cs
```diff
@@ -37,6 +37,7 @@ public ICredential Get(string service, string account)
             return _backingStore.Get(service, account);
         }
 
+
```
ldennington, line 40: Nit: I think we can remove this.

src/shared/Core/CredentialStore.cs
```diff
@@ -66,7 +79,7 @@ private void EnsureBackingStore()
             {
                 case StoreNames.WindowsCredentialManager:
                     ValidateWindowsCredentialManager();
-                    _backingStore = new WindowsCredentialManager(ns);
+                        _backingStore = new WindowsCredentialManager(ns);
```
ldennington, line 82: Nit: I think we can remove this.

src/shared/Core/HostProvider.cs
```diff
@@ -58,6 +58,9 @@ public interface IHostProvider : IDisposable
         /// </summary>
         /// <param name="input">Input arguments of a Git credential query.</param>
         Task EraseCredentialAsync(InputArguments input);
+
+        Task<bool> ValidateCredentialAsync(Uri remoteUri, ICredential credential)
+        => Task.FromResult(credential.PasswordExpiry == null || credential.PasswordExpiry >= DateTimeOffset.UtcNow);
```
ldennington, line 63: Ditto here on removing default interface implementation.

src/shared/Core/ICredentialStore.cs
```diff
@@ -28,14 +29,25 @@ public interface ICredentialStore
         /// <param name="service">Name of the service this credential is for. Use null to match all values.</param>
         /// <param name="account">Account associated with this credential. Use null to match all values.</param>
         /// <param name="secret">Secret value to store.</param>
+//        [Obsolete("Prefer AddOrUpdate(string, ICredential)")]
         void AddOrUpdate(string service, string account, string secret);
 
+        void AddOrUpdate(string service, ICredential credential)
+        => AddOrUpdate(service, credential.Account, credential.Password);
```
ldennington, line 36: Sorry that this issue is touching so many areas 😞. Probably best to switch to the old-school way in this file as well.

src/shared/Core/Interop/Linux/SecretServiceCredential.cs
```diff
@@ -18,6 +18,8 @@ internal SecretServiceCredential(string service, string account, string password
 
         public string Password { get; }
 
-        private string DebuggerDisplay => $"[Service: {Service}, Account: {Account}]";
+        public string OAuthRefreshToken { get; init; }
```
ldennington, line 21: Ditto on this file with removing `init`s.

src/shared/Core/Interop/Windows/WindowsCredentialManager.cs
```diff
@@ -226,7 +245,17 @@ private WindowsCredential CreateCredentialFromStructure(Win32Credential credenti
...
             serviceName = RemoveUriUserInfo(serviceName);
 
-            return new WindowsCredential(serviceName, credential.UserName, password, targetName);
+            // TODO: read password_expiry_utc from attribute
+            // int ptrSize = Marshal.SizeOf<IntPtr>();
+            // for (int i = 0; i < credential.AttributeCount; i++)
+            // {
+                // IntPtr attrPtr = Marshal.ReadIntPtr(credential.Attributes, i * ptrSize);
+                // Win32CredentialAttribute attr = Marshal.PtrToStructure<Win32Credential>(attrPtr);
+            //  }
```
ldennington, line 254: Will want to add this attribute before merging.

src/shared/GitLab/GitLabHostProvider.cs
```diff
@@ -175,50 +175,14 @@ internal AuthenticationModes GetSupportedAuthenticationModes(Uri targetUri)
             return modes;
         }
 
-        // <remarks>Stores OAuth tokens as a side effect</remarks>
-        public override async Task<ICredential> GetCredentialAsync(InputArguments input)
+        public override async Task<bool> ValidateCredentialAsync(Uri remoteUri, ICredential credential)
```
ldennington, line 178: Wow! Lovely cleanup.

src/shared/GitLab/GitLabHostProvider.cs
```diff
@@ -276,18 +225,28 @@ protected override void ReleaseManagedResources()
             base.ReleaseManagedResources();
         }
 
-        private string GetRefreshTokenServiceName(InputArguments input)
+        private string GetRefreshTokenServiceName(Uri remoteUri)
```
ldennington, line 228: Is the `remoteUri` parameter being used here?

src/shared/TestInfrastructure/Objects/TestCredentialStore.cs
```diff
@@ -93,7 +94,7 @@ private IEnumerable<TestCredential> Query(string service, string account)
         }
     }
 
-    public class TestCredential : ICredential
+    public record TestCredential : ICredential
```
ldennington, line 97: Curious again on the change to `record`.

src/shared/Core/Interop/Windows/WindowsCredential.cs
```diff
@@ -1,7 +1,7 @@
 
 namespace GitCredentialManager.Interop.Windows
 {
-    public class WindowsCredential : ICredential
+    public record WindowsCredential : ICredential
```
ldennington, line 4: Curious again on the change to `record`.

ldennington (changes requested):
Overall looks good! Most of my comments are small things - the only major issue that I'd need to see updated before approving is not having expiry implemented for Windows Credential Manager.

LLNobles (commented):
thank you :)

src/shared/Core/Credential.cs
hickford: It aided debugging, but not strictly necessary. 

Reverted.

src/shared/GitLab/GitLabHostProvider.cs
hickford: Good spot. Fixed bug.

src/shared/Core/CredentialCacheStore.cs
hickford: Done.

src/shared/Core/CredentialStore.cs
Mariotilla: $ git branch -m master main
$ git fetch origin
$ git branch -u origin/main main

ldennington (approved):
Thank you for the updates!

src/shared/Core/Interop/Windows/WindowsCredentialManager.cs
```diff
@@ -90,6 +98,25 @@ public void AddOrUpdate(string service, string account, string secret)
...
+                    var attribute = new Win32CredentialAttribute()
+                    {
+                        // consistent with https://github.com/git-for-windows/git/blob/main/contrib/credential/wincred/git-credential-wincred.c
+                        Keyword = "git_password_expiry_utc",
+                        Value = expiryPtr,
+                        ValueSize = expiryBytes.Length,
+                    };
+
+                    newCred.AttributeCount = 1;
+                    newCred.Attributes = Marshal.AllocHGlobal(Marshal.SizeOf(attribute));
```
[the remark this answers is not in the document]
hickford: Added.

src/shared/Core/Interop/Windows/WindowsCredentialManager.cs
```diff
@@ -90,6 +98,25 @@ public void AddOrUpdate(string service, string account, string secret)
...
+                    Marshal.StructureToPtr(attribute, newCred.Attributes, false);
```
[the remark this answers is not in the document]
hickford: Added.

src/shared/Core/HostProvider.cs
```diff
@@ -125,24 +130,50 @@ public virtual async Task<ICredential> GetCredentialAsync(InputArguments input)
...
+                        if (credential.OAuthRefreshToken != null)
+                        {
+                            Context.Trace.WriteLine("Found OAuth refresh token.");
+                            input = new InputArguments(input, credential.OAuthRefreshToken);
+                        }
+                        Context.Trace.WriteLine("Erasing invalid credential...");
+                        // Why necessary to erase? We can't be sure that storing a fresh
+                        // credential will overwrite the invalid credential, particularly
+                        // if the usernames differ.
+                        Context.CredentialStore.Remove(service, credential);
```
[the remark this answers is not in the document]
hickford: Here the whole credential is removed from storage, including any OAuth refresh token.

Typically GenerateCredentialAsync reauthenticates if OAuth refresh fails for any reason.

Is improvement possible in the network failure case? I can't see how.

LLNobles:
Hi, what is going on with this request? The last time I checked my GitHub
account was suspended.

src/shared/Core/HostProvider.cs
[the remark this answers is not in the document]
hickford: src/shared/Core.Tests/HostProviderTests.cs has examples including HostProvider_GetCredentialAsync_InvalidCredentialsStored_ReturnsNewGeneratedCredential

PetSerAl (commented):
My opinion (feal free to disregard it): it is better not to remove refresh token unnecessary. I understand, that GCM can just reauthenticate, but that require extra, otherwise unnecessary, browser interaction from user, thus providing slightly worse user experience.

Anyways, I have no intentions to spend my time to pursue this. Error rate unlikely would be that hi to bother me that much.

hickford:
@mjcheetham any time to look at this?

LLNobles:
Yes

dscho:
> @mjcheetham any time to look at this?

@hickford: @mjcheetham is no longer working at GitHub (but is once again my teammate, this time at Microsoft). Git Credential Manager, however, is still owned by GitHub and it is currently not clear to me who at GitHub takes care of this here repository.

LLNobles:
I am reporting spam.

src/shared/Core/GenericHostProvider.cs
```diff
@@ -139,33 +139,19 @@ private async Task<ICredential> GetOAuthAccessToken(Uri remoteUri, string userNa
...
-                    // Store new refresh token if we have been given one
-                    if (!string.IsNullOrWhiteSpace(refreshResult.RefreshToken))
-                    {
-                        Context.CredentialStore.AddOrUpdate(refreshService, refreshToken.Account, refreshToken.Password);
-                    }
+                    var refreshResult = await client.GetTokenByRefreshTokenAsync(refreshToken, CancellationToken.None);
 
                     // Return the new access token
-                    return new GitCredential(oauthUser,refreshResult.AccessToken);
+                    return new GitCredential(refreshResult, oauthUser);
```
becm, line 154: Delaying the save of a `refresh_token` will likely result in issues.

New value must be saved, even if the generated `auth_token` may be invalid for the requested operation.
This means we'd have to check (and rely on git) to also provide the _new_ `refresh_token` and store its value during an `erase` callback (needs ugly hack in `EraseCredentialAsync`).

Granted the current state (without #1838) is _also_ broken.
But I think this is still the correct location to issue a store.
Otherwise we risk client desync due to missed `refresh_token` value updates.

src/shared/Core.Tests/Commands/GetCommandTests.cs
```diff
@@ -16,14 +17,21 @@ public async Task GetCommand_ExecuteAsync_CallsHostProviderAndWritesCredential()
...
             var stdin = $"protocol=http\nhost=example.com\n\n";
             var expectedStdOutDict = new Dictionary<string, string>
             {
                 ["protocol"] = "http",
                 ["host"]     = "example.com",
                 ["username"] = testUserName,
-                ["password"] = testPassword
+                ["password"] = testPassword,
+                ["password_expiry_utc"] = testExpiry.ToString(),
+                ["oauth_refresh_token"] = testRefreshToken,
```
becm, line 34: The `refresh_token` will likely also have an expiry date.
At least one observed implementation however does not report the constraint.

becm:
Hm, lumping together these 2 distinct entities opens a can of worms when trying to do updates _consistently_.
- update of `refresh_token` can not be missed in _any_ code path
- erase may NOT invalidate the `refresh_token` part (as hinted by @PetSerAl)

So this would mean (to be consistent with current/fallback behavior):
- during initial registration, a partial secret must already be constructed and saved (invalid `access_token`)
- during refresh a partial update of the secret is required (`refresh_token` only)
- the erase must be limited to the `access_token` part
  (if `refresh_token` is the same value, else it came from a successful `git` call that was started later)

When updating the `access_token` part, the value for `refresh_token` MUST be read again,
it may have been updated in the time window between token refresh and token store by a parallel `git` call.

The whole approach of partial updates also massively increases potential for conflicting parallel operations.
Now: Each `get` needs consistent read/store
After: Each store/erase also has the potential to mess something up.
Not sure _how_ often these conflicts appear in regular use, but the devil is in the details…
And each failure will lead to client invalidation in case of single-use `refresh_token` [requirements](https://datatracker.ietf.org/doc/html/rfc9700#section-4.14.2).

becm:
To be honest, in the OAuth case I would opt to
- store the `access_token` during the _get_ phase and
- make _store_ and _erase_ a NOOP.

This is likely to reduce potential conflicts even in the current approach.
But there may be reservations to deviate even more from the _classical_ `git credential` flow.
(Which in my opinion only fits well with a fully interactive flow, but not matching well with OAuth exchanges)

PR #1479, opened 2023-11-24
Use Avalonia 11 generated view codes

de1acr0ix:
Avalonia 11 automatically generates some view codes.

- Properties for controls with `x:Name` or `Name`.
- `InitializeComponent()` method to initialize such properties.
- DevTools for debug builds.

This change uses such generated codes in views and reduces manual efforts. The names of the controls are changed to start with underscore, so that the generated property names align with the naming style in this project.


mjcheetham (approved):
Looks great! Thanks for this quality of life improvement! :-)

Issue #1477, opened 2023-11-21
GCM 2.2.0 and newer prompt for Sign in on every command to GitHub


chrisallennc:
### Version

2.2.0 and newer

### Operating system

Windows

### OS version or distribution

Windows 11

### Git hosting provider(s)

GitHub

### (Azure DevOps only) What format is your remote URL?

None

### Can you access the remote repository directly in the browser?

Yes, I can access the repository

### Expected behavior

Git command executes and user is not prompted to Sign in. Up until GCM 2.1.2 this was the case.

### Actual behavior

Every time a Git command is executed, like "pull", a prompt appears to Sign in. Based on the log it appears to not be finding an existing credential. This happens with GCM 2.2.0 and every version following. 

### Logs

GCM 2.2.0 and newer:
$ git pull
15:07:07.434480 exec-cmd.c:244          trace: resolved executable dir: C:/Users/<userName>/AppData/Local/Programs/Git/mingw64/bin
15:07:07.450204 git.c:463               trace: built-in: git pull
15:07:07.450204 run-command.c:659       trace: run_command: git fetch --update-head-ok
15:07:07.496986 exec-cmd.c:244          trace: resolved executable dir: C:/Users/<userName>/AppData/Local/Programs/Git/mingw64/libexec/git-core
15:07:07.496986 git.c:463               trace: built-in: git fetch --update-head-ok
15:07:07.506999 run-command.c:659       trace: run_command: GIT_DIR=.git git remote-https GitHub https://github.com/<orgName>/<repoName>.git
15:07:07.559999 exec-cmd.c:244          trace: resolved executable dir: C:/Users/<userName>/AppData/Local/Programs/Git/mingw64/libexec/git-core
15:07:07.559999 git.c:749               trace: exec: git-remote-https GitHub https://github.com/<orgName>/<repoName>.git
15:07:07.559999 run-command.c:659       trace: run_command: git-remote-https GitHub https://github.com/<orgName>/<repoName>.git
15:07:07.623023 exec-cmd.c:244          trace: resolved executable dir: C:/Users/<userName>/AppData/Local/Programs/Git/mingw64/libexec/git-core
15:07:07.953546 run-command.c:659       trace: run_command: 'git credential-manager get'
15:07:08.133038 exec-cmd.c:244          trace: resolved executable dir: C:/Users/<userName>/AppData/Local/Programs/Git/mingw64/libexec/git-core
[...]

PR #1525, opened 2024-01-27
Consider domain matched when no domain hint is provided in GitHub account filtering

de1acr0ix:
Several users mentioned that GitHub account filtering is not working. It turns out domain hint is not provided in their user cases. This change considers domain is matched in that case.


de1acr0ix:
This fixes #1477.

PR #1526, opened 2024-01-28
set runtime variable arch dependant

trueToastedCode:
- necessary for arm64 build

### DIFF
--- a/install-dpkg.sh
+++ b/install-dpkg.sh
@@ -1,0 +1,40 @@
+#!/bin/bash
+
+set -e
+
+if dpkg -s "gcm" &> /dev/null; then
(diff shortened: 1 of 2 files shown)


install-dpkg.sh
```diff
@@ -0,0 +1,43 @@
...
+    exit 1
+fi
+
+if [ "$EUID" -ne 0 ]; then
+    echo "This script must be run with sudo."
+    exit 1
+fi
+
+base_path=$(pwd)
+linux_packaging_path="$base_path/src/linux/Packaging.Linux"
```
Presidentialjaygolden35, line 16: Nice

mjcheetham:
ARM64 builds for Linux are now being published.

PR #1535, opened 2024-02-11
Alpine Linux: avoid requiring unsupported OpenSSL version

dscho:
OpenSSL v1.1.* is no longer provided by Alpine Linux v3.19 according to https://pkgs.alpinelinux.org/packages?name=libssl1.1&branch=v3.19&repo=&arch=&maintainer=

However, OpenSSL v3.* is, so let's use that instead.

[EDIT] This fixes [the currently-broken `validate-install-from-source` build](https://github.com/git-ecosystem/git-credential-manager/actions/runs/7862395316/job/21451819126#step:5:46) (see [here](https://github.com/dscho/Git-Credential-Manager-Core/actions/runs/7864697417/job/21456763533) for a successful build of that workflow at the tip of this PR's branch).

### DIFF
--- a/src/linux/Packaging.Linux/install-from-source.sh
+++ b/src/linux/Packaging.Linux/install-from-source.sh
@@ -189,7 +189,7 @@ case "$distribution" in
         $sudo_cmd apk update
 
         # Install dotnet/GCM dependencies.
-        install_packages apk add "curl git icu-libs krb5-libs libgcc libintl libssl1.1 libstdc++ zlib which bash coreutils gcompat"
+        install_packages apk add "curl git icu-libs krb5-libs libgcc libintl libssl3 libstdc++ zlib which bash coreutils gcompat"
 
         ensure_dotnet_installed
     ;;


src/linux/Packaging.Linux/install-from-source.sh
```diff
@@ -189,7 +189,7 @@ case "$distribution" in
         $sudo_cmd apk update
 
         # Install dotnet/GCM dependencies.
-        install_packages apk add "curl git icu-libs krb5-libs libgcc libintl libssl1.1 libstdc++ zlib which bash coreutils gcompat"
+        install_packages apk add "curl git icu-libs krb5-libs libgcc libintl libssl3 libstdc++ zlib which bash coreutils gcompat"
```
mjcheetham, line 192: According to the .NET docs, the dependency for OpenSSL should still be 1.1 on 3.14.x and older Alpine.. we may need to switch on the Alpine version in this script.

https://learn.microsoft.com/en-us/dotnet/core/install/linux-alpine#dependencies

mjcheetham (changes requested):

src/linux/Packaging.Linux/install-from-source.sh
dscho: But it builds ;-)

mjcheetham (changes requested):
I'd like to propose this PR address this issue instead:
https://github.com/git-ecosystem/git-credential-manager/pull/1582

It copes with both 3.14 and 3.15+ Alpine requirements.

mjcheetham:
Addressed in #1583

Issue #1554, opened 2024-03-19
"git clone" yields error "fatal: --local can only be used inside a git repository"


jazzdelightsme:
Using git.exe 2.44.0.1, a very simple `git clone` call:
```
git clone --quiet --depth 1 $Source $cloneDest
```
yielded an unexpected error:
```
fatal: --local can only be used inside a git repository
```
The clone apparently succeeded (we got the files on disk), but the error looks very concerning ("fatal"!).

Downgrading to git 2.43 makes the error go away.

Derrick Stolee helped us isolate this to GCM--attached is a trace (using `GIT_TRACE2_PERF=1`) that he captured for us. He hypothesized:

> It's possible that there was a change in GCM but also it could be a change in upstream Git that changed the timing of these GCM calls that think they can update the credentials.

> Also, the --quiet, --depth 1, and $cloneDest parts of your command are not necessary to reproduce this problem.

[trace-results.txt](https://github.com/git-ecosystem/git-credential-manager/files/14655861/trace-results.txt)

jazzdelightsme:
Curiously, I was not able to repro (`git version 2.44.0.windows.1`)... but apparently Derrick was. Perhaps I already had credentials cached?

adriancontrerasdsm:
Honestly when I pushed back it  it  came off  when I was looking for it the credits should be there

Josan66:
Hey I don't need more securities against my self I need access to my
accounts lock everyone else out of my shit please and thank u

Alexandreslv24:
git clone --quiet --depth 1 $Source $cloneDest

fatal: --local can only be used inside a git repository

mjcheetham:
This error should be fixed in https://github.com/git-ecosystem/git-credential-manager/pull/1561

The next release of GCM should include this fix.

PR #1561, opened 2024-03-26
Do not query local bindings if outside repo in Azure Repos

mjcheetham:
Do not attempt to query for an organisation/user binding in the local repository configuration if we're not currently inside of a repository!

This will prevent the warning message from Git's standard error stream from flowing to our caller's stderr stream when GCM tries to use the `--local` option in a `git config` call.

cc @markphip


vdye (approved):

Luis19877:
> Do not attempt to query for an organisation/user binding in the local repository configuration if we're not currently inside of a repository!
> 
> This will prevent the warning message from Git's standard error stream from flowing to our caller's stderr stream when GCM tries to use the `--local` option in a `git config` call.
> 
> cc @markphip
> 
> Fixes #1554

PR #1573, opened 2024-04-08
Update the projects to .NET 8, drop support for .Net Framework, clean up unreachable code.

rakitaj:
My attempt at dropping support for .Net Framework and supporting .NET 8.

I did these changes mechanically as possible, and would love discussing if there is a desire for this? I'm happy to make any/all changes the team wants.

My goal here is using .NET 8 to support all OSes and making it easy to use the dotnet tool version of this repo.

1. Removed `net472` from all `<TagetFramework>` and `<TargetFrameworks>` elements.
2. Removed all conditional MSBuild for .Net Framework.
3. Moved from `<TagetFrameworks>` and `<TargetFramework>` because this only targets only framework now.
4. Remove all references to `net472` build artifacts.
5. Remove and (hopefully) improve the code involved with conditional compilation. I got rid of most of the conditional mechanically - that is I just removed the stuff inside `if NETFRAMEWORK`. In a few cases I tried to chase down the resulting unreachable code this caused and remove it from the project  as well.

The two spots I am the most unsure of my changes are removing support for the OsBroker, which looked like it only worked under .Net Framework and Windows, and removing the interactive auth window.

I did not clean up the Auth options enum because the values had assigned numbers and I didn't want to affect anything which relies on the already assigned numbers.


mjcheetham:
This PR is a duplicate of the effort in https://github.com/git-ecosystem/git-credential-manager/pull/1418

We are unable to fully drop .NET Framework at this time due to a courtesy to Visual Studio who bundles the same GCM across all their supported versions, and they need older VSs and GCMs to work on Windows 8.1 until 2029(!) 😢

rakitaj:
Thanks for pointing that out, I didn't see there is a PR already open. I haven't made many open source contributions before, is closing my PR the right move? (I'm happy to do that)

Wow 2029! I'll poke around the issues and see if there is anything else I can help with. 

What do you think about upgrading the project to .NET 8 from 7 while keeping everything else the same? I can put together a proof of concept PR for that.

.github/workflows/continuous-integration.yml
```diff
@@ -37,9 +37,9 @@ jobs:
       shell: bash
       run: |
```
shopdrop1, line 38: Geephillips

rakitaj:
Hey @mjcheetham I'm closing this PR. I _think_ that's the right move in this case, if you think otherwise just let me know?

PR #1574, opened 2024-04-08
install-from-source: update libssl for alpine

ldennington:
The `validate-install-from-source` workflow is currently failing [1]. According
to [2] and [3], the solution is to upgrade the libssl version from 1.1 to 3, as
version 1.1 is out of support. Since this version is only supported on Alpine
3.15 and later, we detect the version of Alpine the user is running on and set
this dependency version accordingly.
    
1: https://github.com/git-ecosystem/git-credential-manager/actions/runs/8604126129
2: https://stackoverflow.com/questions/77640298/dockerfile-unable-to-select-packages-linux-alpine
3: https://learn.microsoft.com/en-us/dotnet/core/install/linux-alpine

### DIFF
--- a/src/linux/Packaging.Linux/install-from-source.sh
+++ b/src/linux/Packaging.Linux/install-from-source.sh
@@ -186,10 +186,16 @@ case "$distribution" in
         ensure_dotnet_installed
     ;;
     alpine)
+        libssl_version=libssl3
+
+        if [[ $(echo "$version 3.15" | awk '{print ($1 < $2)}') -ne 0 ]]; then
+            libssl_version=libssl1.1
+        fi
+
         $sudo_cmd apk update
 
         # Install dotnet/GCM dependencies.
-        install_packages apk add "curl git icu-libs krb5-libs libgcc libintl libssl1.1 libstdc++ zlib which bash coreutils gcompat"
+        install_packages apk add "curl git icu-libs krb5-libs libgcc libintl $libssl_version libstdc++ zlib which bash coreutils gcompat"
 
         ensure_dotnet_installed
     ;;


mjcheetham:
This is the same PR as what @dscho did in https://github.com/git-ecosystem/git-credential-manager/pull/1535

The same issue occurs however in that we need to still use libssl1.1 for older versions of Alpine, so we need to update the script to switch on the Apline version.

ldennington:
> The same issue occurs however in that we need to still use libssl1.1 for older versions of Alpine, so we need to update the script to switch on the Alpine version.

Not quite sure if I picked up what you're putting down, but I added a check to make sure the user is running on 3.15 or greater. Will that suffice?

mjcheetham (changes requested):
I'd like to propose this PR address this issue instead:
https://github.com/git-ecosystem/git-credential-manager/pull/1582

It copes with both 3.14 and 3.15+ Alpine requirements.

ldennington:
> Not quite sure if I picked up what you're putting down, but I added a check to make sure the user is running on 3.15 or greater. Will that suffice?

Actually, thinking about this a little more, it seemed like the best idea would be to set the version of `libssl` we use based on the version of Alpine the user is running. I've made updates accordingly and tested both scenarios (>3.14 and 3.14 and below) in my fork.

ldennington:
> I'd like to propose this PR address this issue instead:
> 
> https://github.com/git-ecosystem/git-credential-manager/pull/1582
> 
> 
> 
> It copes with both 3.14 and 3.15+ Alpine requirements.

Sorry! I didn't see this before submitting my updates. I'll take a look at your PR.

ldennington:
Closed in favor of #1582

PR #1579, opened 2024-04-15
Update Mac and Linux targets to .NET 8 LTS

mjcheetham:
Update the target framework version for Mac and Linux to .NET 8, which is the latest LTS. .NET 7 is [leaving support on 2024-05-14](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core) so we need to do this.

Windows is not being updated at this time due our courtesy to Visual Studio (that bundle the same GCM version across all their versions) who need Windows 8.x support until.. 2029... 😢
and Windows 10 is required for .NET (Core) 8.

I plan to revisit the .NET Framework to Core on Windows effort after a discussion with the VS team at Microsoft.


vdye (approved):

dscho:
> Windows is not being updated at this time due our courtesy to Visual Studio (that bundle the same GCM version across all their versions) who need Windows 8.x support until.. 2029... 😢

Could we conceivably start a long-running `windows-lts-2.4` branch, and release Windows-only releases targeting .NET Framework, then backport all the PRs to that branch that would still be applicable? These would essentially be releases specifically for our Visual Studio friends. I know, this would be a bit more maintenance burden, just when funding for that kind of stuff was pulled... But I still think it'd be the right thing to do.

PR #1582, opened 2024-04-16
install-from-source.sh: fix libssl Alpine pkgs

mjcheetham:
The version of libssl available in the Alpine package feeds changed with version 3.15. The latest libssl package is 1.1 in 3.14.x and earlier, but newer versions only have libssl3.

Add a check for the Alpine version in the install from source script and switch between installing libssl3 and libssl1.1 depending on the distro version.

Also add another entry to the test matrix of distributions to include a 3.14.x Alpine version.

This replaces the two other PRs that aimed to address this issue:
- https://github.com/git-ecosystem/git-credential-manager/pull/1535
- https://github.com/git-ecosystem/git-credential-manager/pull/1574

Tested the updated workflow here: https://github.com/git-ecosystem/git-credential-manager/actions/runs/8698286553/job/23855027830

### DIFF
--- a/.github/workflows/validate-install-from-source.yml
+++ b/.github/workflows/validate-install-from-source.yml
@@ -24,6 +24,7 @@ jobs:
         - image: tgagor/centos-stream
         - image: redhat/ubi8
         - image: alpine
+        - image: alpine:3.14.10
         - image: opensuse/leap
         - image: opensuse/tumbleweed
         - image: registry.suse.com/suse/sle15:15.4.27.11.31
--- a/src/linux/Packaging.Linux/install-from-source.sh
+++ b/src/linux/Packaging.Linux/install-from-source.sh
@@ -137,6 +137,10 @@ print_unsupported_distro() {
     echo "See https://gh.io/gcm/linux for details."
 }
 
+version_at_least() {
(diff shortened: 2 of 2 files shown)


.github/workflows/validate-install-from-source.yml
```diff
@@ -24,6 +24,7 @@ jobs:
         - image: tgagor/centos-stream
         - image: redhat/ubi8
         - image: alpine
+        - image: alpine:3.14.10
```
ldennington, line 27: Not sure how I feel about this - I'd prefer to continue using the latest version in this workflow but could be persuaded otherwise.

dscho: I share that sentiment. If there is a reasonable way to keep targeting the latest version (which many of our users will do, I expect), I'd love to see that.

dscho (approved):
Thank you!

.github/workflows/validate-install-from-source.yml
mjcheetham: We are targeting both!

```
- image: alpine
- image: alpine:3.14.10
```

The first line is the latest, the second is a specific tag/version.

ldennington: Sorry I missed that! Was reviewing from my phone 😂.

ldennington (approved):

PR #1583, opened 2024-04-16
azrepos: fix error message to stderr when not in repo

mjcheetham:
Suppress the standard error stream from Git when checking if we're inside a repository or not, because the failure message is otherwise printed to the user's console when we are not (which is what we're trying to decide!).

Note that we normally don't suppress or redirect the standard error stream for Git commands as the user may have enabled Git tracing (v1 or v2) to the stderr file and we don't want to mute these for all calls.

Example before:

```shell
% pwd
/Users/mjcheetham
% gcm azure-repos list
fatal: not a git repository (or any of the parent directories): .git
devdiv:
  (global) -> USER@DOMAIN
```

Example now:
```shell
% pwd
/Users/mjcheetham
% gcm azure-repos list
devdiv:
  (global) -> USER@DOMAIN
```

### DIFF
--- a/src/shared/Core/Git.cs
+++ b/src/shared/Core/Git.cs
@@ -21,6 +21,12 @@ namespace GitCredentialManager
         ChildProcess CreateProcess(string args);
 
         /// <summary>
+        /// Returns true if the current Git instance is scoped to a local repository.
+        /// </summary>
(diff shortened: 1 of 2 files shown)


dscho (approved):
❤️

Issue #1460, opened 2023-11-03
Bitbucket DC reports `Must scpecify at least on AuthenticationModes` and fallsback to cli input


bmarroquin:
### Version

2.3.2

### Operating system

Windows

### OS version or distribution

Windows 10 x64

### Git hosting provider(s)

Bitbucket Server/DC

### (Azure DevOps only) What format is your remote URL?

None

### Can you access the remote repository directly in the browser?

Yes, I can access the repository

### Expected behavior

I expected the UI to pop up and prompt me for my username + token

### Actual behavior

A fatal message is displayed `Must specify at least one AuthenticationModes` and then a CLI prompt asks for username and password.

### Logs

The issue is that `BitbucketHelper.GetBaseUri` doesn't consider that Bitbucket DC could be proxied behind a path. For example, `https://<host>/bitbucket`. The method drops the needed path and causes 404 errors in the following API requests.
Then `DataCenter.BitbucketRestApi.GetAuthenticationMethodsAsync` and `DataCenter.BitbucketRestApi.IsOAthInstalledAsync` swallow the 404 errors and 0 Authentication modes are returned.

A quick fix would be to have `DataCenter.BitbucketRestApi.GetAuthenticationMethodsAsync` raise an exception on a 404 and the fallback logic in 
`BitbucketHostProvider.GetSupportedAuthenticationsModesAsync` should kick in and i would expect everything to work.

mjcheetham:
Thanks for raising this issue!

As a workaround until this is fixed, could you try manually setting the authentication modes in Git config?

```shell
git config --global credential.https://<host>.bitbucketAuthModes "oauth,basic"
```

where `<host>` is the hostname of the DC instance.

https://gh.io/gcm/config#credentialbitbucketauthmodes

bmarroquin:
Yeah, the workaround works. Thanks.

bmarroquin:
I finally got time to dig into this more. Setting the authModes works but it prompts for credentials every time. This happens because it tries to validate the credentials against the path less URL. See log below.
```
17:37:44.356528 ...tHostProvider.cs:113 trace: [GetStoredCredentials] Found stored credentials: <user>/********
17:37:44.451164 ...tHostProvider.cs:393 trace: [ValidateCredentialsWork] Validate credentials (<user>/********) are fresh for https://<host>/ ...
[...]

PR #1584, opened 2024-04-16
adds httpPath variable for Bitbucket DC

bmarroquin:
A workaround for #1460. Allows setting a path for Bitbucket DC so Rest and OAuth endpoints can be relative to the host.

Adds a benefit that Bitbucket DC OAuth can be fully configured via registry settings. Previously we needed useHttpPath which, as far as i know, cannot be set via registry.

### DIFF
--- a/src/shared/Atlassian.Bitbucket.Tests/BitbucketHelperTest.cs
+++ b/src/shared/Atlassian.Bitbucket.Tests/BitbucketHelperTest.cs
@@ -1,7 +1,6 @@
 using Atlassian.Bitbucket.DataCenter;
 using GitCredentialManager;
 using Moq;
-using Newtonsoft.Json.Linq;
 using System;
 using Xunit;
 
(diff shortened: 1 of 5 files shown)


Aamesay:
> A workaround for #1460. Allows setting a path for Bitbucket DC so Rest and OAuth endpoints can be relative to the host.
> 
> Adds a benefit that Bitbucket DC OAuth can be fully configured via registry settings. Previously we needed useHttpPath which, as far as i know, cannot be set via registry.

[release-homebrew](https://github.com/git-ecosystem/git-credential-manager/actions/workflows/release-c

Aamesay (commented):
Ok

Zerocoolooog (commented):
- - [ ] - - - [ ] # @~~@ ~~@hickford ****[]()

Zerocoolooog (commented):
zerocoolooog

PR #1590, opened 2024-04-17
diagnose: avoid poluting stderr stream in Git diagnostic

mjcheetham:
Avoid poluting the standard error stream with a 'fatal' Git error message during the Git diagnostic run as part of the `diagnose` command.

When checking if we are inside of a Git repo we should be using an explicit check for `IGit::IsInsideRepository` rather than trying to get the current repo path and checking for null.


vdye (approved):

jekkyll76:
Update

PR #1642, opened 2024-06-23
Update install.md with respect to latest ubuntu that uses Gdebi

genie360s:
Latest  Ubuntu 24.04 uses Gdebi to install Debian packages

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -75,7 +75,7 @@ installation method.
 Download the latest [.deb package][latest-release]*, and run the following:
 
 ```shell
-sudo dpkg -i <path-to-package>
+sudo gdebi <path-to-package>
 git-credential-manager configure
 ```
 


docs/install.md
```diff
@@ -75,7 +75,7 @@ installation method.
 Download the latest [.deb package][latest-release]*, and run the following:
 
 ```shell
-sudo dpkg -i <path-to-package>
+sudo gdebi <path-to-package>
```
dscho, line 78: > Latest Ubuntu 24.04 uses Gdebi to install Debian packages

Hrmmmm. I tried to reconstruct the (missing) reference to the official Ubuntu announcement to that end, and failed. It does not look like `gdebi` is _the_ Ubuntu 24.04 way to install `.deb` packages. In fact, all information I could find on the matter suggested that `dpkg` is still installed by default, and `gdebi` is not.

In any case, there's more versions of Ubuntu, and there's quite a few Debian package-based Linux distributions that happen not to be Ubuntu. Given this lack of ubiquity, I would have expected _maybe_ to augment the existing command with a suggestion to use `gdebi` instead, on Ubuntu 24.04, but certainly not to replace the advice to use `dpkg`.

So: hrmmmm.

genie360s:
Oohkay , makes sense , definitely can act as a supplement

dav3yblaz3 (commented):
lets get this going team thank you

PR #1644, opened 2024-06-26
Update code signing with latest Azure Trusted Signing tools/tasks

mjcheetham:
- Update to the latest version of the GitHub Action (the `azure/azure-code-signing-action` action has been replaced by the `azure/trusted-signing-action` one).

- Deploy a forked version of the `Sign.Cli` tool for Trusted Signing, which includes the ability to export the certificate. The fork can be found here https://github.com/mjcheetham/sign/tree/export-opt, and the PR to submit this change upstream here https://github.com/dotnet/sign/pull/734.

With these changes we are now completely secret/credential free, and rely on federation only.


dscho (approved):
Nice!

PR #1647, opened 2024-07-02
Use published sign.exe CLI and 3rd part certificate extractor tools

mjcheetham:
Use a 3rd party tool to extract the NuGet signing certificate for upload rather than relying on an option added to the sign.exe tool in a private fork. At the same time let's use the `dotnet tool install` command to acquire the code signing tool, rather than rely on our Azure blob store.

Also let's drop the `-d` and `-u` options from the sign.exe CLI; they are no longer required ([source](https://github.com/dotnet/sign/pull/734#issuecomment-2196589982)).


dscho (approved):
Looks good to me!

PR #1666, opened 2024-07-18
Add support for sending X5C when using a service principal with certificate for authentication

spongemike2:
When using a service principal with certificate authentication, every time the certificate is renewed, the new certificate needs to be uploaded to the service principal's AAD app registration in order for authentication to continue to work.

However, a technology called "X5C" has made this unnecessary by allowing _any_ certificate, with a _specific_ subject, issued by a known, trusted, predetermined CA, to be used.

For this to work, the AAD app registration's manifest needs to be updated to reflect the subject name, and during authentication, the request for "X5C" authentication needs to be sent along with the certificate's signature.

This change enables that to take place.


spongyryno (commented):
Should version be removed?

src/shared/Microsoft.AzureRepos/AzureReposHostProvider.cs
```diff
@@ -549,6 +549,14 @@ private bool UseServicePrincipal(out ServicePrincipalIdentity sp)
 
             if (hasCertThumbprint)
             {
+                bool hasX5CSetting = _context.Settings.TryGetSetting(
+                    AzureDevOpsConstants.EnvironmentVariables.ServicePrincipalCertificateSendX5C,
+                    Constants.GitConfiguration.Credential.SectionName,
+                    AzureDevOpsConstants.GitConfiguration.Credential.ServicePrincipalCertificateSendX5C,
+                    out string certHasX5C);
+
+                sp.SendX5C = !hasX5CSetting || certHasX5C != "false";
```
mjcheetham, line 558: For this we have some helpers for parsing Git config 'bool' strings to `bool` - values like `"1"` or `"on"` should also be treated as `true`.

Taking the raw string value from settings, you can call `ToBooleanyOrDefault` which will do this conversion for you.. use `ToBooleanyOrDefault(false)` to set the default, if unset or invalid, to `false` to like below:

```suggestion
                sp.SendX5C = _context.Settings.TryGetSetting(
                    AzureDevOpsConstants.EnvironmentVariables.ServicePrincipalCertificateSendX5C,
                    Constants.GitConfiguration.Credential.SectionName,
                    AzureDevOpsConstants.GitConfiguration.Credential.ServicePrincipalCertificateSendX5C,
                    out string certHasX5CStr) && certHasX5CStr.ToBooleanyOrDefault(false);
```

mjcheetham (changes requested):
Thank you for this addition! Just a few things please:

- address the comment about parsing the config values
- drop the VERSION file change
- please can you also add docs for the `GCM_AZREPOS_SP_CERT_SEND_X5C` and `credential.azreposServicePrincipalCertificateSendX5C` settings to the corresponding markdown docs in `/docs`!

😄

VERSION
```diff
@@ -1 +1 @@
-2.5.1.0
```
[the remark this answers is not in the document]
spongyryno: Fix forthcoming...

copiden (commented):
KIN7PMLPEVOBUTSBS6OMMW6YV673EDJS

Findrown:
Used and suggested it to you for above reasons.

src/shared/Microsoft.AzureRepos/AzureReposHostProvider.cs
kamsatantalani19, line 1: brew install --cask git-credential-manager

PR #1688, opened 2024-08-13
Update install.md

tungstengmd:
emphasises existence of linuxbrew


dscho:
> emphasises existence of linuxbrew

Does GCM have support in Linuxbrew? I think not. And if that is correct, then this PR would add misleading information.

mjcheetham:
We do not publish Linux compatible packages on Home/Linuxbrew.

PR #1694, opened 2024-08-22
Remove misleading error message

Regenhardt:
GitLab is used for internal hosting, which may often use unencrpted http. This error message is misleading, as the process continues and works regardless, and also falsely mentions Github instead of GitLab.

Got a new PC at work. First clone of the GitLab repo I'm working on went like this:

 - run git clone on an http GitLab instance, in my case I ran `git clone http://gitlab.company-local-domain.de/group/project LocalRepo-Name`
   Note:
    - http, not https
    - GitLab
 - Get the following error message:
   fatal: Unencrypted HTTP is not supported for GitHub. Ensure the repository remote URL is using HTTPS.
 - Get prompted for and enter credentials
 - Get warning about redirecting to .git URL
 - Cloning commences and finishes successfully

### DIFF
--- a/src/shared/GitLab/GitLabHostProvider.cs
+++ b/src/shared/GitLab/GitLabHostProvider.cs
@@ -41,9 +41,6 @@ namespace GitLab
                 return false;
             }
 
-            // We do not support unencrypted HTTP communications to GitLab,
-            // but we report `true` here for HTTP so that we can show a helpful
-            // error message for the user in `CreateCredentialAsync`.
             if (!StringComparer.OrdinalIgnoreCase.Equals(input.Protocol, "http") &&
                 !StringComparer.OrdinalIgnoreCase.Equals(input.Protocol, "https"))
             {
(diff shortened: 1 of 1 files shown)


goxijana3:
> GitLab is used for internal hosting, which may often use unencrpted http. This error message is misleading, as the process continues and works regardless, and also falsely mentions Github instead of GitLab.
> 
> Got a new PC at work. First clone of the GitLab repo I'm working on went like this:
> 
>     * run git clone on an http GitLab instance, in my case I ran `git clone http://gitlab.company-local-domain.de/group/project LocalRepo-Name`
>       Note:
>       
>       * http, not https
>       * GitLab
> 
>     * Get the following error message:
>       fatal: Unencrypted HTTP is not supported for GitHub. Ensure the repository remote URL is using HTTPS.
> 
>     * Get prompted for and enter credentials
> 
>     * Get warning about redirecting to .git URL
> 
>     * Cloning commences and finishes successfully

Regenhardt:feature/gitlab-gets-github-error

mjcheetham:
Thank you for your PR! We have addressed this issue of not being able to use `http://` remotes with GitLab with this other PR https://github.com/git-ecosystem/git-credential-manager/pull/1721

Regenhardt:
That just changes the error message at default settings, but doesn't change the behaviour, does it? The message is then still misleading, as it implies that you have to change a setting and try again, but then still proceeds with the authentication?

PR #1711, opened 2024-09-30
Fix the build in Visual Studio

dscho:
Michael J. Lyons reached out privately and reported a problem with building inside Visual Studio. The symptom is:

| Error (active) | MSB4044 | The "GenerateWindowsAppManifest" task was not given a value for the required parameter "Version". |
| - | - | - |

This PR fixes this error, and then also addresses all the warnings pointed out by Visual Studio.


mjcheetham (approved):
Thanks @dscho !

PR #1715, opened 2024-10-01
(spam)

IAN0073:


Scutua:
why i didn't see eny coin in my wallet 😒 do you kniw that? i miss
something?

thanks
tua

Scutua:
hi

PR #1721, opened 2024-10-07
Add control over use of unsafe remotes

mjcheetham:
Today, all the custom host providers (Azure Repos, Bitbucket, GitHub, GitLab) block the use of HTTP (unencrypted) remote URLs and error out. Only the generic host provider permits HTTP remotes.

From https://github.com/git-ecosystem/git-credential-manager/pull/1694, we learn that a common use case for self/corporate hosted Git servers is to use HTTP remotes. Even if this is **_not recommended_**, GCM should not outright block these.

Instead, we now add an option, `GCM_ALLOW_UNSAFE_REMOTES` or `credential.allowUnsafeRemotes`, for the user to explicitly set to allow the use of these unsafe remotes.

For the generic host provider we only print a warning when using HTTP remotes to reduce the churn for existing users who rely on GCM for HTTP remotes.

### DIFF
--- a/docs/environment.md
+++ b/docs/environment.md
@@ -302,6 +302,32 @@ Defaults to false (use hardware acceleration where available).
 
 ---
 
+### GCM_ALLOW_UNSAFE_REMOTES
+
+Allow transmitting credentials to unsafe remote URLs such as unencrypted HTTP
+URLs. This setting is not recommended for general use and should only be used
+when necessary.
+
+Defaults false (disallow unsafe remote URLs).
+
+#### Example
+
+##### Windows
+
+```batch
+SET GCM_ALLOW_UNSAFE_REMOTES=true
+```
+
+##### macOS/Linux
+
+```bash
+export GCM_ALLOW_UNSAFE_REMOTES=true
+```
+
+**Also see: [credential.allowUnsafeRemotes][credential-allowunsaferemotes]**
+
+---
+
 ### GCM_AUTODETECT_TIMEOUT
 
 Set the maximum length of time, in milliseconds, that GCM should wait for a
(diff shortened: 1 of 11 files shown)


src/shared/Core/Constants.cs
```diff
@@ -226,6 +228,7 @@ public static class HelpUrls
             public const string GcmAutoDetect          = "https://aka.ms/gcm/autodetect";
             public const string GcmDefaultAccount      = "https://aka.ms/gcm/defaultaccount";
             public const string GcmMultipleUsers       = "https://aka.ms/gcm/multipleusers";
+            public const string GcmUnsafeRemotes       = "https://aka.ms/gcm/unsaferemotes";
```
mjcheetham, line 231: This shortlink has been configured to point to

```
https://github.com/git-ecosystem/git-credential-manager/blob/release/docs/netconfig.md#unsafe-remote-urls
```

dscho (approved):
This PR looks good to me!

There is [a new error about a broken link](https://github.com/git-ecosystem/git-credential-manager/actions/runs/11215496399/job/31172801736#step:3:41):

> ✗ [404] https://specifications.freedesktop.org/secret-service/ | Failed: Network error: Not Found

This link is in GCM's documentation apparently 3 times. I think that this should be updated to https://specifications.freedesktop.org/secret-service-spec/.

dscho:
> There is [a new error about a broken link](https://github.com/git-ecosystem/git-credential-manager/actions/runs/11215496399/job/31172801736#step:3:41):

I have opened https://github.com/git-ecosystem/git-credential-manager/pull/1722 to fix this; [This successful doc linting run](https://github.com/git-ecosystem/git-credential-manager/actions/runs/11215944101/job/31174163849?pr=1722) suggests that I've managed 😊

docs/environment.md
katnessgumyblack-kawatan: #2259

PR #1345, opened 2023-07-24
docs: update required dotnet-sdk version in install.md

kawana77b:
Update the version notation because since the release of git-credential-manager version 2.2.1, [the sdk version required for installation is .NET 7](https://www.nuget.org/packages/git-credential-manager/2.2.1).

PR #1744, opened 2024-10-18
Update install.md to reference NET 8

derekantrican:
I just tried to run the `dotnet tool install -g git-credential-manager` command and it failed because I did not have .NET 8  installed. So updating the docs.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -210,7 +210,7 @@ the preferred install method for Linux because you can use it to install on any
 distribution][dotnet-supported-distributions]. You
 can also use this method on macOS if you so choose.
 
-**Note:** Make sure you have installed [version 7.0 of the .NET
+**Note:** Make sure you have installed [version 8.0 of the .NET
 SDK][dotnet-install] before attempting to run the following `dotnet tool`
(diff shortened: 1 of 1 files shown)


dscho (commented):
Does this leave some ancient CentOS systems behind?

derekantrican:
> Does this leave some ancient CentOS systems behind?

I have no idea. I just know that I tried to follow the install instructions for .NET 7 and it said I needed .NET 8. So I'm updating the instructions

dscho:
> > Does this leave some ancient CentOS systems behind?
> 
> I have no idea. I just know that I tried to follow the install instructions for .NET 7 and it said I needed .NET 8. So I'm updating the instructions

It would be good if you could do the research lest you break _other_ people's scenarios in an attempt to fix your own.

derekantrican:
This is only a documentation change, this will not break anything. .NET 8 is required as of #1579 - _that_ change is what will actually "break other people's scenarios".

I see a similar change was made to only this file in #1345

mjcheetham (approved):

Lauramansanarez42 (commented):
- [ ]

PR #1747, opened 2024-10-22
ci: add Mariner and Arch Linux

dscho:
After dropping the now-obsolete `tgagor/centos-stream` value from the matrix in #1746, let's now instead add the "somewhat" supported Mariner Linux and Arch Linux to the testing matrix.


dscho:
For the record, here is a manually-triggered, successful run of the `validate-install-from-source` workflow that uses this PR branch: https://github.com/dscho/git-credential-manager/actions/runs/11458170852.

mjcheetham (approved):

PR #1757, opened 2024-11-06
don't require runtime to be set to install from source

theofficialgman:
@mjcheetham this should fix this https://github.com/git-ecosystem/git-credential-manager/pull/1633#issuecomment-2459973610

it also fixes as a side effect the alpine linux build from source being built against the wrong [glibc linux (linux-x64)](https://github.com/git-ecosystem/git-credential-manager/actions/runs/11573830139/job/32216761503#step:5:142) runtime instead of the correct [musl linux (linux-musl-x64)](https://github.com/theofficialgman/git-credential-manager/actions/runs/11706600286/job/32604070640#step:5:142) runtime

full test run: https://github.com/theofficialgman/git-credential-manager/actions/runs/11706600286

### DIFF
--- a/src/linux/Packaging.Linux/build.sh
+++ b/src/linux/Packaging.Linux/build.sh
@@ -49,27 +49,9 @@ if [ ! -d "$INSTALL_PREFIX" ]; then
     mkdir -p "$INSTALL_PREFIX"
 fi
 
-# Fall back to host architecture if no explicit runtime is given.
-if test -z "$RUNTIME"; then
-    HOST_ARCH="`dpkg-architecture -q DEB_HOST_ARCH`"
-
-    case $HOST_ARCH in
-        amd64)
-            RUNTIME="linux-x64"
-            ;;
-        arm64)
-            RUNTIME="linux-arm64"
-            ;;
-        armhf)
-            RUNTIME="linux-arm"
-            ;;
-        *)
-            die "Could not determine host architecture!"
-            ;;
-    esac
+if [ ! -z "$RUNTIME" ]; then
+    echo "Building for runtime ${RUNTIME}"
 fi
-
-echo "Building for runtime ${RUNTIME}"
 
 # Perform pre-execution checks
 CONFIGURATION="${CONFIGURATION:=Debug}"
--- a/src/linux/Packaging.Linux/layout.sh
+++ b/src/linux/Packaging.Linux/layout.sh
@@ -44,10 +44,6 @@ PROJ_OUT="$OUT/linux/Packaging.Linux"
 # Build parameters
 FRAMEWORK=net8.0
 
-if [ -z "$RUNTIME" ]; then
-    die "--runtime was not set"
-fi
-
 # Perform pre-execution checks
 CONFIGURATION="${CONFIGURATION:=Debug}"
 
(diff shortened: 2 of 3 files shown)


src/linux/Packaging.Linux/build.sh
```diff
@@ -49,28 +49,10 @@ if [ ! -d "$INSTALL_PREFIX" ]; then
     mkdir -p "$INSTALL_PREFIX"
 fi
 
-# Fall back to host architecture if no explicit runtime is given.
-if test -z "$RUNTIME"; then
-    HOST_ARCH="`dpkg-architecture -q DEB_HOST_ARCH`"
```
mjcheetham, line 54: How about we just move away from using the `dpkg-architecture` command all together? That way we don't need to rely on something that isn't always installed.

`uname -m` should give us what we want.

```shell
if test -z "$RUNTIME"; then
    HOST_ARCH="$(uname -m)"

    case "$HOST_ARCH" in
        x86_64)
        amd64)
            RUNTIME="linux-x64"
            ;;
        arm64)
        aarch64)
            RUNTIME="linux-arm64"
            ;;
        arm)
        armhf)
        aarch32)
            RUNTIME="linux-arm"
            ;;
        *)
            die "Unknown or unsupported host architecture: '$(HOST_ARCH)'"
            ;;
    esac
fi
```

src/linux/Packaging.Linux/layout.sh
```diff
@@ -76,13 +72,22 @@ fi
...
+        --output="$(make_absolute "$PAYLOAD")" || exit 1
+else
+    $DOTNET_ROOT/dotnet publish "$GCM_SRC" \
+        --configuration="$CONFIGURATION" \
+        --framework="$FRAMEWORK" \
+        --runtime="$RUNTIME" \
+        --self-contained \
+        -p:PublishSingleFile=true \
+        --output="$(make_absolute "$PAYLOAD")" || exit 1
+fi
```
mjcheetham, line 90: I think we should keep layout.sh always producing a runtime-specific output.

src/linux/Packaging.Linux/pack.sh
```diff
@@ -100,6 +100,26 @@ INSTALL_TO="$DEBROOT/usr/local/share/gcm-core/"
...
+            ;;
+        armhf)
+            RUNTIME="linux-arm"
+            ;;
+        *)
+            die "Could not determine host architecture!"
+            ;;
+    esac
+fi
+
```
mjcheetham, line 122: If we move back to requiring `$RUNTIME` and use the `uname -m` command instead in `build.sh` we won't need this.

mjcheetham (changes requested):
I think we should move away from using `dpkg-architecture` altogether if we can. How about `uname -m` instead in build.sh, and revert the other files?

src/linux/Packaging.Linux/build.sh
theofficialgman: nope. that is unfortunately a very common misconception. uname gives the kernel architecture. it knows nothing of the system OS architecture. these commonly do not match as many 32bit distros use 64bit kernels.

src/linux/Packaging.Linux/layout.sh
theofficialgman: it does produce a runtime specific output. if runtime is not passed dotnet publish automatically determines the appropriate runtime from the running system. Refer to the linked actions in the first message showing that.

src/linux/Packaging.Linux/pack.sh
theofficialgman: Refer to https://github.com/git-ecosystem/git-credential-manager/pull/1757#discussion_r1831288265

src/linux/Packaging.Linux/layout.sh
theofficialgman: To add to that, it would be very difficult to correctly determine each appropriate runtime to use ourselves for every type of system combination when there are dozens. Letting dotnet handle that (because it knows what runtime it was built on itself) is the correct way https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.NETCore.Platforms/src/PortableRuntimeIdentifierGraph.json#L146-L304

src/linux/Packaging.Linux/build.sh
theofficialgman: Not to mention we also cannot detect what libc (glibc/musl) the running OS uses based on the kernel. Using uname is totally flawed.

PR #1811, opened 2025-01-24
Add support for macOS enterprise deployable default settings

mjcheetham:
Similar to the functionality to set default settings for GCM via the Registry on Windows, we implement support on macOS using the Apple preferences system. Preferences can be deployed as payloads via MDM configuration profiles.

The domain for GCM is `git-credential-manager` and configuration settings should take the form of a dictionary under the `configuration` key.

PR #1814, opened 2025-01-27
Add support for Linux enterprise deployable default settings

mjcheetham:
Mirroring #1811, add support to set default settings for Linux, via MDM tooling. For this platform we look for files in the `/etc/git-credential-manager/config.d` directory.

Files are a simple format of `key=value\n`, where the `key` is a configuration value name from `docs/config.md`. This format prevents values from containing a `\n` line feed character, but this is unlikely to be an issue in practice.


dscho (approved):
Looks good to me!

I am stumped by the compile errors on Windows, hopefully you can figure them out!

mjcheetham:
> I am stumped by the compile errors on Windows, hopefully you can figure them out!

It's because we target .NET Framework on Windows, and some of the overloads are not available, and I wrote these patches on Mac and Linux. Will check my syntax for .NET FX compat.

PR #1816, opened 2025-01-29
uninstall.md update

briant112708:
Disconnect -device system 
Disconnect -log devices
Disconnect -server log


briant112708:
C

```kotlin

```

PR #1838, opened 2025-02-15
fix(generic): save new refresh_token value

becm:
Missing the update of single-use-tokens will lead to continuous re-registration requirement for GCM.

Check in statement is done correctly for new value.
The update request just used the wrong variable.

### DIFF
--- a/src/shared/Core/GenericHostProvider.cs
+++ b/src/shared/Core/GenericHostProvider.cs
@@ -161,7 +161,7 @@ namespace GitCredentialManager
                     // Store new refresh token if we have been given one
                     if (!string.IsNullOrWhiteSpace(refreshResult.RefreshToken))
                     {
-                        Context.CredentialStore.AddOrUpdate(refreshService, refreshToken.Account, refreshToken.Password);
+                        Context.CredentialStore.AddOrUpdate(refreshService, refreshToken.Account, refreshResult.RefreshToken);
                     }
 
                     // Return the new access token


hickford (approved):

dscho:
Thank you for bearing with us for such a long time in limbo. It's not quite over yet; @mjcheetham is busy right now with the huge task to re-establish a working release process (currently all signing/notarizing is missing all prerequisite secrets, and as a consequence no binaries would be signed, which would break Git Credential Manager on macOS). So please bear with us a bit longer.

Just a quick note why this has not yet been merged: While on the face of it, it should be a no-brainer to store any new refresh token (at least it was me, until I asked Matthew about it and he patiently explained to me that I need to consider more scenarios than the obvious one), a new refresh token can be provided in circumstances where we would need to ask the user for confirmation (for example if the account name was changed, in which case the account _name_ also would have to be updated). Another plausible situation where this would happen is when there is a malicious actor who uses the same account elsewhere, in which case Git Credential Manager should probably ring the alarm bells.

Also, there are now finally credential protocol enhancements in Git itself that would allow refresh tokens to be processed in a stateful manner, with enough context to avoid, say, storing them for the wrong account. For more details see https://github.com/git-ecosystem/git-credential-manager/issues/2059.

In short: More things need to be considered before this Pull Request can go forward. And the release process needs to be fixed before it would be well-spent time to enhance GCM... I therefore have to ask for even more patience...

becm:
This update happens _only_ in the context of an already completely resolved account.
The degrees of freedom at this point are limited to bound token values and their expiration info.

At this point we:
- have resolved a certain fixed account scope
- successfully used the previous `refresh_token` (so _the account_ was correctly identified) and
- the server told us explicitly to use a different value next time we come back _in this context_.

**Assuming** we are dealing with a server that is already set up according to current [recommendations](https://datatracker.ietf.org/doc/html/rfc6749#section-10.4)
→ single use `refresh_token`:
- failing to update the token value at this stage **will** trigger auth issues on the next request
- using outdated credentials should also trigger invalidation of the new value.

The current behavior is unlikely to have an effect on most aspects of MITM attacks.
It even prevents use of _proper_ server settings to have at least _a chance_ to detect token leaks.

This update is actually the **only** operation that MUST be performed (immediately) at this point.
We MAY NOT wait until `git store/erase` comes back with a new `refresh_token` at some later time.
In case of a crash or when getting interrupted it may _never come back **at all**_.

If the superseded `request_token` is used during a parallel `git credential` operation, this would _already_ trigger credential invalidation of the **new** in-flight value:
- correctly updating the `refresh_token` value is MANDATORY, any reuse of old value breaks client registration
- the time between API reply and value update MUST be kept minimal to avoid conflicting use.

_Also; the comment above it says the code should be this way!_ ☺

dscho:
I have to admit that I lack understanding about the logic. For one, I do not even understand how the current code works, and when I look at [the `OAuth2TokenResult` class](https://github.com/git-ecosystem/git-credential-manager/blob/e0e310edf74dee872849abe2c893a8df76916a83/src/shared/Core/Authentication/OAuth/OAuth2TokenResult.cs#L5-L22) I see that there is more information that should be stored. But I do not have the big picture (is it really appropriate to store the refresh token if it turns out that the password does not work, e.g. due to missing scopes?), therefore I'll defer to @mjcheetham (who is still quite occupied with repairing processes that fell into disarray).

becm:
Only the `AccessToken` member of the `OAuth2TokenResult` is handled (correctly) at the moment.

This PR is to address a bug in processing the `RefreshToken` value
(as required by spec and in accordance with @mjcheetham's comment in the [original code](/git-ecosystem/git-credential-manager/commit/e323c8345762e0082ad2c40d9f5c54b6872050e2#diff-0284d23631d6a317ab2b79005b1583dd2592850fd7d5aca3b4623cb606ced238R148-R152)).
The `refresh_token` value (if supplied) MUST be stored, the old value may NEVER be reused.

Even a later `git credential erase` should NOT touch a stored `refresh_token` → different auth level
(will become a headache if stored together, since this _somehow_ MUST clear the `access_token`).

The members `ExpiresIn` and `TokenType` (both metadata for `AccessToken`) are currently discarded.

Use and persistent storage of `TokenType` will be required for effective support of #2058.

The `ExpiresIn` _can_ be transformed to an absolute value and passed to `git` in the _initial reply_.
Not sure it makes sense in a later `git credential get` or better to directly trigger refresh on expiry.

dscho:
@becm I hear you, but I do have to ask you for patience because I cannot make the call on my own whether to just take this bug fix or whether to focus on #2059 _instead_.

In any case, the primary attention right now must lie with fixing the broken release process, anything else will have to wait for that fix.

mr-otmarg:
Similar topic #2003

mjcheetham (approved):
After some thought, I think it'll be fine to take this as is. It's clearly a bug in the current setup that we're not actually writing the new refresh token. As to what about if the account name changes? Well, it'll be safer for now to keep the existing account IMO.

mjcheetham:
Closing and re-opening to force a rerun on the check builds.

PR #1839, opened 2025-02-15
point out arm location in docs

paulgmiller:


ridgunn:
These changes are unrelated to the GCM repository. The line you're modifying refers specifically to the Git for Windows installation, not to GCM.

PR #1905, opened 2025-04-26
FIX: Install from source script error caused by shell expecting value

AguirreMoy:
Existing behavior:

CleanShot 2025-04-25 at 16 35 53@2x

Correct behavior
CleanShot 2025-04-25 at 17 36 40@2x


ridgunn:
Fixed by #2167. Thanks for contributing the issue!

PR #1909, opened 2025-04-30
Fix a few issues with TRACE2 output

mjcheetham:
I noticed a few problems with the TRACE2 support when writing to a pipe/socket. The handling of the `trace2.eventTarget` setting to get the correct pipe name/socket file path was incorrect resulting in names like "pipe\foobar" when given "//./pipe/foobar" on Windows, when it should be "foobar".

Also the event field name 'format' is incorrect; it should be 'fmt'.


dscho (approved):
Looks good to me!

ghost:
Fixed issues.

PR #1986, opened 2025-07-09
Fixed a tiny grammar mistake in readme.md

Moomoobeef:
# About
Under the "Supported Git Versions" section there was
> However there are some know problematic releases of Git that are not compatible.

Which is not quite right, "know" should instead be "known". So I fixed it. This is an incredibly minor thing, I just figured I would fix this since I noticed it.


dscho:
Thanks! Fixed via https://github.com/git-ecosystem/git-credential-manager/pull/1884.

PR #2042, opened 2025-09-17
Create a shell script for fetching and installing the 'latest' or any specific release tag for GCM.

eschwieb:
Large mono-repos that want to have tooling for installing a particular version of GCM, but who don't want to use homebrew, can leverage this script to do the selection and download. Right now this is Mac-specific (for the Microsoft Office Monorepo Mac setup systems).


castellanosrober103-ctrl:
> Large mono-repos that want to have tooling for installing a particular version of GCM, but who don't want to use homebrew, can leverage this script to do the selection and download. Right now this is Mac-specific (for the Microsoft Office Monorepo Mac setup systems).

castellanosrober103-ctrl:
Se ve j

PR #2087, opened 2025-10-24
merge

Chriswsi:


dscho:
@Chriswsi prerequisite question: are you prepared to _actually_ work on this? I ask because it does not seem as if you had pre-reviewed this PR before opening it, which is a big no-no.

Issue #2028, opened 2025-09-02
.deb package signed with a deprecated SHA1 algorithm


junoslukan:
When following the [guide](https://github.com/git-ecosystem/git-credential-manager/blob/release/docs/linux-validate-gpg.md#debian-package) to verify the signature for the .deb package, I got the following error:

```
$ debsig-verify --debug gcm.deb

debsig: Starting verification for: gcm.deb
debsig:         getSigKeyID: got 189ABF20BC4D22098078A6403C853823978B07FA for origin key
debsig: getDbPathname: using /etc/debsig/policies/3C853823978B07FA keyring
debsig: Using policy directory: /etc/debsig/policies/3C853823978B07FA
debsig:   Parsing policy file: /etc/debsig/policies/3C853823978B07FA/generic.pol
debsig:     parsePolicyFile: parsing '/etc/debsig/policies/3C853823978B07FA/generic.pol'
debsig:     parsePolicyFile: completed
debsig:     Checking Selection group(s).
debsig:       Processing 'origin' key...
debsig: getDbPathname: using /usr/share/debsig/keyrings/3C853823978B07FA/gcm-public.gpg keyring
debsig:         getKeyID: mapped 3C853823978B07FA -> 189ABF20BC4D22098078A6403C853823978B07FA
debsig:         getSigKeyID: got 189ABF20BC4D22098078A6403C853823978B07FA for origin key
debsig:     Selection group(s) passed, policy is usable.
debsig: Using policy file: /etc/debsig/policies/3C853823978B07FA/generic.pol
debsig:     Checking Verification group(s).
debsig:       Processing 'origin' key...
debsig: getDbPathname: using /usr/share/debsig/keyrings/3C853823978B07FA/gcm-public.gpg keyring
debsig:         getKeyID: mapped 3C853823978B07FA -> 189ABF20BC4D22098078A6403C853823978B07FA
debsig:         getSigKeyID: got 189ABF20BC4D22098078A6403C853823978B07FA for origin key
gpg: Signature made Wed 30 Oct 2024 11:59:51 CET
gpg:                using RSA key 189ABF20BC4D22098078A6403C853823978B07FA
gpg: Note: signatures using the SHA1 algorithm are rejected
gpg: Can't check signature: Invalid digest algorithm
debsig: sigVerify: gpg exited abnormally or with non-zero exit status
debsig: verifyGroupRules: failed for origin
debsig:     Verification group failed checks.
[...]

PR #2165, opened 2025-11-11
release: sign .deb files with a stronger algorithm

dscho:
It has been reported in https://github.com/git-ecosystem/git-credential-manager/issues/2028 that Git Credential Manager's `.deb` files cannot be validated with `debsig-verify`. A deeper investigation brought to light that the reason is a weak digest algorithm: `debsigs` calls `gpg --opengpg` which uses SHA-1, but `debsig-verify` rejects such weak signatures.

This was actually already fixed in `debsigs`, but current Ubuntu versions will not get those fixes, being stuck with older `debsigs` versions.

It's easy enough to work around, though, so just like https://github.com/microsoft/git/pull/815, let's let Git Credential Manager's `release` workflow work around that.

This closes https://github.com/git-ecosystem/git-credential-manager/issues/2028

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -270,6 +270,15 @@ jobs:
         # Install debsigs
         sudo apt install debsigs
 
+        # Stop using SHA-1 for the signature. For details, see
+        # https://gitlab.com/debsigs/debsigs/-/commit/75c6c8f96e6cdc33bca9c5f32195b68ff35bc32f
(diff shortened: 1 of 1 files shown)


mjcheetham:
No longer required as this workflow has been replaced by #2176, and ESRP uses SHA-256 for GPG signing.

dscho:
> No longer required as this workflow has been replaced by #2176, and ESRP uses SHA-256 for GPG signing.

Makes sense! Let's keep #2028 in mind, though, when we make the first ESRP-based release just so that we can ping them that it's fixed.

PR #2176, opened 2025-11-13
(Re-)introduce release builds with Azure Pipelines

mjcheetham:
Since stewardship of the project has transitioned back to Microsoft from GitHub, we're moving the official release builds from GitHub Actions to Azure Pipelines. This is mainly a requirement to enable Microsoft code-signing of executables, packages, etc.

The release pipeline will also publish a GitHub release and the .NET Tool packages to NuGet.org.

The GitHub Actions release workflow is no longer in use (it also required access to approve releases from internal GitHub code-signing systems, which we no longer have access to).

### DIFF
--- a/.azure-pipelines/release.yml
+++ b/.azure-pipelines/release.yml
@@ -12,15 +12,15 @@ resources:
 parameters:
   - name: 'esrp'
     type: boolean
-    default: false
+    default: true
     displayName: 'Enable ESRP code signing'
   - name: 'github'
     type: boolean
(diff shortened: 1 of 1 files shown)


mjcheetham:
For Microsoft employees, here's a link to the test release build showcasing this pipeline:
https://dev.azure.com/mseng/1ES/_build/results?buildId=30708065

ridgunn (approved):
Looks good to me

PR #2217, opened 2025-12-18
Detect Linux runtime from host architecture if not specified

mjcheetham:
Detect the runtime (linux-x64/linux-arm64/etc) from the current host architecture if not specified via the `--runtime` argument.

We had already been doing this in the script, but after we already `die`-d when `--runtime` was missing! Oops.

Let's move this auto-detection logic up-front, and also no longer rely on `dpkg-architecture` to determine the host arch; we can use `uname -m` instead.


dscho (approved):
Looks good to me!

PR #2230, opened 2025-12-31
Add support for Windows x64 and arm64 builds

dennisameling:
Successor of https://github.com/git-ecosystem/git-credential-manager/pull/1846

Currently, GCM is only offered as an x86 build for Windows. On arm64, this results in poor performance because the code has to be fully emulated.

Luckily, `dotnet` fully supports `win-arm64` nowadays. Even the .NET Framework [supports](https://devblogs.microsoft.com/dotnet/announcing-dotnet-framework-481/) it! [Here's some more details](https://github.com/dotnet/core/issues/7709).

This PR follows the same logic as https://github.com/git-ecosystem/git-credential-manager/pull/1633, which added support for Linux arm64. The main `git-credential-manager.exe` executable now becomes native, while the DLLs remain `Any CPU`. This allows for .NET Framework to run natively on arm64. I also added `win-x64` while at it.

Confirmed to work correctly both on x64 and arm64 for GitHub operations.

### x64

Schermafbeelding 2025-02-27 154529

### arm64

image

Also confirmed to correctly generate the installers for Windows arm64. The system-wide installer also correctly installs into the 64-bit `C:\Program Files\Git Credential Manager` folder:

image

### DIFF
--- a/.azure-pipelines/release.yml
+++ b/.azure-pipelines/release.yml
@@ -46,8 +46,8 @@ parameters:
       - id: windows_arm64
         jobName: 'Windows (ARM64)'
         runtime: win-arm64
-        pool: GitClientPME-1ESHostedPool-intel-pc
-        image: win-x86_64-ado1es
+        pool: GitClientPME-1ESHostedPool-arm64-pc
+        image: win-arm64-ado1es
         os: windows
 
   - name: macos_matrix
@@ -207,7 +207,7 @@ extends:
                         --no-dependencies `
                         -p:NoLayout=true `
                         -p:PayloadPath="$(Build.ArtifactStagingDirectory)\payload" `
-                        -p:OutputPath="$(Build.ArtifactStagingDirectory)\installers"
+                        -p:OutputPath="$(Build.ArtifactStagingDirectory)\installers" `
                         -p:RuntimeIdentifier="${{ dim.runtime }}"
                 - task: EsrpCodeSigning@5
                   condition: and(succeeded(), eq('${{ parameters.esrp }}', true))


src/windows/Installer.Windows/Installer.Windows.csproj
```diff
@@ -1,12 +1,12 @@
...
+<Project>
   <!-- Implicit SDK props import -->
   <Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk" />
 
   <PropertyGroup>
     <TargetFramework>net472</TargetFramework>
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
     <EnableDefaultItems>false</EnableDefaultItems>
-    <PayloadPath>$(PlatformOutPath)Installer.Windows\bin\$(Configuration)\net472\win-x86</PayloadPath>
+    <PayloadPath>$(PlatformOutPath)Installer.Windows\bin\$(Configuration)\net472\$(RuntimeIdentifier)</PayloadPath>
```
mjcheetham: The `RuntimeIdentifier` is not set when building without a specific RID/Architecture parameter, such as when building the solution from inside of an IDE, or with `dotnet build -c WindowsDebug`.

>  Installer.Windows net472 failed with 1 error(s) (1.4s)
>    >D:\src\gcm\src\windows\Installer.Windows\Installer.Windows.csproj(43,5): error Layout script failed with exit code 1 and message Output: D:\src\gcm\out\windows\Installer.Windows\bin\Debug\net472\;Unsupported RuntimeIdentifier:

The other 'Installer/Packacing' projects for macOS/Linux don't do much except shell out to a shell script that dynamically computes the current host's RID when it's not specified.

We can probably do some MSBuild fun with properties and discovering the current host's arch if RID has not been explicitly specified.

OR.. mimic the macOS/Linux implementations and create some layout.ps1 and pack.ps1 scripts; not relying on project-project dependencies to build the PayloadPath contents.

stephnewme100-cloud:
Lol I'm the owner

mjcheetham (approved):

PR #2232, opened 2026-01-06
Add linux-arm64 builds to official releases

mjcheetham:
The build process already supports ARM64 for Linux, so let's extend the official build process to also produce these binaries.

Note that we have already been performing [CI builds on GitHub of `linux-arm64`](https://github.com/git-ecosystem/git-credential-manager/blob/1b3f77efe12efffaadac7c5a218822896e657a39/.github/workflows/continuous-integration.yml#L59), so this is just about extending the official build pipeline.


ridgunn (approved):
I verified that the changes work successfully with arm64 pipeline.

Issue #1813, opened 2025-01-27
Bug: GCM Misidentifies Docker Containers Running on WSL2 Due to /proc/version


vinitsiriya:
**Description** Git Credential Manager (GCM) incorrectly identifies Docker containers running on WSL2 as WSL distributions. This is caused by `/proc/version` in the container inheriting WSL-related strings (`WSL2`, `microsoft`) from the underlying WSL2 host kernel. As a result, GCM incorrectly applies WSL-specific logic and attempts to invoke Windows interop tools (e.g., `powershell.exe`), which are not available in Linux-native Docker containers.This results in **failed Git operations**  inside the Docker container, even when the container is a full Linux environment.

---

**Steps to Reproduce** 
#### Prerequisites: 

1. Docker Desktop is installed on Windows and configured to use WSL2 as the backend.

2. GCM for Linux is installed inside the Docker container.

#### Steps: 
 
1. **Start a Full Desktop Environment Docker Container** :
Run a desktop-enabled container like `linuxserver/rdesktop`:

```bash
docker run -it linuxserver/rdesktop
```
 
2. **Check Kernel Version Inside the Container** :
Inside the container, run:

```bash
cat /proc/version
```
**Output** :

```less
Linux version 5.15.167.4-microsoft-standard-WSL2 (root@f9c826d3017f) (gcc (GCC) 11.2.0, GNU ld (GNU Binutils) 2.37) #1 SMP Tue Nov 5 00:21:55 UTC 2024
```

This indicates the container inherits the WSL2 kernel from the host.
 
3. **Install Git and GCM** :
Inside the container:

```bash
apt update && apt install -y git curl
curl -sSL https://github.com/GitCredentialManager/git-credential-manager/releases/download/v2.0.935/gcm-linux-x86_64.2.0.935.tar.gz | tar -xz
sudo ./install.sh
git config --global credential.helper /usr/local/bin/git-credential-manager
```
 
4. **Attempt to Clone a Repository** :
Run:

```bash
git clone https://<username>@dev.azure.com/<project>/<repo>
```
**Observe the Error** :

```vbnet
fatal: An error occurred trying to start process '/mnt/c/Windows/System32/WindowsPowerShell/v1.0/powershell.exe' with working directory '/root/'. No such file or directory
```

---

**Current Behavior**  
[...]

PR #2254, opened 2026-01-29
Check for xdg-open availability on Linux for browser auth

kyle-rader-msft:
## Summary

Check for browser launcher availability (xdg-open, gnome-open, etc.) in addition to checking for desktop session presence to gate web auth flow.

## Problem

GCM currently only checks for `DISPLAY` or `WAYLAND_DISPLAY` environment variables to determine if browser-based authentication is available on Linux. However, MSAL checks for `xdg-open` availability in addition to display environment variables. This creates a mismatch where GCM reports "browser not available" even though MSAL can successfully launch a browser using `xdg-open`.

## Real-World BLocker: WaveSpace VMs

This fix isimportant for **WaveSpace VMs** and similar environments where:
- Users connect via SSH without X11 forwarding (no `DISPLAY` environment variable)
- `xdg-open` is available and can launch the host system's browser (this is setup by VSCode remote tunnels)
- Browser-based authentication needs to work for conditional access policies, satisifed by the host's browser (Microsoft Edge in this case).

## Solution

Modified `LinuxSessionManager.GetWebBrowserAvailable()` to return `true` if:
- An interactive desktop session is detected (existing behavior via `DISPLAY`/`WAYLAND_DISPLAY`), **OR**
- A shell execute handler like `xdg-open` is available (new behavior)

This matches [MSAL's browser detection logic](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/8faea30d3e43ea88c10603312f65ac0633907775/src/client/Microsoft.Identity.Client/Platforms/netstandard/NetCorePlatformProxy.cs#L202-L210) and ensures GCM, AzureAuth and MSAL behave consistently.

## Behavior Changes

| Scenario | Before | After |
|----------|--------|-------|
| Linux with DISPLAY set | ✅ Browser available | ✅ Browser available |
| Linux with WAYLAND_DISPLAY set | ✅ Browser available | ✅ Browser available |
| Linux with xdg-open but no DISPLAY | ❌ Browser NOT available | ✅ Browser available (FIXED) |
| Linux without display vars or browser launchers | ❌ Browser NOT available | ❌ Browser NOT available |

## Testing

- ✅ Solution builds successfully with no warnings
- Manual testing recommended on Linux system without DISPLAY but with xdg-open available
- Verify OAuth authentication works in SSH sessions with xdg-open present

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/shared/Core/Interop/Linux/LinuxSessionManager.cs
+++ b/src/shared/Core/Interop/Linux/LinuxSessionManager.cs
@@ -21,53 +21,34 @@ public class LinuxSessionManager : PosixSessionManager
     
     private bool GetWebBrowserAvailable()
     {
-        // Check if we have a way to launch the user's browser via a shell execute handler
-        // (xdg-open, gnome-open, kfmclient, wslview)
+        // Determine system properties
         bool hasHandler = BrowserUtils.TryGetLinuxShellExecuteHandler(Environment, out _);
+        bool isWsl = WslUtils.IsWslDistribution(Environment, FileSystem, out _);
+        bool isWslSession0 = isWsl && WslUtils.GetWindowsSessionId(FileSystem) == 0;
+        bool isDesktopSession = IsDesktopSession;
 
-        if (hasHandler)
-        {
-            // If this is a Windows Subsystem for Linux distribution, we need
-            // additional checks even though we have a handler (e.g., wslview)
-            if (WslUtils.IsWslDistribution(Environment, FileSystem, out _))
-            {
-                //
-                // If we are in Windows logon session 0 then the user can never interact,
-                // even in the WinSta0 window station. This is typical when SSH-ing into a
-                // Windows 10+ machine using the default OpenSSH Server configuration,
-                // which runs in the 'services' session 0.
-                //
-                // If we're in any other session, and in the WinSta0 window station then
-                // the user can possibly interact. However, since it's hard to determine
-                // the window station from PowerShell cmdlets (we'd need to write P/Invoke
-                // code and that's just messy and too many levels of indirection quite
-                // frankly!) we just assume any non session 0 is interactive.
-                //
-                // This assumption doesn't hold true if the user has changed the user that
-                // the OpenSSH Server service runs as (not a built-in NT service) *AND*
-                // they've SSH-ed into the Windows host (and then started a WSL shell).
-                // This feels like a very small subset of users...
-                //
-                if (WslUtils.GetWindowsSessionId(FileSystem) == 0)
-                {
-                    return false;
-                }
-            }
+        //
+        // WSL session 0 note:
+        // If we are in Windows logon session 0 then the user can never interact,
+        // even in the WinSta0 window station. This is typical when SSH-ing into a
+        // Windows 10+ machine using the default OpenSSH Server configuration,
+        // which runs in the 'services' session 0.
+        //
+        // If we're in any other session, and in the WinSta0 window station then
+        // the user can possibly interact. However, since it's hard to determine
+        // the window station from PowerShell cmdlets (we'd need to write P/Invoke
+        // code and that's just messy and too many levels of indirection quite
+        // frankly!) we just assume any non session 0 is interactive.
+        //
+        // This assumption doesn't hold true if the user has changed the user that
+        // the OpenSSH Server service runs as (not a built-in NT service) *AND*
+        // they've SSH-ed into the Windows host (and then started a WSL shell).
+        // This feels like a very small subset of users...
+        //
 
-            // We have a handler and either not WSL or not in session 0
-            return true;
-        }
-
-        // No explicit handler found
-        // If this is WSL without a handler, we cannot launch the browser
-        if (WslUtils.IsWslDistribution(Environment, FileSystem, out _))
-        {
-            return false;
-        }
-
-        // For non-WSL Linux without an explicit handler, fall back to checking for
-        // a desktop session (X11/Wayland). Desktop environments typically have handlers
-        // installed, so this maintains backward compatibility.
-        return IsDesktopSession;
+        // We can launch a browser if:
+        // 1. We have a shell execute handler (xdg-open, wslview, etc.) and not blocked by WSL session 0, OR
+        // 2. We have a desktop session on non-WSL Linux (backward compatibility - assumes handler exists)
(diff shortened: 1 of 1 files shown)


becm:
@mjcheetham return of `GetWebBrowserAvailable()` may **not** block on missing Linux utilities.
~~Current WSL detection redirects requests to the host system _just fine_ without them.~~
Or maybe _not yet_, [BrowserUtils.OpenDefaultBrowser()](https://github.com/git-ecosystem/git-credential-manager/blob/d31f62ce60c3c473b639b90f3bf3f8c35aad33ab/src/shared/Core/BrowserUtils.cs#L29) seems to not handle WSL in a special way…

The presence of `wslview` is an esoteric corner case and likely breaks similar to our existing logic
(e.g.: WSL interop is disabled → #1813).

The problem here is HOW to detect if any `open` tool is properly set up to redirect to a UI-capable system?

becm:
Next debatable design; none of the listed _open_ tools in [BrowserUtils.cs](https://github.com/git-ecosystem/git-credential-manager/blob/d31f62ce60c3c473b639b90f3bf3f8c35aad33ab/src/shared/Core/BrowserUtils.cs#L75) guarantee there IS a usable browser installed (same issue in the linked MSAL proxy lookup).

So in a proper desktop session, valid programs would rather be
`["x-www-browser", "www-browser", <other known UI HTTPS applications>]`.

If possible, detection of WSL-interop or _other redirect condition_ is independent from availability of _local_ browsers.

kyle-rader-msft:
The primary issue we are hitting is that out the box - GCM does not work anymore on remote WaveSpace VMs for Msft devs using Linux VMs through VSCode remote tunnel. GCM falls back to device code, and the device code flow no longer passes conditional access and we are totally blocked. 

The VSCode remote tunnel sets up a `$BROWSER` and `xdg-open`. The `$BROWSER` script triggers launching through the host and your host web browser on the Intuned machine passes CA and can log in just fine.

AzureAuth does this, and so does the Azure CLI.

becm:
@kyle-rader-msft based on your description, the presence of a `BROWSER` environment variable would be the highest ranked indicator for using _exactly this_ (non-standard) tool.
Even most UI sessions no longer set this variable (or override).

Based on that assumption my personal preference would be to _just_ add code path for using this variable at the beginning of `GetWebBrowserAvailable()` and `OpenDefaultBrowser()`.
It's a **much stronger** indicator the user wants to use a specific HTTP handler, than any of the other _magic_ we can do to deduce a working setup.

But @mjcheetham  (or other maintainers) will have to decide if this may be a more robust approach.
Or if there are some bad drawbacks.

becm:
The code for [xdg-open](https://gitlab.freedesktop.org/xdg/xdg-utils/-/blob/master/scripts/xdg-open.in) has nice examples for:
- multi value split of `$BROWSER` using `:` as delimiter
- non-standard invocation using `%s` to support non-standard calls apart from `<executable> <URL>`

There is also a good default selection of _actual_ web browsers for UI and command line.

kyle-rader-msft:
So @becm @mjcheetham - are you saying simply detch if `$BRWOSER` is set, (don't parse it) just it's prescense we can allow web? That would be my preference. 

Another option would be to stop tryign to detect this, because MSAL already has this detection and it's working in this environment. in AzureAuth this is not aan isuse, because we simply attempt the web auth flow, and if it's not supported, MSAL throws the exception and we move on to device code if it's in the allowed modes.

mjcheetham:
> [..] the presence of a BROWSER environment variable would be the highest ranked indicator for using exactly this (non-standard) tool.
Even most UI sessions no longer set this variable (or override).
>
> Based on that assumption my personal preference would be to just add code path for using this variable at the beginning of GetWebBrowserAvailable() and OpenDefaultBrowser().
It's a much stronger indicator the user wants to use a specific HTTP handler, than any of the other magic we can do to deduce a working setup.

I would think an up-front explicit check for `$BROWSER` in `IsWebBrowserAvailable`, and then _using that variable_ in `BrowserUtils.OpenDefaultBrowser` (when `IsLinux() == true` in both cases) would be a good thing to add here.

@kyle-rader-msft would this also satisfy the VSCode remote tunnels scenario since they set `$BROWSER`? 

HOWEVER.. updating the `LinuxSessionManager.IsWebBrowserAvailable` implementation will not be sufficient to allow browser-based MSAL interactive auth to proceed.

In `MicrosoftAuthentication.CanUseSystemWebView` we check `ISessionManager.IsWebBrowserAvailable` _and_ MSAL's `PCA.IsSystemWebViewAvailable`:

https://github.com/git-ecosystem/git-credential-manager/blob/ece43792a8335393a3611d1e9ad529edd7b7501c/src/shared/Core/Authentication/MicrosoftAuthentication.cs#L860-L864

[MSAL's implementation on Linux](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/efed09b70652f28e12d51f5f0976674aaa2228c0/src/client/Microsoft.Identity.Client/Platforms/netstandard/NetCoreWebUIFactory.cs#L17) just checks for [the presence of `$DISPLAY`](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/efed09b70652f28e12d51f5f0976674aaa2228c0/src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/DesktopOsHelper.cs#L199-L202) which will not be set via VSCode Remote Tunnels.

> Another option would be to stop tryign to detect this, because MSAL already has this detection and it's working in this environment. 

I wonder if this means these other tools are _not_ calling `PCA.IsSystemWebViewAvailable`?

Either way, we'd also want to drop the `PCA.IsSystemWebViewAvailable` check in `MicrosoftAuthentication.CanUseSystemWebView`, else you'd also still need to set `$DISPLAY` for the VSCode scenario you mention.

kyle-rader-msft:
Hmm, yeah in azureauth we do not ask the PCA if system browser is available we just try it, and it works. Since [MSAL does look for xdg-open on Linux](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/8faea30d3e43ea88c10603312f65ac0633907775/src/client/Microsoft.Identity.Client/Platforms/netstandard/NetCorePlatformProxy.cs#L253) and in the remote VSCode tunnel we have xdg-open, and xdg-open parses `$BROWSER`.

mjcheetham:
Closing as the other PRs have now been merged that address this scenario! Thanks for pointing out the issue and providing the framework to address this!

PR #2256, opened 2026-02-03
Guard against failure to launch PowerShell from WSL

mjcheetham:
Let's not crash the process if we are unable to launch PowerShell to determine the Windows session ID.


mjcheetham:
cc @becm as I know you've been looking at the WSL interop scenarios recently.

PR #2257, opened 2026-02-03
Make browser detection more robust on Linux

mjcheetham:
In non-WSL environments we make our checks for a browser more robust by checking for a 'shell execute' handler (like xdg-open).

Previously we just relied on a desktop session (`DISPLAY` or `WAYLAND_DISPLAY`), which isn't accurate since in `OpenBrowser` we require a shell execute handler. The check and implementation are now aligned!

In a Visual Studio Code remote session, we may be able to launch a browser without the `DISPLAY` environment variable set. This is because VSCode sets the `BROWSER` environment variable in remote sessions, such that it can forward requests to start a browser to the client machine.

However! There are several types of remote session in VSCode, and the way they handle automatic port forwarding differs slightly. Since a very common case in GCM for launching the user browser is the ability to connect back to GCM via `localhost:<port>`, we only consider a subset of remote sessions to be able to launch a browser:

* Remote SSH     ✅ 
image

* Dev Containers ✅ 
image

* Remote Tunnel  ❌ 
image
    
Sadly, for whatever reason, the remote connection over Microsoft Dev Tunnels does NOT automatically surface a forwarded port to localhost on the client. The forwarded port is only available via the devtunnels.ms URLs.

Dev Tunnels|SSH
-|-
image|image

Detecting the different types of remote session is tricky as it's not always as explicit as we'd like.
    
All types: `VSCODE_IPC_HOOK_CLI` is set.

Dev Containers : `REMOTE_CONTAINERS` is set.
Remote SSH     : `SSH_CONNECTION` is set.
Remote Tunnel  : absense of `REMOTE_CONTAINERS` and `SSH_CONNECTION`.
    
Note, when starting a tunnel server from a regular SSH connection the `SSH_CONNECTION` variable gets inherited by the tunnel connections themselves! This means we need to also check for the absence of `SSH_TTY` because the tunnel server unsets this. `SSH_TTY` is set only in regular SSH sessions.

### DIFF
--- a/src/shared/Core/Interop/Linux/LinuxSessionManager.cs
+++ b/src/shared/Core/Interop/Linux/LinuxSessionManager.cs
@@ -95,6 +95,53 @@ public class LinuxSessionManager : PosixSessionManager
             return true;
         }
 
+        //
+        // We may also be able to launch a browser if we're inside a Visual Studio Code remote session.
+        // VSCode overrides the BROWSER environment variable to a script that allows the user to open
+        // the browser on their client machine.
+        //
+        // Even though we can start a browser, one piece of critical functionality we need is the ability
+        // to have that browser be able to connect back to GCM over localhost. There are several types
+        // of VSCode remote session, and only some of them automatically forward ports in such a way that
+        // the client browser can automatically connect back to GCM over localhost.
+        //
+        // * SSH [OK]
+        //   Connection over SSH to a remote machine.
+        //
+        // * Dev Containers [OK]
+        //   Connection to a container.
+        //
+        // * Dev Tunnels [Not OK - forwarded ports not accessible on the client via localhost]
+        //   Connection to a remote machine over the Internet using Microsoft Dev Tunnels.
+        //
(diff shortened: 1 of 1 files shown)


dscho (approved):
I like when changes make it easier to debug things!

mjcheetham:
> Sadly, for whatever reason, the remote connection over Microsoft Dev Tunnels does NOT automatically surface a forwarded port to localhost on the client. The forwarded port is only available via the devtunnels.ms URLs.

@kyle-rader-msft are you able to confirm how you connect to the "WaveSpace VMs" (dev tunnel or SSH), and confirm the behaviour of the auto-forwarded ports?

In my testing, forwarded ports in a tunnel remote session are not accessible via `localhost:<port>` on the _client_ end of the tunnel. This means the `localhost:<port>` redirect required by MSAL cannot work. I'm curious you seemed to indicate this _does_ work?

I'd imagine because dev tunnel ports are forwarded only to the relay, and not the client. SSH forwarding happens peer-to-peer.

PR #2258, opened 2026-02-03
Override MSAL's system webview handling

mjcheetham:
Since we have special handling on Linux (eg WSL support), we try to use our own logic for launching the user's default web browser for interactive authentication.

In the event we fail to start the browser, we fall back to using MSAL's logic, with a preference of Edge (Chromium-based) - largely because this is the only way to hook ourselves back into MSAL's default handling.


mjcheetham:
This depends on [this other PR](https://github.com/git-ecosystem/git-credential-manager/pull/2261) completing first.

PR #2259, opened 2026-02-03
Fix bug when reading linux defaults

mjcheetham:
If there is no GCM configuration defaults directory on Linux, we had been inadvertently returning `null` for the setting value!

Fix the issue by explictly returning `false` (no setting found) to `TryGetExternalDefault` rather than `true` (setting found).


src/shared/Core/Interop/Linux/LinuxSettings.cs
orbitalvoodoo: u

PR #2269, opened 2026-02-17
Allow installing on azurelinux

agollivermsft:
[`install-from-source.sh`](https://github.com/git-ecosystem/git-credential-manager/blob/v2.7.2/src/linux/Packaging.Linux/install-from-source.sh#L234) already allows you to install on `mariner`. But `mariner` is now `azurelinux`, so we ought to allow both.

The installer works fine without any other modifications. And I have had no issues using gcm on azurelinux3 so far.

I also add azurelinux3 to the test matrix.


mjcheetham (approved):

ubutatu (commented):
th

PR #2285, opened 2026-03-10
fix ambiguous call to string.Split

agollivermsft:
DEPENDS ON: #2286 (update from net472 to net8.0 on windows)

On an Ubuntu 24.04 image I get from work `./src/linux/Packaging.Linux/install-from-source.sh` fails with

```
dotnet --version
8.0.124
```

```
MSBuild version 17.8.49+7806cbf7b for .NET
  Determining projects to restore...
  Restored /tmp/tmp.LyZev1gOKL/src/linux/Packaging.Linux/Packaging.Linux.csproj (in 361 ms).
  Restored /tmp/tmp.LyZev1gOKL/src/shared/Microsoft.AzureRepos/Microsoft.AzureRepos.csproj (in 5.46 sec).
  Restored /tmp/tmp.LyZev1gOKL/src/shared/Core/Core.csproj (in 5.46 sec).
  Restored /tmp/tmp.LyZev1gOKL/src/shared/Atlassian.Bitbucket/Atlassian.Bitbucket.csproj (in 5.46 sec).
  Restored /tmp/tmp.LyZev1gOKL/src/shared/GitHub/GitHub.csproj (in 5.46 sec).
  Restored /tmp/tmp.LyZev1gOKL/src/shared/GitLab/GitLab.csproj (in 5.46 sec).
  Restored /tmp/tmp.LyZev1gOKL/src/shared/Git-Credential-Manager/Git-Credential-Manager.csproj (in 7.67 sec).
/tmp/tmp.LyZev1gOKL/src/shared/Core/Interop/Linux/LinuxConfigParser.cs(34,45): error CS0121: The call is ambiguous between the following methods or properties: 'string.Split(char[]?, StringSplitOptions)' and 'string.Split(string?, StringSplitOptions)' [/tmp/tmp.LyZev1gOKL/src/shared/Core/Core.csproj::TargetFramework=net8.0]

Build FAILED.

/tmp/tmp.LyZev1gOKL/src/shared/Core/Interop/Linux/LinuxConfigParser.cs(34,45): error CS0121: The call is ambiguous between the following methods or properties: 'string.Split(char[]?, StringSplitOptions)' and 'string.Split(string?, StringSplitOptions)' [/tmp/tmp.LyZev1gOKL/src/shared/Core/Core.csproj::TargetFramework=net8.0]
    0 Warning(s)
    1 Error(s)

Time Elapsed 00:00:17.29
```

[...]

PR #1418, opened 2023-09-20
Drop .NET Framework and target .NET 8 on all platforms

mjcheetham:
~*Depends on: https://github.com/git-ecosystem/git-credential-manager/pull/1417*~

Drop .NET Framework as the target framework on Windows in favour of .NET 8. Also update our Mac and Linux builds to .NET 8 from .NET 7. .NET 8 is an LTS release that is currently in a release candidate, which has a "go live" licence, and is due to GA in November 2023.

Now that we are targeting .NET 8 on all platforms we can drop all the .NET Framework specific code that we've been carrying. We also annotate all OS-specific code with the various attributes like `[SupportedOSPlatform]` to silence warnings and more importantly guard against accidentally using such code on the wrong platform (like in #1146).

PR #2286, opened 2026-03-10
update from net472 to net8.0 on windows

agollivermsft:
When working on #2285 (fix ambiguous call to string.Split) I saw a compile difference between windows and linux, which seemed to be caused by a mismatch between their target frameworks.

So I search and replaced `net472` with `net8.0` and simplified some of the logic in the `.csproj` files.

No idea if this is the right thing to do, or if any more work needs to be done, but the CI for my original PR seems to finish without issues now.


dscho:
How does this PR relate to/conflict with https://github.com/git-ecosystem/git-credential-manager/pull/1418?

mjcheetham:
> How does this PR relate to/conflict with #1418?

Sadly we are not able to accept this PR at this time to move from .NET Framework on Windows. We have some issues with supporting older Windows versions that we must first resolve. We are hopeful however to drop .NET Framework this year!

PR #2287, opened 2026-03-11
linuxsettings: Fix CS0121 string.Split call ambiguity with .NET 8.0

chewi:
The error in full:

    error CS0121: The call is ambiguous between the following methods or properties: 'string.Split(char[]?, StringSplitOptions)' and 'string.Split(string?, StringSplitOptions)' [src/shared/Core/Core.csproj::TargetFramework=net8.0]


chewi:
Evidently, that doesn't work for Windows for some reason. There is another way to do it, so I'll try that.

mjcheetham (changes requested):

mjcheetham (approved):

pand222 (commented):
Ac no 3349101005350 IFSC code CNRB003349 Canara Bank

PR #2294, opened 2026-03-16
Add Windows Group Policy ADMX/ADML files and generator script

pl4nty:
Windows registry keys are easier to configure using Group Policy templates rather than directly, especially with products like Intune that can only set registry keys through scripts or templates.

I've included a PowerShell script to generate the ADMX/ADML files from `configuration.md`, so they can be regenerated as the doc changes. This approach is used by large projects like browsers and vscode, but the files could be updated manually instead if you think the script is too brittle.

Example of generated UX in the local Group Policy editor:

image

### DIFF
--- a/docs/enterprise-config.md
+++ b/docs/enterprise-config.md
@@ -43,6 +43,34 @@ those of the [Git configuration][config] settings.
 The type of each registry key can be either `REG_SZ` (string) or `REG_DWORD`
 (integer).
 
+### Group Policy templates
+
+Windows Group Policy Administrative Templates (ADMX/ADML) for GCM are available
+in the [`src/windows/policy-templates`][policy-templates] directory. Copy the
+files to the local or domain policy definitions folder:
+
+**Local machine:**
+
+```text
+GitCredentialManager.admx        → %SystemRoot%\PolicyDefinitions\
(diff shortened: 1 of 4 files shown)


mjcheetham:
Hello @pl4nty ! Thanks for the PR - this looks very large! One concern I have is that the ADML/ADMX files will fall out of sync with the config/env docs, and although you have provided a script to generate the ADML/ADMX files, the docs are not strictly structured so we're opening ourselves up for pain.

I would be happier if we had first migrated to generating both our config & env docs from a structured (JSON? XML?) set of settings, with values etc. Then we could more reliably depend on generation of things like ADML/ADMX.

pl4nty:
No worries, I don't really like relying on structured markdown either. I could make an issue instead for further discussion? I've seen a few different implementations with pros/cons, eg Chromium's is pretty complex but does great i18n

mjcheetham:
> I don't really like relying on structured markdown either. I could make an issue instead for further discussion?

That would be great - thanks! I think moving to a generated config.md and environment.md from some structured definition of the available settings/values would be a good thing to have.. I always dislike having to maintain the env <-> config cross doc links etc.

PR #2299, opened 2026-03-23
Update Kerberos/NTLM docs

mjcheetham:
Clarify the risks of NTLM fallback during SPNEGO negotiation and offer instructions for disabling NTLM.

Detail the bug and workaround for enabling Kerberos authentication in Git.


dscho (approved):
Looks good to me!

PR #2303, opened 2026-03-31
Fix several small bugs that an AI agent detected

mjcheetham:
A collection of small bug fixes found with the help of an AI agent scanning the codebase. None of these are known to be actively exploited or causing user-reported issues, but each is a latent defect that could cause crashes, hangs, or incorrect behaviour under the right conditions.

## Fixes

**Credential UI** — `github/gitlab: use correct param order`
- Custom credential UI commands had swapped URL and username arguments.

**Stream extensions** — `streamextensions: fix a bug in multi-var reset handling`
- Multi-dictionary writer emitted the wrong value when a reset marker was followed by a single entry.

**GitHub auth** — `github: handle empty domain or enterprise hints`
- Null-reference exception when WWW-Authenticate header is missing domain/enterprise hints.

**GitHub account filtering** — `github: do not filter accounts outside of dotcom`
- Account filtering was applied outside of GitHub.com despite detecting it shouldn't be.

**Network diagnostics** — `diagnose: fix network diag to await HTTP requests`
- HTTP requests in the network diagnostic were not properly awaited.

**macOS notarize script** — `macos: add die function to notarize.sh script`
- Missing `die` function.

**Git stderr handling** — `git: drain stderr on IsInsideRepository`
- Potential hang when Git writes enough to stderr to fill the pipe buffer.

**Windows layout script** — `windows: fix layout.ps1 if symboloutput is not set`
- Null-reference trimming `SymbolOutput` when the variable is unset.

**OAuth device code UI** — `oauth: pass cancellation token to in-proc device code UI`
- Cancellation token was not forwarded, so dismissing the dialog didn't work.

**TRACE2 thread ID** — `trace2: fix main thread identification`
- Main thread ID starts at 1, not 0.

**TRACE2 perf format** — `trace2: fix crash in perf format for large elapsed times`
- `ArgumentOutOfRangeException` crash when elapsed time exceeds 9999 seconds.

**HTTP config** — `http: use correct http.sslAutoClientCert setting name`
- Setting was read from the wrong Git config section.

**TRACE2 writer cleanup** — `trace2: fix incomplete disposal of writers on cleanup`
- Forward iteration with removal skipped every other writer, leaking file handles.

**Git stderr redirect** — `git: fix crash when reading stderr from non-redirected processes`
- `InvalidOperationException` when `GetRemotes`/`CreateGitException` read non-redirected stderr.

**Executable lookup** — `environment: check execute permission in TryLocateExecutable`
- PATH scan didn't verify execute bits on POSIX, unlike the `which` it replaced.

**Installer Exec command** — `windows: fix en-dash characters in installer Exec command`
- Unicode en-dash characters instead of ASCII hyphens broke PowerShell 5.1 parameter parsing.

### DIFF
--- a/src/shared/Core/EnvironmentBase.cs
+++ b/src/shared/Core/EnvironmentBase.cs
@@ -138,7 +138,8 @@ namespace GitCredentialManager
                 {
                     string candidatePath = Path.Combine(basePath, program);
                     if (FileSystem.FileExists(candidatePath) && (pathsToIgnore is null ||
-                        !pathsToIgnore.Contains(candidatePath, StringComparer.OrdinalIgnoreCase)))
+                        !pathsToIgnore.Contains(candidatePath, StringComparer.OrdinalIgnoreCase))
+                        && FileSystem.FileIsExecutable(candidatePath))
                     {
                         path = candidatePath;
                         return true;
--- a/src/shared/Core/Git.cs
+++ b/src/shared/Core/Git.cs
@@ -146,6 +146,15 @@ namespace GitCredentialManager
                 }
 
                 git.Start(Trace2ProcessClass.Git);
+
+                // Drain and throw away stderr asynchronously to avoid a deadlock
+                // if the child process fills the stderr pipe buffer.
+                if (suppressStreams)
+                {
+                    git.Process.ErrorDataReceived += (_, _) => { };
+                    git.Process.BeginErrorReadLine();
+                }
+
                 string data = git.StandardOutput.ReadToEnd();
                 git.WaitForExit();
 
--- a/src/shared/Core/Git.cs
+++ b/src/shared/Core/Git.cs
@@ -176,6 +176,8 @@ namespace GitCredentialManager
         {
             using (var git = CreateProcess("remote -v show"))
             {
+                // Redirect stderr so we can check for 'not a git repository' errors
+                git.StartInfo.RedirectStandardError = true;
                 git.Start(Trace2ProcessClass.Git);
                 // To avoid deadlocks, always read the output stream first and then wait
                 // TODO: don't read in all the data at once; stream it
@@ -276,7 +278,9 @@ namespace GitCredentialManager
 
         public static GitException CreateGitException(ChildProcess git, string message, ITrace2 trace2 = null)
         {
-            var gitMessage = git.StandardError.ReadToEnd();
+            var gitMessage = git.StartInfo.RedirectStandardError
+                ? git.StandardError.ReadToEnd()
+                : null;
 
             if (trace2 != null)
                 throw new Trace2GitException(trace2, message, git.ExitCode, gitMessage);
--- a/src/shared/GitHub/GitHubAuthChallenge.cs
+++ b/src/shared/GitHub/GitHubAuthChallenge.cs
@@ -107,7 +107,15 @@ public class GitHubAuthChallenge : IEquatable<GitHubAuthChallenge>
 
     public override int GetHashCode()
     {
-        return Domain.GetHashCode() * 1019 ^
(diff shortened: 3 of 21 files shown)


mufasacubana8-pixel:
var = response  {
justo =this.name
}

cyberbossone-coder:
Nc

src/shared/GitHub/GitHubAuthChallenge.cs
```diff
@@ -107,7 +107,15 @@ public override bool Equals(object obj)
 
     public override int GetHashCode()
     {
-        return Domain.GetHashCode() * 1019 ^
-               Enterprise.GetHashCode() * 337;
+        int domainHash = Domain is null
+            ? 0
+            : StringComparer.OrdinalIgnoreCase.GetHashCode(Domain);
```
dscho, line 112: The original called `Domain.GetHashCode()`, which is simpler (but I guess does the same thing?). Maybe go back to the simpler expression? Same for `Enterprise.GetHashCode()`.

src/shared/Core/Git.cs
```diff
@@ -147,6 +147,13 @@ private string GetCurrentRepositoryInternal(bool suppressStreams)
 
                 git.Start(Trace2ProcessClass.Git);
                 string data = git.StandardOutput.ReadToEnd();
+
+                // Read all of standard error to prevent the process from hanging if it writes enough data to fill the buffer
```
dscho: Teeny tiny nit: `s/Ggit/Git/` in the commit message.

src/shared/Core/EnvironmentBase.cs
```diff
@@ -138,7 +138,8 @@ internal virtual bool TryLocateExecutable(string program, ICollection<string> pa
                 {
                     string candidatePath = Path.Combine(basePath, program);
                     if (FileSystem.FileExists(candidatePath) && (pathsToIgnore is null ||
-                        !pathsToIgnore.Contains(candidatePath, StringComparer.OrdinalIgnoreCase)))
+                        !pathsToIgnore.Contains(candidatePath, StringComparer.OrdinalIgnoreCase))
+                        && FileSystem.FileIsExecutable(candidatePath))
```
dscho, line 142: At first I was unsure that this is the specified behavior as per the POSIX standard. But [it is](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html#:~:text=The%20list%20shall%20be%20searched%20from%20beginning%20to%20end%2C%20applying%20the%20filename%20to%20each%20prefix%2C%20until%20an%20executable%20file%20with%20the%20specified%20name%20and%20appropriate%20execution%20permissions%20is%20found%2E). Maybe you want to quote that specification in the commit message?

dscho (approved):
I have three (minor) suggestions, but don't want to block merging if you feel that the PR branch is fine as-is.

src/shared/GitHub/GitHubAuthChallenge.cs
mjcheetham: Since `Domain` and `Enterprise` are not case sensitive, we should be generating a hash code that is the same regardless of case of these strings, so I think the change here it correct.

`GetHashCode` should align with `Equals`.

src/shared/Core/Git.cs
```diff
@@ -147,6 +147,13 @@ private string GetCurrentRepositoryInternal(bool suppressStreams)
...
+                if (suppressStreams)
+                {
+                    git.StandardError.ReadToEnd();
+                }
+
```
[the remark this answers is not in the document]
mjcheetham: Replaced with async read (and throw away) of stderr.

waseemazra19-cyber (commented):
Hi

PR #2326, opened 2026-04-23
http: fix SYSLIB0057 warning for X509Certificate2Collection.Import

mjcheetham:
Use `ImportFromPemFile` on modern .NET to resolve the `SYSLIB0057` deprecation warning, while keeping the original `Import` call on .NET Framework where the new API is not available.

We know this is always a PEM file since this is a Git defined option, and it is passed to libcurl via `CURLINFO_CAINFO` which expects PEM format.

### DIFF
--- a/src/shared/Core/HttpClientFactory.cs
+++ b/src/shared/Core/HttpClientFactory.cs
@@ -130,7 +130,11 @@ namespace GitCredentialManager
 
                     // Import the custom certs
                     X509Certificate2Collection certBundle = new X509Certificate2Collection();
+#if NETFRAMEWORK
                     certBundle.Import(certBundlePath);
+#else
+                    certBundle.ImportFromPemFile(certBundlePath);
+#endif
 
                     try
                     {


derrickstolee (approved):
My understanding, aided by Copilot, is that this doesn't fix any behavior issues right now but is helping to get ahead of a potential removal of this method from .NET (Core) while also preserving compilation in .NET Framework.

I also had to ask what a PEM file is, so Today I Learned:

```
What's a PEM file?

PEM (Privacy-Enhanced Mail) is a widely-used text-based format for storing cryptographic data — certificates, keys,
and certificate chains. PEM files are Base64-encoded, wrapped between distinctive headers like:

 -----BEGIN CERTIFICATE-----
 MIIDxTCCAq2gAwIBAgIQAqxc...
 -----END CERTIFICATE-----

They're the de facto standard for SSL/TLS certificates, especially in Unix/Git/curl ecosystems. The alternative
binary format is DER. PEM files typically use .pem, .crt, or .cer extensions.
```

mjcheetham:
> My understanding, aided by Copilot, is that this doesn't fix any behavior issues right now

As of .NET 10 (our target for Mac and Linux) the `Import` method is now deprecated, and as such we should move to an alternative. The documented alternative is to use a totally different type, but that isn't available in .NET Framework (our target for Windows), so we have to split compilation using `#if NETFRAMEWORK` (which we do all over the place.. sad times..) to keep both our .NET Framework and .NET Core builds happy.

dscho (approved):

PR #2330, opened 2026-04-28
install: update install-from-source to use .NET 10.0

mjcheetham:
The project now targets net10.0 but the install-from-source script still referenced .NET SDK 8.0. Update all references to 10.0 and fix the version parsing to use field-based extraction instead of fixed-width character slicing, which broke for two-digit major versions.


dscho (approved):
I particularly like that we now see breakages in PR builds.

PR #2333, opened 2026-04-28
release: run nuget publish job on windows

mjcheetham:
The legacy `NuGetCommand@2` task (used internally by the 1ES `output: nuget` template) requires Mono on Linux. Recent updates to the `ubuntu-x86_64-ado1es` image to Ubuntu 24.04+ removed Mono availability, breaking the publish step with:

> The task has failed because you are using Ubuntu 24.04 or later without mono installed.

Move the nuget publish job to the `win-x86_64-ado1es` image so the task uses the native `nuget.exe` and avoids the Mono dependency entirely. The same Windows pool is already used by the `dotnet_tool` build job that produces these packages.

There will be a small performance hit in the build pipeline since Windows agents take longer to provision (but better than this job failing as it does now).

### DIFF
--- a/.azure-pipelines/release.yml
+++ b/.azure-pipelines/release.yml
@@ -872,9 +872,13 @@ extends:
             dependsOn: release_validation
             condition: and(succeeded(), eq('${{ parameters.nuget }}', true))
             pool:
+              # Run on Windows so the underlying NuGetCommand@2 task can use the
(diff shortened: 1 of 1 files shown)


dscho (approved):

albertoblue87-netizen (commented):
Done

PR #2345, opened 2026-05-14
fix(config): enable file include in scoped queries

### LINKED ISSUE #1337: git-credential-manager adds `credential.helper` to git config even if it's already configured
git-credential-manager keeps adding the `credential.helper` configuration option to my $HOME/.config/git/config file repeatedly. This config option is already present in a file that is included by $HOME/.config/git/config, but gcm is unable to see this.

### LINKED ISSUE #1696: Allow specifying a custom gitconfig file
### Feature description

Right now GCM adds a `[credential]` directive to `~/.gitconfig`. I'd like to be able to tell GCM which gitconfig file to write the directive to.

Rationale: I manage my dotfiles in a Git repo and don't want to commit sensitive info to that repo. Since the `[credential]` directive may sometimes contain sensitive info, I'd like to add it to a separate `.gitconfig_local` file which I include from the main file and don't commit to the dotfiles repo.

### LINKED ISSUE #2328: includeIf'd git configs are ignored when checking for existing `credential.username` config
This is the same as https://github.com/git-ecosystem/git-credential-manager/issues/1581 , which OP closed because they claimed they could not repro it, but it definitely repros.

The issue is that gcm calls `git config --global` and `git config --local` to look for the `credential.username` config, both of which do not expand `includeIf'd` files.

(All of these commands run in the repository directory.)

```sh
$ git --version

git version 2.53.0

$ ~/.local/lib/git-credential-manager/git-credential-manager --version

2.7.3+5fa7116896c82164996a609accd1c5ad90fe730a

$ cat ~/.config/git/config

...
[includeIf "hasconfig:remote.*.url:https://dev.azure.com/**"]
	path = ~/.config/git/config-ado

$ cat ~/.config/git/config-ado   # Contains exactly what gcm tries to add to ~/.config/git/config because it think it's not set

...
[credential "azrepos:org/msazure"]
	username = foo@example.com

$ strace -fe execve -s 99 -- git pull --all --prune |& grep 'git.*config.*credential'

...
[pid 754151] execve("/usr/libexec/git/git", ["/usr/libexec/git/git", "config", "--null", "--local", "credential.azrepos:org/msazure.username"], 0x7f6e040cf250 /* 93 vars */) = 0
[pid 754152] execve("/usr/libexec/git/git", ["/usr/libexec/git/git", "config", "--null", "--global", "credential.azrepos:org/msazure.username"], 0x7f6e040cf250 /* 93 vars */) = 0
...

^C

$ git config --local 'credential.azrepos:org/msazure.username'   # With --local as gcm calls it, returns nothing

$ git config --global 'credential.azrepos:org/msazure.username'   # With --global as gcm calls it, returns nothing

$ git config 'credential.azrepos:org/msazure.username'   # Without --local or --global, returns value from includeIf'd config

foo@example.com

$ git config --global --list   # The reason is that config --global just returns the `includeIf` directives as-is instead of expanding them.

...
includeif.hasconfig:remote.*.url:https://dev.azure.com/**.path=~/.config/git/config-ado

$ git config --list   # ... whereas without --global it does expand them.

...
includeif.hasconfig:remote.*.url:https://dev.azure.com/**.path=~/.config/git/config-ado
...
credential.azrepos:org/msazure.username=foo@example.com
...
```

### LINKED ISSUE #2342: Each `brew upgrade`/`reinstall` writes `credential.helper` to `~/.gitconfig` even when already configured
## Problem
Each `brew upgrade` or `brew reinstall` writes `credential.helper` entries to `~/.gitconfig`,
even when `git-credential-manager` is already set as the `credential.helper` elsewhere — for example via an `[include]`d file — resulting in duplicate entries.

## Steps to reproduce

1. Set up `~/.gitconfig`:
   ```ini
   [include]
       path = .gitconfig.d/manager.gitconfig
   ```
1. Set up `.gitconfig.d/manager.gitconfig` with the `git-credential-manager` settings:
   ```ini
   [credential]
       helper =
       helper = /usr/local/share/gcm-core/git-credential-manager
       useHttpPath = true
   ```
1. Confirm the `credential.helper` is resolved:
   (EDIT)
   ```console
   $ git config --show-origin credential.helper
   file:/Users/skkzsh/.gitconfig.d/manager.gitconfig	/usr/local/share/gcm-core/git-credential-manager
   ```
1. Run `brew upgrade --cask git-credential-manager` or `brew reinstall --cask git-credential-manager`
1. Inspect `~/.gitconfig`

### Expected
`~/.gitconfig` should remain unchanged.

### Actual
The following duplicate entries are written to `~/.gitconfig`:

```ini
[credential]
    helper =
    helper = /usr/local/share/gcm-core/git-credential-manager
```

> [!NOTE]
> `[credential "https://dev.azure.com"] useHttpPath = true` is correctly detected and not duplicated.

## Version

```console
$ git credential-manager --version
2.8.0+d5e8a3442d48cc5efe7014e093e44242c0bd395f

$ git --version
git version 2.54.0
```

## Related

- #1337

becm:
Avoid duplication of entries already present in included files:
- [x] enable config cache for `Raw` values in all Git versions supporting `--show-scope`
- [x] consistent _include_ of files in Git config queries with [scope limit](https://git-scm.de/docs/git-config#Documentation/git-config.txt---includes)
- [x] add test for file includes

Improves consistency regarding _include_ behavior and

Fixes #2342
Fixes #2328
Fixes #1696
Fixes #1337

### DIFF
--- a/src/shared/Core.Tests/GitConfigurationTests.cs
+++ b/src/shared/Core.Tests/GitConfigurationTests.cs
@@ -656,5 +656,31 @@ namespace GitCredentialManager.Tests
             // Value should be canonicalized path, not raw "~/example"
             Assert.NotEqual("~/example", value);
         }
+
+        [Fact]
+        public void GitConfiguration_ScopedConfig_ProcessesIncludes()
+        {
+            const string configKey = "test.value";
+            const string configValue = "foo123";
+            string repoPath = CreateRepository(out string workDirPath);
+            ExecGit(repoPath, workDirPath, $"config --file {workDirPath}/git.config {configKey} {configValue}").AssertSuccess();
+            ExecGit(repoPath, workDirPath, $"config --global include.path {workDirPath}/git.config").AssertSuccess();
+            ExecGit(repoPath, workDirPath, $"config --local include.path {workDirPath}/git.config").AssertSuccess();
+
+            string gitPath = GetGitPath();
+            var trace = new NullTrace();
+            var trace2 = new NullTrace2();
+            var processManager = new TestProcessManager();
+
+            var git = new GitProcess(trace, trace2, processManager, gitPath, repoPath);
+            IGitConfiguration config = git.GetConfiguration();
+
+            // value of included file set in scopes
+            foreach (var scope in new GitConfigurationLevel[] { GitConfigurationLevel.Global, GitConfigurationLevel.Local })
+            {
+                Assert.True(config.TryGet(scope, GitConfigurationType.Raw, configKey, out string value));
+                Assert.Equal(configValue, value);
+            }
+        }
     }
 }
--- a/src/shared/Core/Application.cs
+++ b/src/shared/Core/Application.cs
@@ -204,7 +204,7 @@ namespace GitCredentialManager
 
             // Try to locate an existing app entry with a blank reset/clear entry immediately preceding,
             // and no other blank empty/clear entries following (which effectively disable us).
-            int appIndex = Array.FindIndex(currentValues, x => Context.FileSystem.IsSamePath(x, appPath));
+            int appIndex = Array.FindLastIndex(currentValues, x => Context.FileSystem.IsSamePath(x, appPath));
             int lastEmptyIndex = Array.FindLastIndex(currentValues, string.IsNullOrWhiteSpace);
             if (appIndex > 0 && string.IsNullOrWhiteSpace(currentValues[appIndex - 1]) && lastEmptyIndex < appIndex)
             {
@@ -219,7 +219,10 @@ namespace GitCredentialManager
 
                 // Add an empty value for `credential.helper`, which has the effect of clearing any helper value
                 // from any lower-level Git configuration, then add a second value which is the actual executable path.
-                config.Add(configLevel, helperKey, string.Empty);
+                if ((currentValues.Length == 0) || (lastEmptyIndex != (currentValues.Length - 1)))
+                {
+                    config.Add(configLevel, helperKey, string.Empty);
+                }
                 config.Add(configLevel, helperKey, appPath);
             }
 
--- a/src/shared/Core/GitConfiguration.cs
+++ b/src/shared/Core/GitConfiguration.cs
@@ -463,7 +463,7 @@ namespace GitCredentialManager
 
             // Fall back to original implementation
             string levelArg = GetLevelFilterArg(level);
-            using (ChildProcess git = _git.CreateProcess($"config --null {levelArg} --list"))
+            using (ChildProcess git = _git.CreateProcess($"config --null {levelArg} --includes --list"))
             {
                 git.Start(Trace2ProcessClass.Git);
                 // To avoid deadlocks, always read the output stream first and then wait
@@ -545,7 +545,7 @@ namespace GitCredentialManager
             // Fall back to individual git config command if cache not available
             string levelArg = GetLevelFilterArg(level);
             string typeArg = GetCanonicalizeTypeArg(type);
-            using (ChildProcess git = _git.CreateProcess($"config --null {levelArg} {typeArg} {QuoteCmdArg(name)}"))
+            using (ChildProcess git = _git.CreateProcess($"config --null {levelArg} --includes {typeArg} {QuoteCmdArg(name)}"))
             {
                 git.Start(Trace2ProcessClass.Git);
                 // To avoid deadlocks, always read the output stream first and then wait
@@ -667,7 +667,7 @@ namespace GitCredentialManager
             string levelArg = GetLevelFilterArg(level);
             string typeArg = GetCanonicalizeTypeArg(type);
 
-            var gitArgs = $"config --null {levelArg} {typeArg} --get-all {QuoteCmdArg(name)}";
+            var gitArgs = $"config --null {levelArg} --includes {typeArg} --get-all {QuoteCmdArg(name)}";
 
             using (ChildProcess git = _git.CreateProcess(gitArgs))
             {
@@ -705,7 +705,7 @@ namespace GitCredentialManager
             string levelArg = GetLevelFilterArg(level);
             string typeArg = GetCanonicalizeTypeArg(type);
 
-            var gitArgs = $"config --null {levelArg} {typeArg} --get-regex {QuoteCmdArg(nameRegex)}";
+            var gitArgs = $"config --null {levelArg} --includes {typeArg} --get-regex {QuoteCmdArg(nameRegex)}";
             if (valueRegex != null)
             {
                 gitArgs += $" {QuoteCmdArg(valueRegex)}";


derrickstolee:
@becm Thanks for identifying this key piece that connects to several issues. Could you add a test to demonstrate the behavior change to prevent regressions in the future?

becm:
@derrickstolee sadly only the 1st step, since:
- the assignment seems to be unconditional
- I did not even manage to get all instances where the `--list` request is issued.

becm:
@derrickstolee since the `--include` stuff is dependent on partnered Git version and file setup, I'd have no idea how to attempt a _proper_ test (existing examples just operate on unrelated mock data).

Will try to implement a test for (abstract) config data with the (hopefully) resolved corner case of `configure`.

becm:
Reverted PR to _primary_ goal, will tackle `git-credential-manager configure` issue in separate task.

@derrickstolee I'll try to add a test for file inclusion (global scope).

xz2m9kg2cw-oss:
Whats the most straightforward way to fix the problems?  Unfortunately, I am clueless about the program haha. Willing to learn more about this!Sent from my iPhoneOn May 22, 2026, at 5:58 AM, Marc Becker ***@***.***> wrote:﻿becm left a comment (git-ecosystem/git-credential-manager#2345)
Reverted PR to primary goal, will tackle git-credential-manager configure issue in separate task.
@derrickstolee I'll try to add a test for file inclusion (global scope).

becm:
@mjcheetham any further actions to be taken to accept this?

This change will essentially harmonizes behavior before and after `Git 2.54` .
#318 changed the default _includes_ behavior by splitting `global` and `local`.
#2268 basically introduced holistic config parsing with `--show-scope` again.

becm:
@derrickstolee it seems `git-credential-manager` only ever uses `GitConfigurationType.Raw` (in non-test code).

Would it be possible to introduce a further code path for `Git 2.26+` without the `--no-type` argument that _only_ uses/generates raw cache and only depends on `--show-scope`?

This would be an extension or (for non-ancient Git versions) a replacement for the current PR.

dscho:
The way I read this change, it constitutes a change of behavior. A good one, granted, but one that has a high potential of affecting existing users' setups (especially complex setups). As such, I'd be much more comfortable to make this behavioral change part of a major version bump. Seeing as v3.x is starting to take concrete shape, I would consider that version to be the natural inflection point to integrate this here PR.

becm:
@dscho behavior change already happened for `2.8.0` with #2268 and #2296 for _[certain newer](https://github.com/git-ecosystem/git-credential-manager/blame/e3e079e9bb4cbcf64f0612ad79fd3c630e5e8277/src/shared/Core/GitConfiguration.cs#L364)_ Git versions.

This combination effectively uses `--show-scope` and full config parsing again (which has [further impacts](https://github.com/git-ecosystem/git-credential-manager/issues/2328#issuecomment-4356583787)).
So this would be an attempt to harmonize behavior in a `2.8.1` for older Git versions.

dscho:
> behavior change already happened for `2.8.0` with #2268 and #2296 for _[certain newer](https://github.com/git-ecosystem/git-credential-manager/blame/e3e079e9bb4cbcf64f0612ad79fd3c630e5e8277/src/shared/Core/GitConfiguration.cs#L364)_ Git versions.

@becm sure, but the way I see it was that this was an inadvertent behavioral change. Here, we are aware of the change. That makes a huge difference.

becm:
@dscho the problem is inconsistent behavior when different Git versions [are involved](https://github.com/git-ecosystem/git-credential-manager/issues/2342#issuecomment-4466868483).

So a "bugfix" is to _intentionally lean into_ the (unintended) change.
The less desirable option would be to **revert** the introduction of the config cache.

If adding config cache did not enforce `--type` support (only `GitConfigurationType.Raw` is used at present)
the new approach would already be applicable down to `Git v2.26` (introduction of `--show-scope`).

dscho:
Now that `main` targets v3.x, I think this change should go in.

becm:
If there is _serious_ intent to merge, I'd invest work to submit a better (more extensive) approach.
The code paths requiring `--type` are only used in test code so there are 2 viable options:
- drop the support for typed config cache (less complex code) or
- add [separate version checks](https://github.com/git-ecosystem/git-credential-manager/blob/c268a7b63687d53381d2b3e6a18f077127e8b73c/src/Core/GitConfiguration.cs#L387) for `useRawCache` and `useTypedCache`
(and skip the `--no-type` [option](https://github.com/git-ecosystem/git-credential-manager/blob/c268a7b63687d53381d2b3e6a18f077127e8b73c/src/Core/GitConfiguration.cs#L403)) for the `Raw` config type.

Only using `--includes` still has a issues when using [includeif.hasconfig](https://git-scm.com/docs/git-config#Documentation/git-config.txt-hasconfigremoteurl)
→ local config can effect behavior in global scope (#2328)

The approach using `--includes` would still be applicable for _ancient_ Git versions.

becm:
Revised for `v3.x` with cache enabled for `GitConfigurationType.Raw` starting with [Git v2.26](https://github.com/git/git/commit/145d59f48233c64cb8a9262c9f1451cc7d66b530).
Arguments to `git config` invocations adjusted to:
- allow use of raw types with older Git versions missing config type support
- use `--includes` argument for scoped config queries to get closest match to cached results.

dscho:
@becm thank you for rebasing! @mjcheetham and I should be able to focus on getting this merged this coming week.

src/Core/GitConfiguration.cs
```diff
@@ -349,11 +351,23 @@ internal GitProcessConfiguration(ITrace trace, GitProcess git, bool useCache)
...
+                _useConfigTypeCache = false;
+            }
+            else
+            {
+                _useConfigTypeCache = useCache;
             }
 
-            _useCache = useCache;
-            _cache = useCache ? new Dictionary<GitConfigurationType, ConfigCache>() : null;
+            if (useCache && !SupportsConfigListRaw(git))
```
mjcheetham, line 361: The `list` subcommand to `git config` was only [added in Git 2.46](https://github.com/git/git/commit/14970509c6266d2347fe1a15dbdc0b9b697d2e1a), but we are gating the ability to use the (raw) cache where we call `git config list` on Git 2.26.

We should update the cache loading command to use the `git config --list` syntax instead.

src/Core.Tests/GitConfigurationTests.cs
```diff
@@ -656,5 +656,37 @@ public void GitConfiguration_TypedQuery_CanonicalizesValues()
...
+        [Fact]
+        public void GitConfiguration_ScopedConfig_ProcessesIncludes()
+        {
+            const string configKey = "test.value";
+            const string configTrigger = "remote.*.url:http://localhost/**";
+            const string configValue = "foo123";
+            string repoPath = CreateRepository(out string workDirPath);
+
+            ExecGit(repoPath, workDirPath, $"config --file {workDirPath}/git.config.glob {configKey} {configValue}").AssertSuccess();
+            ExecGit(repoPath, workDirPath, $"config --global includeif.hasconfig:{configTrigger}.path {workDirPath}/git.config.glob").AssertSuccess();
```
mjcheetham, line 669: Running this test will permanently modify the developer's `--global` config. We should either wrap this test in a `try-finally` and undo this global config mutation in the finally block, or better yet use the [`GIT_CONFIG_GLOBAL` environment variable](https://git-scm.com/docs/git#Documentation/git.txt-GITCONFIGGLOBAL) to redirect reads and writes to the 'global' config to a temporary file we can throw away.

mjcheetham (changes requested):
Looking good so far, but a few issues at the moment.

src/Core/GitConfiguration.cs
becm: Added additional version check to use (now deprecated) `--list` for Git versions `<= 2.46.0`.

src/Core.Tests/GitConfigurationTests.cs
becm: Global environment guard added for new test.
_there are however still **other** tests that touch (and hopefully restore) the global settings file!_

mjcheetham (approved):
Thanks for persisting with this!

PR #2346, opened 2026-05-17
azrepos: fall back to OAuth when org DisablePatCreationPolicyViolation policy is set

p00j4:
## Problem

When an Azure DevOps organization disables PAT creation via policy, GCM surfaces a raw fatal error with no recovery path:

```
fatal: Failed to create PAT: DisablePatCreationPolicyViolation
```
image

The user has no indication what to do. The [documented fix](https://learn.microsoft.com/en-us/azure/devops/repos/git/set-up-credential-managers?view=azure-devops#configure-microsoft-entra-id-authentication-recommended) (`git config --global credential.azreposCredentialType oauth`) is buried in docs.

The reason is, after installing the Extension and trying to clone a repo, it ends-up creating teh PAT with "package" only permissions because the org's default PAT creation policy is been disabled.  And it's not there in the default allowed list (AZDO PAT Allowlist - CORP Tenant - SERVICE ACCOUNTS ONLY)

## Fix/WorkAround
In `AzureReposHostProvider.GetCredentialAsync`, wrap PAT creation in a `catch` that recognizes the `DisablePatCreationPolicyViolation` error and:

1. Falls back to OAuth for the current invocation
2. Prints a clear warning with the exact `git config` command to make it permanent

A `_forcedOAuth` instance flag propagates the fallback to `StoreCredentialAsync` and `EraseCredentialAsync` within the same process (necessary because `Settings` caches git config entries on first read and won't pick up a disk write mid-invocation).

**Deliberately not done:** auto-writing `credential.azreposCredentialType = oauth` to global git config — that setting affects all Azure DevOps orgs, not just the one with the policy, which would be too broad.

## User experience after this change

```
warning: PAT creation is disabled by your Azure DevOps organization policy.
hint: To permanently use OAuth, run:
hint:   git config --global credential.azreposCredentialType oauth
```

The clone then succeeds via OAuth without any user action needed for that invocation.

## Test plan
- [ ] Clone from an org with `DisablePatCreationPolicyViolation` policy — should fall back to OAuth and print the hint
- [ ] Clone from a normal org — PAT path unchanged
- [ ] After running the hinted `git config` command, subsequent clones go directly to OAuth (no policy call at all)

### DIFF
--- a/src/shared/Microsoft.AzureRepos/AzureReposHostProvider.cs
+++ b/src/shared/Microsoft.AzureRepos/AzureReposHostProvider.cs
@@ -20,6 +20,8 @@ namespace Microsoft.AzureRepos
         private readonly IMicrosoftAuthentication _msAuth;
         private readonly IAzureDevOpsAuthorityCache _authorityCache;
         private readonly IAzureReposBindingManager _bindingManager;
+        // Set to true if PAT creation was blocked by org policy this invocation, forcing OAuth fallback.
+        private bool _forcedOAuth;
 
         public AzureReposHostProvider(ICommandContext context)
             : this(context, new AzureDevOpsRestApi(context), new MicrosoftAuthentication(context),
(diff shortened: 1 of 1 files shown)


src/shared/Microsoft.AzureRepos/AzureReposHostProvider.cs
```diff
@@ -118,20 +120,28 @@ public async Task<GetCredentialResult> GetCredentialAsync(InputArguments input)
...
                 }
                 else
                 {
                     _context.Trace.WriteLine("Existing credential found.");
+                    return new GetCredentialResult(credential);
                 }
-
-                return new GetCredentialResult(credential);
             }
-            else
```
mr-otmarg, line 131: when removing this `else` you should clean up curly braces also or just leave the else.

src/shared/Microsoft.AzureRepos/AzureReposHostProvider.cs
```diff
@@ -20,6 +20,8 @@ public class AzureReposHostProvider : DisposableObject, IHostProvider, IConfigur
         private readonly IMicrosoftAuthentication _msAuth;
         private readonly IAzureDevOpsAuthorityCache _authorityCache;
         private readonly IAzureReposBindingManager _bindingManager;
+        // Set to true if PAT creation was blocked by org policy this invocation, forcing OAuth fallback.
+        private bool _forcedOAuth;
```
mjcheetham, line 24: Sadly this is an incomplete implementation. Because Git calls GCM multiple times, as a new process invocation each time, we 'forget' the value of `_forcedOAuth` here when it comes to the `store` call that Git does following a successful `get`  + remote op.

Without first implementing the functionality of https://github.com/git-ecosystem/git-credential-manager/issues/2057, we'd be better just printing the error and resolution message (changing to OAuth mode).

mjcheetham (changes requested):

mjcheetham:
Thank you for your contribution here! However, with GCM 3.0 (the next release) we plan to make `oauth` mode the default, so this fallback is not required.

Issue #2057, opened 2025-09-26
Credential protocol enhancement: Retry mechanism and stateful operations


dscho:
The Git credential protocol is really a set of three commands that credential helpers (such as GCM) should implement: `get`, `store`, and `erase`. The credential helpers are run with these commands as separate processes at different points of the overall Git remote operation, such as `fetch` or `push`.

Starting in Git version 2.46 the protocol now supports persisting state between `get + store/erase` command pairs. Since each credential helper command is invoked as a new process, it was previously not possible to persist metadata between calls. Examples of state that we may wish to keep are options to 'remember this account' to help with multi-account scenarios or  maintain other caches of dynamically resolved information. 

Also starting in version 2.46, the protocol added the ability to advertise a multi-stage authentication process using the continue protocol key. This lets Git know that if a subsequent  HTTP 401 response is encountered following an authenticated HTTP request, that this is not the end of the operation, and that the credential helper should be consulted once more. 

Using both the `continue` and `state[]` protocol capabilities, GCM will be able to perform optimistic account selection in the case of multiple accounts, or an available OS/broker account in the case of MSAL-based authentication. Without the ability to 'retry', selecting the wrong account means GCM doesn't get the opportunity to present an account selection, nor to understand that its previous 'best guess' was incorrect.

Author: `@mjcheetham`

PR #2357, opened 2026-06-11
Protocol foundation: capability negotiation, `GitRequest`/`GitResponse`, `state[]` + `continue`

mjcheetham:
Refs: #2057

## Summary

Git 2.46 added a capability negotiation handshake to the credential helper protocol, gating a set of new attributes — `state[]`, `continue`, `authtype`, `credential`, `ephemeral`, `password_expiry_utc`, `oauth_refresh_token`.

This PR lays the foundation for GCM to participate in that protocol and ships the first capability end-to-end (`state` + `continue`). It is purely infrastructure: behaviour for end-users is unchanged for any existing flow, but the host-provider surface is now in a shape where individual provider features (smarter account selection, multistage auth, etc.) can land one commit at a time without churning `IHostProvider` again.

## What's in this PR

Eight commits, ordered as a readable story:

1. **`hostprovider: rename InputArguments to GitRequest`**
   Pure rename. `InputArguments` framed the type as a low-level argument bag; `GitRequest` matches its actual role as the anchor for everything a host provider needs to know about a single credential helper invocation. No methods change shape.

2. **`hostprovider: add capability negotiation and GitResponse`**
   Introduces the `GitCapabilities` `[Flags]` enum and a `GitCapabilitiesExtensions` helper that parses incoming `capability[]` lines, advertises GCM's own set (empty for now), and renders flags back to wire names. Surfaces a typed `Capabilities` property on `GitRequest`. Replaces the old `GetCredentialResult` with a new `GitResponse` type. Wires the handshake through `GetCommand` end-to-end — the intersection of what Git and GCM advertised is echoed back on the response (currently empty, so no `capability[]` lines appear in practice, but the plumbing is live).

3. **`hostprovider: add 'git credential capability' action`**
   Git 2.46 also added a `capability` action distinct from get/store/erase. Implements it as `CapabilityCommand` (no stdin, no provider selection — capabilities are a global property of the helper). Today emits just `version 0`; advertised capabilities will appear here automatically as flags are added to `GitCapabilitiesExtensions.Advertised`.

4. **`commands: hide Git credential helper actions from --help`**
   `get`/`store`/`erase`/`capability` are protocol entry points for Git, not user-facing commands. Mark them `IsHidden = true` so `git-credential-manager --help` no longer lists them alongside the commands a human actually runs (configure, diagnose, per-provider subcommands).

5. **`response: add Ok/Cancel factories and wire quit=1`**
   Gives providers a non-exceptional way to express "user cancelled the prompt" / "no eligible account". `GitResponse.Cancel()` emits `quit=1` so Git aborts the credential pipeline (no terminal re-prompt after the user already cancelled in a GUI dialog). `GitResponse.Ok(credential)` is the named factory for the success case. Mutually-exclusive shape enforced by the constructor.

6. **`response: add Yield() for helpers that have nothing to offer`**
   Third shape: "I have nothing to contribute for this request, but I'm not stopping you — try other helpers or fall back to your interactive prompt". Translates to the polite empty response (terminating blank line, no credential fields, no quit signal) so Git proceeds to the next helper. Today providers had no clean way to say this — they either threw (wrong: a no-op isn't an error) or returned empty credentials (wrong: that's the WIA signal and gets stored).

7. **`hostprovider: wire the state and continue protocol capability`**
   The marquee feature. Adds the `State` flag to the advertised set, surfaces `GitRequest.State` as a lazy `IReadOnlyDictionary<string,string>` filtered to the reserved `gcm.` prefix, and grows `GitResponse` with a fourth shape: `Continue(credential)` returns a credential while signalling that another exchange is expected. State writes go through a single validating `SetState` (or the fluent `WithState`); reads through the read-only view. The shape matrix becomes `Ok / Continue / Cancel / Yield`, all mutually exclusive. This is what unlocks optimistic account selection, multistage NTLM/Kerberos flows, and any provider scenario that benefits from per-request memory across the get → store/erase cycle.

8. **`github: delete stranded SelectAccountCommandImpl`**
   Cleanup that fell off the bottom of a previous refactor (`aab6fef` folded `.UI.Avalonia` into `.UI` but missed this file). True no-op — no `.csproj`, no consumer, no behaviour change.

## What this enables (deliberately *not* in this PR)

These belong in follow-up PRs that each ship a concrete user-visible win on top of this foundation:

- **AzRepos smarter account selection.**
  With `state[]` + `continue` the provider can return a token optimistically (e.g. the only cached account in the request's tenant), record which account it tried in state, and pick a different one on a 401 retry — all within a single Git operation. Sits on top of the binding-manager refactor and account-pool commands that already use this infrastructure on a separate branch.

- **Picker UI prompt-and-remember.**
  The picker emits a "remember this choice" checkbox; the result is threaded through state on the `get` response and persisted on the next `store` — so a binding is only written when the chosen credential actually worked.

- **GitHub / Bitbucket / GitLab state-driven UX.**
  Each is its own PR when there's a concrete win.

- **Other capability-gated attributes** (`authtype`, `credential`, `password_expiry_utc`, `oauth_refresh_token`, `ephemeral`).
  One capability flag at a time, each in its own PR.

## Compatibility

- **Older Git (no capability handshake).**
  Git never sends `capability[]`, so `GitRequest.Capabilities` is `None`, and the helper-side handshake echoes nothing back. Every existing provider path works exactly as today.

- **Existing providers.**
  The `InputArguments` → `GitRequest` rename is mechanical and complete across the repo; third-party providers building against GCM source will pick up the rename but need no behavioural changes. The `GetCredentialResult` → `GitResponse` swap preserves the existing `AdditionalProperties` escape hatch (still used by `GenericHostProvider` for the `ntlm=allow` signal).

- **`quit=1` on Cancel.**
  Previously, a cancelled prompt surfaced as `fatal: ...` via the top-level exception handler. After this PR, providers that have migrated to `GitResponse.Cancel()` emit `quit=1` instead — same end result for the user (no credential), cleaner contract with Git. Providers that still throw are unchanged.

## Testing

- Full unit test suite passes (Core, AzureRepos, GitHub, Bitbucket, GitLab).
- New tests cover capability parsing/rendering, response shape mutual exclusion, state validation rules, and the `quit=1` / yield emission paths.
- The `capability` action is verified end-to-end against the documented `version 0` + `capability <name>` response format.

## Notes for reviewers

- The eight commits are organised so each one tells a focused story; reviewing commit-by-commit is recommended over the squashed diff. The renames in commit 1 are noisy but mechanical; the substance starts in commit 2.
- The `Capabilities.None` advertised set in commit 2 is deliberate. Commit 7 is the one that flips on the first real capability; landing it as a separate commit makes the end-to-end wiring visible without buried-in-noise.
- No provider behaviour changes in this PR. Anyone landing a feature on top is expected to add tests for the user-visible win, not for the underlying capability plumbing.

Closes part of #2057.


src/Core/GitRequest.cs
```diff
@@ -27,27 +26,21 @@ public InputArguments(IDictionary<string, string> dict)
...
-        public InputArguments(IDictionary<string, IList<string>> dict)
+        public GitRequest(IDictionary<string, IList<string>> dict)
         {
             EnsureArgument.NotNull(dict, nameof(dict));
 
             // Wrap the dictionary internally as readonly
             _dict = new ReadOnlyDictionary<string, IList<string>>(dict);
         }
 
-        #region Common Arguments
```
dscho, line 38: Did this kind of `#region ... #endregion` annotation do anything load-bearing before?

src/Core.Tests/GitRequestTests.cs
```diff
@@ -387,5 +387,81 @@ public void GitRequest_Capabilities_UnknownNames_AreSilentlyDiscarded()
...
+                    "gcm.azure.tenant=contoso",
+                },
+            };
+
+            var request = new GitRequest(dict);
+
+            Assert.Equal(2, request.State.Count);
+            Assert.Equal("alice", request.State["github.account"]);
+            Assert.Equal("contoso", request.State["azure.tenant"]);
+            Assert.False(request.State.ContainsKey("other-helper.foo"));
```
dscho, line 425: It is theoretically possible for another credential helper to inject `gcm.*` values, right? This means that users should be careful which credential helpers they install...

src/Core/Commands/GetCommand.cs
```diff
@@ -19,27 +17,27 @@ public GetCommand(ICommandContext context, IHostProviderRegistry hostProviderReg
...
             if (response.IsYielded)
             {
-                // Provider has nothing to contribute but does not want to stop the
-                // pipeline. Emit an empty response (just the terminating blank line)
-                // so Git proceeds to the next helper or its interactive prompt.
+                // Empty response (just the terminating blank line) lets Git
+                // proceed to the next helper or its interactive prompt.
                 Context.Trace.WriteLine("Provider yielded; emitting empty response.");
-                Context.Streams.Out.WriteLine();
+                stdout.WriteLine();
```
dscho, line 40: It's a bit funny that `Context.Streams.Out` is simplified via assigning it to `stdout`, but `Context.Trace` does not enjoy the same treatment. Just a nit, not a blocker.

dscho (approved):
Wow, judging by the time it took to review this, this must have been a _ton_ of work. Well done!

src/Core/GitRequest.cs
mjcheetham: Regions are just like smart comments, that allow collapsing/folding of regions of code in the editor. Opinions on them vary but have shifted towards "don't bother using them".

src/Core.Tests/GitRequestTests.cs
mjcheetham: Correct. The credential helper system is supposed to be cooperative/global. There is no affinity between helpers and their responses.

PR #2365, opened 2026-06-29
build: drop the .NET Framework target

mjcheetham:
## Summary

Git Credential Manager no longer targets .NET Framework. This drops the `net472` TFM everywhere, leaving `net10.0` as the single target on all platforms (Windows included), then deletes the code that only existed to support the old runtime and tightens our OS-platform guarantees.

## Why

The product builds against the CoreCLR on every platform now. The .NET Framework target only added a second build pivot, a pile of conditional code, and Windows-only baggage. Collapsing to one modern TFM simplifies the build, the source, and everything downstream (packaging, CI, trimming/AOT work that follows).

## What changes

- **Drop `net472` from all projects and build scripts.** Windows now builds the single `net10.0` TFM like every other platform.
- **Remove dead `#if NETFRAMEWORK` code.** With only the CoreCLR left, this code was unreachable.
- **Add `[SupportedOSPlatform]` attributes** to OS-specific types so use of platform-specific behaviour is gated at compile time.
- **Suppress CA1416 in test projects.** Our xUnit attributes (`[WindowsFact]`, `[PosixFact]`, …) skip OS-specific tests at runtime; the analyzer can't see that, so the warning is noise in tests.

## Caveat

Removing the .NET Framework target also removes the **embedded web view** for Microsoft authentication — a netfx-only, Windows-only feature. The goal will be to move to the Windows "WAM" broker by default before release to replace the UX offered by embedded wv.

### DIFF
--- a/src/shared/Core/Authentication/AuthenticationBase.cs
+++ b/src/shared/Core/Authentication/AuthenticationBase.cs
@@ -60,11 +60,7 @@ namespace GitCredentialManager.Authentication
             // Write the standard input to the process if we have any to write
             if (standardInput is not null)
             {
-#if NETFRAMEWORK
-                await standardInput.BaseStream.CopyToAsync(process.StandardInput.BaseStream);
-#else
                 await standardInput.BaseStream.CopyToAsync(process.StandardInput.BaseStream, ct);
-#endif
                 process.StandardInput.Close();
             }
 
--- a/src/shared/Core/Authentication/MicrosoftAuthentication.cs
+++ b/src/shared/Core/Authentication/MicrosoftAuthentication.cs
@@ -16,10 +16,7 @@ using GitCredentialManager.UI.Controls;
 using GitCredentialManager.UI.ViewModels;
 using GitCredentialManager.UI.Views;
 using Microsoft.Identity.Client.AppConfig;
-
-#if NETFRAMEWORK
 using Microsoft.Identity.Client.Broker;
-#endif
 
 namespace GitCredentialManager.Authentication
 {
@@ -598,7 +595,6 @@ namespace GitCredentialManager.Authentication
             // to save on the distribution size of the .NET builds (no need for MSALRuntime bits).
             if (enableBroker)
             {
-#if NETFRAMEWORK
                 appBuilder.WithBroker(
                     new BrokerOptions(BrokerOptions.OperatingSystems.Windows)
                     {
@@ -606,7 +602,6 @@ namespace GitCredentialManager.Authentication
                         MsaPassthrough = msaPt,
                     }
                 );
-#endif
             }
 
             IPublicClientApplication app = appBuilder.Build();
@@ -939,9 +934,8 @@ namespace GitCredentialManager.Authentication
 
         public bool CanUseBroker()
         {
-#if NETFRAMEWORK
             // We only support the broker on Windows 10+ and in an interactive session
-            if (!Context.SessionManager.IsDesktopSession || !PlatformUtils.IsWindowsBrokerSupported())
+            if (!PlatformUtils.IsWindowsBrokerSupported() || !Context.SessionManager.IsDesktopSession)
             {
                 return false;
             }
@@ -958,34 +952,26 @@ namespace GitCredentialManager.Authentication
             }
 
             return defaultValue;
-#else
-            // OS broker requires .NET Framework right now until we migrate to .NET 5.0 (net5.0-windows10.x.y.z)
-            return false;
-#endif
         }
 
         private bool CanUseEmbeddedWebView()
         {
-            // If we're in an interactive session and on .NET Framework then MSAL can show the WinForms-based embedded UI
-#if NETFRAMEWORK
-            return Context.SessionManager.IsDesktopSession;
-#else
+            // TODO: check for desktop session once embedded web view is added back
+            // return Context.SessionManager.IsDesktopSession;
             return false;
-#endif
         }
 
         private void EnsureCanUseEmbeddedWebView()
         {
-#if NETFRAMEWORK
-            if (!Context.SessionManager.IsDesktopSession)
-            {
-                throw new Trace2InvalidOperationException(Context.Trace2,
-                    "Embedded web view is not available without a desktop session.");
-            }
-#else
+            // TODO: check for desktop session once embedded web view is added back
+            // if (!Context.SessionManager.IsDesktopSession)
+            // {
+            //     throw new Trace2InvalidOperationException(Context.Trace2,
+            //         "Embedded web view is not available without a desktop session.");
+            // }
+
             throw new Trace2InvalidOperationException(Context.Trace2,
                 "Embedded web view is not available on .NET Core.");
-#endif
         }
 
         private bool CanUseSystemWebView(IPublicClientApplication app, Uri redirectUri)
--- a/src/shared/Core/Interop/Windows/WindowsSettings.cs
+++ b/src/shared/Core/Interop/Windows/WindowsSettings.cs
@@ -21,7 +21,6 @@ namespace GitCredentialManager.Interop.Windows
         {
             value = null;
 
-#if NETFRAMEWORK
             // Check for machine (HKLM) registry keys that match the Git configuration name.
             // These can be set by system administrators via Group Policy, so make useful defaults.
             using (Microsoft.Win32.RegistryKey configKey = Microsoft.Win32.Registry.LocalMachine.OpenSubKey(Constants.WindowsRegistry.HKConfigurationPath))
@@ -48,9 +47,6 @@ namespace GitCredentialManager.Interop.Windows
 
                 return true;
             }
-#else
-            return base.TryGetExternalDefault(section, scope, property, out value);
-#endif
         }
     }
 }
--- a/src/shared/Core/Interop/Windows/WindowsSettings.cs
+++ b/src/shared/Core/Interop/Windows/WindowsSettings.cs
@@ -1,9 +1,11 @@
+using System.Runtime.Versioning;
 
 namespace GitCredentialManager.Interop.Windows
 {
     /// <summary>
     /// Reads settings from Git configuration, environment variables, and defaults from the Windows Registry.
     /// </summary>
+    [SupportedOSPlatform("windows")]
     public class WindowsSettings : Settings
     {
         private readonly ITrace _trace;
--- a/src/shared/Core/PlatformUtils.cs
+++ b/src/shared/Core/PlatformUtils.cs
@@ -31,7 +31,6 @@ namespace GitCredentialManager
                 return false;
             }
 
-#if NETFRAMEWORK
             // Check for machine (HKLM) registry keys for Cloud PC indicators
             // Note that the keys are only found in the 64-bit registry view
             using (Microsoft.Win32.RegistryKey hklm64 = Microsoft.Win32.RegistryKey.OpenBaseKey(Microsoft.Win32.RegistryHive.LocalMachine, Microsoft.Win32.RegistryView.Registry64))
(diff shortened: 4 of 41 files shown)


src/shared/Core/Authentication/MicrosoftAuthentication.cs
```diff
@@ -16,10 +16,7 @@
 using GitCredentialManager.UI.ViewModels;
 using GitCredentialManager.UI.Views;
 using Microsoft.Identity.Client.AppConfig;
-
-#if NETFRAMEWORK
 using Microsoft.Identity.Client.Broker;
```
derrickstolee, line 19: Since we are dropping this `#if`, then shouldn't we delete the block it was around?

src/shared/Core/Authentication/MicrosoftAuthentication.cs
```diff
@@ -598,15 +595,13 @@ private async Task<IPublicClientApplication> CreatePublicClientApplicationAsync(
...
             {
-#if NETFRAMEWORK
                 appBuilder.WithBroker(
                     new BrokerOptions(BrokerOptions.OperatingSystems.Windows)
                     {
                         Title = "Git Credential Manager",
                         MsaPassthrough = msaPt,
                     }
                 );
-#endif
```
derrickstolee, line 609: Same here. Shouldn't we be removing these bits now that we are off of the framework?

src/shared/Core/Authentication/MicrosoftAuthentication.cs
```diff
@@ -939,9 +934,8 @@ public HttpClient GetHttpClient()
 
         public bool CanUseBroker()
         {
-#if NETFRAMEWORK
             // We only support the broker on Windows 10+ and in an interactive session
-            if (!Context.SessionManager.IsDesktopSession || !PlatformUtils.IsWindowsBrokerSupported())
+            if (!PlatformUtils.IsWindowsBrokerSupported() || !Context.SessionManager.IsDesktopSession)
```
derrickstolee, line 938: Why the swap here? Isn't this identical outside of the short-circuit behavior?

src/shared/Core/Authentication/MicrosoftAuthentication.cs
```diff
@@ -958,34 +952,26 @@ public bool CanUseBroker()
             }
 
             return defaultValue;
-#else
-            // OS broker requires .NET Framework right now until we migrate to .NET 5.0 (net5.0-windows10.x.y.z)
-            return false;
-#endif
```
derrickstolee, line 964: ah, so since we have migrated we now have a different broker? Interesting.

src/shared/Core/Authentication/MicrosoftAuthentication.cs
```diff
@@ -958,34 +952,26 @@ public bool CanUseBroker()
...
-                throw new Trace2InvalidOperationException(Context.Trace2,
-                    "Embedded web view is not available without a desktop session.");
-            }
-#else
+            // TODO: check for desktop session once embedded web view is added back
+            // if (!Context.SessionManager.IsDesktopSession)
+            // {
+            //     throw new Trace2InvalidOperationException(Context.Trace2,
+            //         "Embedded web view is not available without a desktop session.");
+            // }
```
derrickstolee, line 971: Intentional commented-out-code?

src/shared/Core/Interop/Windows/WindowsSettings.cs
```diff
@@ -21,7 +23,6 @@ protected internal override bool TryGetExternalDefault(string section, string sc
         {
             value = null;
 
-#if NETFRAMEWORK
```
derrickstolee, line 24: This is another example where I think the deleted case is backwards.

src/shared/Core/UI/AvaloniaUi.cs
```diff
@@ -65,22 +65,15 @@ public static Task ShowWindowAsync(Func<Window> windowFunc, object dataContext,
                 {
                     var appBuilder = AppBuilder.Configure<AvaloniaApp>();
 
-#if NETFRAMEWORK
```
derrickstolee, line 68: Another mismatched case?

src/shared/Core/PlatformUtils.cs
```diff
@@ -197,11 +199,9 @@ public static bool IsElevatedUser()
         {
             if (IsWindows())
             {
-#if NETFRAMEWORK
                 var identity = System.Security.Principal.WindowsIdentity.GetCurrent();
                 var principal = new System.Security.Principal.WindowsPrincipal(identity);
                 return principal.IsInRole(System.Security.Principal.WindowsBuiltInRole.Administrator);
-#endif
             }
```
derrickstolee, line 205: Another case of keeping framework-specific code.

src/shared/Core/PlatformUtils.cs
```diff
@@ -48,9 +47,6 @@ public static bool IsDevBox()
 
                 return w365Value is not null && Guid.TryParse(partnerValue, out Guid partnerId) && partnerId == Constants.DevBoxPartnerId;
             }
-#else
-            return false;
-#endif
```
derrickstolee, line 53: More "keeping framework code".

src/shared/Core/UI/AvaloniaUi.cs
```diff
@@ -65,22 +65,15 @@ public static Task ShowWindowAsync(Func<Window> windowFunc, object dataContext,
...
                     // Set custom rendering options and modes if required
                     if (PlatformUtils.IsWindows() && _win32SoftwareRendering)
                     {
                         appBuilder.With(new Win32PlatformOptions
                             { RenderingMode = new[] { Win32RenderingMode.Software } });
                     }
-#endif
```
derrickstolee, line 75: Keeping framework version.

src/shared/Core/Interop/Windows/WindowsSettings.cs
```diff
@@ -48,9 +49,6 @@ protected internal override bool TryGetExternalDefault(string section, string sc
 
                 return true;
             }
-#else
-            return base.TryGetExternalDefault(section, scope, property, out value);
-#endif
```
derrickstolee, line 53: Keeping framework version.

derrickstolee (commented):
I expected the framework version of the code to be deleted with the removal of the `#if NETFRAMEWORK` macros, but some chunks are keeping the framework version.

I suspect that the project dependencies that changed to match the core version make the framework-specific code now work in the core version, but I'm not able to connect the dots myself.

Thus, I only commented on the places where I saw the code stick instead of a deletion, just as a way for you to confirm that these are all intentianal.

src/shared/Core/Authentication/MicrosoftAuthentication.cs
mjcheetham: We now unconditionally include `Microsoft.Identity.Client.Broker` for all OSes. At runtime we still only use broker on Windows.

src/shared/Core/Authentication/MicrosoftAuthentication.cs
mjcheetham: We now unconditionally set up the MSAL broker for all target frameworks. At runtime we still only use broker on Windows.

src/shared/Core/Authentication/MicrosoftAuthentication.cs
mjcheetham: `IsWindowsBrokerSupported = IsWindows + deeper Windows version checks`

Before this was only run on .NET Framework, so we knew we were on Windows. On .NET Core (Mac/Linux) we always returned `false`.

Since removing this `#if` we'd now be checking `IsDesktopSession` first, even on Mac and Linux when we know that the `IsDesktopSession` check is immaterial.. we only support this on Windows, which is the faster check, hence moving it first in the short-circuit.

src/shared/Core/Authentication/MicrosoftAuthentication.cs
mjcheetham: This is a stale comment. The MSAL team moved to a different model (MSAL Runtime + native broker interop library) since this was written. We will still be using the same broker access model on Windows as before.

src/shared/Core/Authentication/MicrosoftAuthentication.cs
mjcheetham: Yes. There is a larger story behind this, sadly. If we wish to drop .NET Framework but also keep the embedded webview on Windows we'd need to bring along the Chromium-based WebView2 component, which is _huge_.

Instead we are going to look at one of two options:
- Avalonia WebView (which uses the OS-bundled webview component, or the IE fallback on Windows),
- dropping embedded entirely, and rely on Broker.

src/shared/Core/UI/AvaloniaUi.cs
mjcheetham: This was an optimisation case, where we _knew_ we were on .NET Framework _and therefore Windows_, so we didn't need to check this flag on .NET Core.

Now we must always check at runtime.

src/shared/Core/UI/AvaloniaUi.cs
mjcheetham: This was an optimisation case, where we _knew_ we were on .NET Framework _and therefore Windows_, so we didn't need to do `PlatformDetect`ion.. now we always do.

src/shared/Core/PlatformUtils.cs
mjcheetham: This is more about keeping the Windows-only functionality and folding it into a runtime check, rather than compiling it out on .NET Core.

src/shared/Core/Interop/Windows/WindowsSettings.cs
mjcheetham: The Registry access here is Windows-only functionality, but we still need this. Previously we compiled it _out_ for Mac/Linux (.NET Core).

Perhaps it would have been clearer had these sorts of optimisations for Mac/Linux were down as `#if !NETFRAMEWORK` and the blocks swapped.

src/shared/Core/PlatformUtils.cs
mjcheetham: We need this on Windows, and now we cannot compile it _away_ on .NET Core.

mjcheetham:
> I expected the framework version of the code to be deleted with the removal of the #if NETFRAMEWORK macros, but some chunks are keeping the framework version.

I understand the confusion. There are two cases where we had done `#if NETFRAMEWORK` conditional compilation:

1. where we needed to use a different API for .NET Framework, or
2. we wanted to compile-away something that was _Windows-specific_.

Many of these you are rightly confused by are the latter. Since we knew Windows was .NET Framework, and Mac/Linux was .NET Core, we could optimise at compile-time to strip extra code and dependencies from the Core (and therefore Mac/Linux) builds.

Now every OS is using .NET Core, we must keep the Windows-only functionality in all builds, but gated behind a `IsWindows` runtime check.

derrickstolee (approved):

src/shared/Core/Authentication/AuthenticationBase.cs
```diff
@@ -60,11 +60,7 @@ protected internal virtual async Task<IDictionary<string, string>> InvokeHelperA
             // Write the standard input to the process if we have any to write
             if (standardInput is not null)
             {
-#if NETFRAMEWORK
```
dscho, line 63: The commit message says:

> Support for embedded web view flows will be restored at a later date via Avalonia WebView.

Is this still true, or will we move to the regular browser flow (where users log in via a web page in their default browser)?

dscho (approved):

src/shared/Core/Authentication/AuthenticationBase.cs
mjcheetham: My experiments so far have proved it is likely not possible to restore embedded flows via Avalonia+MSAL interop, but I'm yet to call time on that effort yet.
The system webview flow (if broker isn't used ofc, which will be our main target) will be the fallback if AvnWV is a no go.

Frulfump:
Nice getting this merged. When is it roughly expected to be a part of a "beta" and then GA?

PR #2366, opened 2026-06-29
aot: enable ahead-of-time (AOT) compilation

mjcheetham:
## Summary

Make Git Credential Manager build as a Native AOT application, and turn AOT on by default at publish. Getting there means removing the reflection-based APIs AOT can't use — chiefly JSON serialisation — and moving onto an Avalonia version that supports the compiled, trim-safe patterns AOT requires.

## Why

AOT gives a self-contained native binary with no managed startup cost and no runtime dependency, which is a better shape for a credential helper that Git launches on every operation. The blockers are reflection-driven code paths; this branch converts the big one (JSON) to source generation and updates the UI stack so the whole product is AOT/trim-safe.

## What changes

- **Upgrade Avalonia to 12.0.5.** Reacts to 12.x breaking changes: explicit `x:DataType` for the now-default compiled bindings across all views, reworked clipboard and hotkey access, and renamed window-chrome / placeholder-text properties.
- **Enable JSON source generation product-wide.** Define explicit JSON contexts per area (with their serialisation options) so `System.Text.Json` generates serialisers instead of using reflection.
- **Enable JSON source generation for Trace2.** Switch to the built-in lower-snake-case naming policy and set enum string values via attributes.
- **Turn on `PublishAot` by default**, so a published build is a native AOT binary.
- **Drop the explicit AOT flags** from the macOS/Linux layout scripts now that the project sets it.

## Notes

- AOT linking needs a native toolchain (clang, zlib). That's fine for the shipped package; a from-source `--no-aot` path is added in the build-system rework that follows.

Requires PR https://github.com/git-ecosystem/git-credential-manager/pull/2365 to be completed first.


src/shared/Core/UI/Controls/DialogWindow.axaml
```diff
@@ -18,54 +19,58 @@
     <Design.DataContext>
         <vm:WindowViewModel/>
     </Design.DataContext>
-    <Border BorderBrush="{DynamicResource DialogWindowBorderBrush}"
-            BorderThickness="{Binding ShowCustomWindowBorder, Converter={x:Static converters:BoolConvertersEx.ToThickness}}">
```
mjcheetham, line 22: The diff looks large, but we just dropped the wrapping `<Border>` element, and the rest moved 'one tab less' and made some minor renames. Diff ignoring whitespace for an easier time.

src/shared/Core/Core.csproj
```diff
@@ -19,9 +19,4 @@
     <PackageReference Include="Avalonia.Themes.Fluent" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(Configuration)' == 'Debug'">
-    <PackageReference Include="Avalonia.Diagnostics" />
```
dscho, line 23: I guess `Avalonia.Diagnostics` isn't compatible with .NET 10 and/or AOT?

src/linux/Packaging.Linux/install-from-source.sh
```diff
@@ -259,5 +259,9 @@ if [ -z "$DOTNET_ROOT" ]; then
 fi
 
 cd "$toplevel_path"
-$sudo_cmd env "PATH=$PATH" $DOTNET_ROOT/dotnet build ./src/linux/Packaging.Linux/Packaging.Linux.csproj -c Release -p:InstallFromSource=true -p:installPrefix=$installPrefix
+# Build a non-AOT binary so installing from source needs only the .NET SDK,
+# not a C toolchain (clang + zlib) to link a native build. PublishAot is read
+# from the environment by Git-Credential-Manager.csproj and inherited by the
+# nested publish.
+$sudo_cmd env "PATH=$PATH" PublishAot=false $DOTNET_ROOT/dotnet build ./src/linux/Packaging.Linux/Packaging.Linux.csproj -c Release -p:InstallFromSource=true -p:installPrefix=$installPrefix
```
dscho, line 266: In the long run, we will want this to be an option, so that running, say, `install-from-source.sh --use-aot` _would_ draw in the native toolchain (or expect it to be installed already) and then produce a fast executable.

dscho (approved):
I particularly like that using `JsonSerializable` makes the code more robust than it was before!

src/shared/Core/Core.csproj
mjcheetham: They've dropped that package for Avalonia 12.

src/linux/Packaging.Linux/install-from-source.sh
mjcheetham: That's something coming up, when we refactor the build system!

PR #2367, opened 2026-06-29
build: rework the build system and project layout

mjcheetham:
## Summary

A top-to-bottom rework of how the repository is laid out and built. Flattens the `src/shared` tree into `src/`, renames the executable to `git-credential-manager`, and replaces the per-platform packaging projects with a uniform `build/<os>` layout driven by thin scripts over a
shared CLI scripting library. Migrates the solution to SLNX, adopts the .NET 10 artifacts-output layout, refreshes CI and docs, and restores the .NET tool package.

## Why

The `src/{windows,osx,linux}` + `src/shared` split is a relic from when code was platform-partitioned. Every library is cross-platform now, so the structure only adds noise, and the old monolithic `layout.sh` / `pack.sh` packaging was hard to follow and inconsistent across platforms. This branch makes the layout flat and the packaging uniform, and brings the build onto current .NET conventions (SLNX, `UseArtifactsOutput`, NoTargets distribution projects).

## What changes

**Project layout**
- Flatten all libraries and tests up one level into `src/` (pure move — identical assemblies); relocate the scoped `Directory.Build.props`.
- Rename the executable project `Git-Credential-Manager` → `git-credential-manager` so project, directory and output binary all match the command name.
- Move the MSBuild tasks under `build/msbuild` (renamed `MSBuildTasks`).

**Packaging (uniform `build/<os>`)**
- Add shared CLI scripting libraries (`lib-cli.sh`, `lib-cli.psm1`): logging, path/version/config resolution, flag parsing.
- Rework Linux, macOS and Windows packaging into `build/<os>` with a `*.Distribution.csproj` (NoTargets) anchoring each in the solution plus thin `publish` / `pack` / `archive` scripts.
- Bump Inno Setup to 6.7.3; fetch the compiler on demand via a download helper instead of from the NuGet package path.

**Solution & .NET 10 config**
- Migrate to the SLNX solution format (terser, merges cleanly, no GUID bookkeeping); rename the ReSharper DotSettings sidecar to match.
- Adopt `UseArtifactsOutput` (everything under `out/`), default to `net10.0`, bump the SDK to 10.0, register `Microsoft.Build.NoTargets`, normalise `VERSION`.
- Mark the three `*.Distribution.csproj` `<Build Project="false" />` so a plain `dotnet build`/`dotnet test` builds only the product and tests — no accidental publish-and-package side effects. They still build directly via `dotnet build build/<os>`.

**install-from-source**
- Promote the Linux-only script to a cross-platform `build/install-from-source.sh`: detect OS, bootstrap deps, publish via the per-OS script, stage under `<prefix>/share/gcm-core` with a `<prefix>/bin` launcher symlink.
- Default to a trimmed, self-contained **non-AOT** build (needs only the .NET SDK); `--aot` opts into the native toolchain. `sudo` only when the prefix isn't writable. Fixes two latent bugs from the old script (multi-SDK detection, cwd change breaking repo detection).
- Add a `--aot` / `--no-aot` toggle to the publish scripts (defaults to AOT, so the shipped package is unchanged).

**.NET tool**
- Restore `git-credential-manager` .NET tool packaging under `build/dntool` (framework-dependent IL; pack the code-signed layout with `--no-build`).

**CI & docs**
- Point Azure Pipelines and GitHub Actions at the new `build/` entry points and flattened paths; add the ESRP `sign.yml` template; add jobs to build and (on release) sign + publish the .NET tool; validate install-from-source on the new path and on macOS.
- Refresh the development guide for the new layout and outputs; document building each platform's distributables and the .NET tool; note the Homebrew-SDK Native-AOT link caveat.

## Notes

- Despite the size, the change is overwhelmingly **moves, renames and build plumbing**; library contents and the produced binaries are unchanged.
- A solution build no longer triggers packaging — produce distributables with `dotnet build build/<os>`.

## Testing

- `dotnet build git-credential-manager.slnx` (product + tests only)
- Core + platform test suites
- Distribution builds via `dotnet build build/<os>`
- install-from-source validated on Linux matrix + macOS

Requires PR https://github.com/git-ecosystem/git-credential-manager/pull/2366 to be completed first.


build/lib-cli.sh
```diff
@@ -0,0 +1,302 @@
...
+# Artifacts layout
+# ---------------------------------------------------------------------------
+
+# _artifacts_dir <kind> <project> <config> [runtime]
+# Shared implementation for bin_dir/publish_dir. Echo the artifacts directory
+# <root>/out/<kind>/<project>/<pivot>, where <pivot> is <config>, or
+# <config>_<runtime> when a runtime is given (the repository root is located
+# relative to this library). Returns 1 without output if <project> or <config>
+# is missing.
+_artifacts_dir () {
```
dscho, line 96: Not a show-stopper, but I'm curious why this function alone starts with an underscore when all the others do not?

Directory.Packages.props
```diff
@@ -22,7 +22,7 @@
     <!-- Other -->
     <PackageVersion Include="System.CommandLine" Version="2.0.0-beta4.22272.1" />
     <PackageVersion Include="System.Text.Json" Version="8.0.5" />
-    <PackageVersion Include="Tools.InnoSetup" Version="6.3.1" GeneratePathProperty="true" />
+    <PackageVersion Include="Tools.InnoSetup" Version="6.7.3" />
```
dscho, line 25: I remember that last time we bumped this, we had to be careful to bump only to a version that was available in a certain feed. We may need to prepare for that by requesting that version to be in that feed.

docs/development.md
```diff
@@ -11,62 +11,106 @@ installed from the [.NET website][dotnet-web].
 
 ## Building
 
-The `Git-Credential-Manager.sln` solution can be opened and built in Visual
-Studio, Visual Studio for Mac, Visual Studio Code, or JetBrains Rider.
```
dscho, line 15: 😢

build/linux/publish.sh
```diff
@@ -41,6 +44,7 @@ RUNTIME=""
 OUTPUT=""
 SYMBOL_OUTPUT=""
 VERSION=""
+AOT="true"
```
dscho, line 47: @mjcheetham Ah, this is what you were referring to in https://github.com/git-ecosystem/git-credential-manager/pull/2366#discussion_r3530091132! I can see the rationale of switching the default from (implicit) ` --no-aot` to `--aot`, even if I would have preferred that change to be made in a separate commit. Just a nit, though.

dscho (approved):
Wow, what a big PR! I was scared of 27 commits that change 488 files, and while it took a while to review, the PR is well-structured, the commits tell a compelling story, and I joyfully approve!

build/linux/publish.sh
mjcheetham: The .NET tool is build as framework-dependent / not AOT, so it was a natural step to make install from source also not AOT by default. The AOT toolchain is OS and CPU arch specific, and since Linux has 50 million different combinations of everything it's very hard to do this.

To upside is that the non-AOT builds (like the .NET tool) will run on _many_ architectures a libc (like muslc) flavours!

Image

build/lib-cli.sh
mjcheetham: The rest of the functions are used externally. The `_` function(s) are only used internally, inside this library script. Do shell scripts support the concept of function being exported like pwsh modules?

docs/development.md
mjcheetham: Is this about the dropping of VS Mac support? Blame Microsoft!

build/lib-cli.sh
dscho: Makes sense!

docs/development.md
dscho: Yes, it was. And I am still sad about it.

PR #2376, opened 2026-07-13
windows: ensure app manifest version is formatted correctly

mjcheetham:
Windows requires the version property in an application manifest to be of the format "x.y.z.w" which specific rules about the min/max values for each component (16-bit numbers).

Use `System.Version` to parse the version and normalise/validate before writing the manifest file.


dscho (approved):
Looks good to me!

PR #2377, opened 2026-07-13
console: replace terminal UI with Spectre.Console

mjcheetham:
## Summary

Replace GCM's line-oriented `ITerminal` / `TerminalMenu` stack with a Spectre.Console-backed console layer. Interactive prompts now use the controlling terminal, while diagnostics and other output-only messages continue to use standard error. Git's credential-protocol stdin and stdout remain untouched.

The branch adds the cross-platform TTY plumbing needed to make that split work on macOS, Linux, and Windows, then migrates every existing terminal prompt and removes the old abstraction. This is deliberately the console foundation for richer account-picker flows; it does not add the follow-on account-selection behaviour itself.

## Why

The existing terminal surface supports basic line output, text/secret input, and numbered one-shot menus. That is enough for today's prompts, but not for the multi-choice, filter-as-you-type, confirmation, and progress experiences needed by the account-picker work.

Using Spectre.Console directly is not safe for a credential helper because its default console uses process stdin/stdout, which Git reserves for the credential protocol. This change introduces an explicit routing layer:

- Output-only messages render to stderr, where they remain visible and capturable even without a terminal.
- Prompts use the controlling TTY through `/dev/tty` on POSIX and `CONIN$` / `CONOUT$` on Windows.

## What changes

**Console architecture**

- Add Spectre.Console 0.57.0 and expose a single `IConsoleService Console` on `ICommandContext`.
- Back the service with two `IAnsiConsole` instances: an output-only stderr console and an interactive controlling-TTY console.
- Preserve plain redirected output by disabling ANSI colour when stderr is redirected.
- Fall back to a non-interactive, null-input console when no controlling TTY can be opened rather than ever borrowing Git's standard streams.
- Add a Spectre-backed test console so prompt tests feed real text, secret, arrow-key, and Enter input through the same prompt implementations used in production.

**Native TTY support**

- Add POSIX output and input adapters over `/dev/tty`. The input side uses `termios` raw mode plus a platform-neutral ANSI escape parser for navigation, function, modifier, control, and UTF-8 key sequences.
- Use `select(2)` for key availability and the short standalone-Escape disambiguation window. Unlike `poll(2)`, this works for TTY character devices on both macOS and Linux.
- Add Windows output and input adapters over `CONOUT$` and `CONIN$`. Output enables virtual-terminal processing where available; input translates `ReadConsoleInput` key events into `ConsoleKeyInfo`.
- Scope raw mode to an active read or prompt instead of the process lifetime. A prompt-level hold prevents fast typeahead from being echoed between reads during masked input, while returning to cooked mode between prompts keeps normal Ctrl+C handling working during network, broker, and GUI operations.
- Restore terminal state on normal completion, exception unwinding, and relevant POSIX signals. Ctrl+C read during a prompt becomes an `InterruptedException`, which exits quietly with the conventional status code 130 after the terminal has been restored.

**Prompt and diagnostic migration**

- Move username, password/token, two-factor, authentication-mode, Windows integrated-authentication, OAuth, Microsoft-authentication, and existing account-selection prompts to Spectre.Console.
- Replace numbered text menus with interactive selection prompts while preserving the available choices, ordering, and default selection.
- Route warnings, errors, informational messages, device-code text, and fatal exception output through `IConsoleService`, giving them one consistent stderr-backed surface.
- Delete `ITerminal`, the POSIX/macOS/Linux/Windows terminal implementations, `TerminalMenu`, and the old test double once their final consumers move.
- Include the process ID in the wait-for-debugger message so the correct GCM process is unambiguous when several helpers are running.

## User-visible behaviour and compatibility

- The Git credential wire protocol is unchanged: stdin and stdout are still used only for communication with Git.
- Terminal selection prompts now use arrow-key navigation instead of asking the user to enter a menu number. Authentication choices and ordering are unchanged.
- Human-facing prefixes are coloured when stderr is attached to a terminal and remain plain text when redirected.
- Ctrl+C during an interactive prompt restores the terminal and exits with status 130 without printing a misleading `fatal:` message.
- Existing GUI prompt paths and authentication decisions are unchanged.
- In headless environments, interactive input degrades to the explicit non-interactive console rather than reading from redirected protocol input.

## Testing

- `dotnet test git-credential-manager.slnx --configuration Debug`

New coverage includes the ANSI escape parser, POSIX `select(2)` semantics, Windows key-event translation, raw-mode prompt lifetime and cleanup, headless console contracts, stderr routing, interrupt detection, and the migrated provider prompt flows.

## Notes for reviewers

- The eight commits are arranged as a progression: console facade, TTY output, POSIX input, Windows input, contract tests, the debugger-message tweak, diagnostic routing, and finally the atomic prompt migration/removal of `ITerminal`. Reviewing commit-by-commit is recommended.
- The highest-risk code is isolated in the native input adapters. The key invariant is that raw mode is reference-counted and held only while a read or prompt owns it; every completion and interruption path must restore the original terminal mode.
- The account-picker work can build on this branch without reopening the credential-protocol or platform-TTY layers.


src/Core/Interop/Windows/WindowsAnsiConsoleInput.cs
```diff
@@ -0,0 +1,268 @@
...
+        }
+
+        _inRawMode = false;
+    }
+
+    private void AcquireRawMode()
+    {
+        lock (_modeLock)
+        {
+            if (_rawDepth++ == 0)
```
dscho, line 132: Is there an atomic increment in .NET? If so, it would probably make sense to use that instead, it's typically cheaper than a full-blown lock.

src/Core/Commands/GetCommand.cs
```diff
@@ -16,7 +18,16 @@ public GetCommand(ICommandContext context, IHostProviderRegistry hostProviderReg
...
         {
-            GitResponse response = await provider.GetCredentialAsync(request);
+            GitResponse response;
+            try
+            {
+                response = await provider.GetCredentialAsync(request);
+            }
+            catch (Exception ex) when (ex is OperationCanceledException || ex is InterruptedException)
+            {
+                response = GitResponse.Cancel();
```
dscho, line 28: Don't we need to re-throw the exception, then?

src/Atlassian.Bitbucket.Tests/BitbucketAuthenticationTest.cs
```diff
@@ -16,8 +16,7 @@ public class BitbucketAuthenticationTest
         public async Task BitbucketAuthentication_GetCredentialsAsync_Basic_SucceedsAfterUserInput(string username, string password)
         {
             var context = new TestCommandContext();
-            context.Terminal.Prompts["Username"] = username;
-            context.Terminal.SecretPrompts["Password"] = password;
+            context.Console.PushText(password);
```
dscho, line 19: Hmm. I do not see where the username is set after this... How is this test passing?

src/Core/Authentication/MicrosoftAuthentication.cs
```diff
@@ -477,7 +471,7 @@ internal MicrosoftAuthenticationFlowType GetFlowType()
                         break;
                 }
 
-                Context.Streams.Error.WriteLine($"warning: unknown Microsoft Authentication flow type '{valueStr}'; using 'auto'");
+                Context.Console.WriteWarning($"unknown Microsoft Authentication flow type '{valueStr}'; using 'auto'");
```
dscho, line 474: Technically this should have been part of a previous commit ("console: route diagnostics to the stderr console"), I guess? Just a nit, I won't insist on any change.

dscho (approved):
Wow, what a _lot_  of work! I have just a couple of remarks/questions below, nothing Earth-shattering :-)

src/Core/Commands/GetCommand.cs
mjcheetham: No. We only hit this block if the exception is `OperationCanceledException` or `InterruptedException`, which means something was cancelled (user dismissed a dialog, pressed Ctrl+C whilst we had been in raw mode).

In this case we want to respond to Git with the 'cancel' response, which is to emit `quit=1` which tells Git to stop.

src/Atlassian.Bitbucket.Tests/BitbucketAuthenticationTest.cs
mjcheetham: This is because we're passing in a username `jsquire` to the `GetCredentialsAsync` call, meaning we don't prompt for the username and go right to prompting for the password.

The previous test set up responses to specific prompts, and the "Username" prompt response was just never encountered.

src/Core/Interop/Windows/WindowsAnsiConsoleInput.cs
mjcheetham: .NET has `Interlocked.Increment`, but it would only make the counter atomic. The lock also serialises the `_rawDepth` transitions with `EnterRawMode()` and `RestoreMode()`. Without it, another acquire could return before raw mode is entered, or race a final release. Avoiding the lock would require additional synchronisation.

mjcheetham: The `lock` keyword is also syntactic sugar for using `Monitor` (a critical section):

```csharp
lock (lockObj)
{
    DoWork();
}
```

.. is the same as ..

```csharp
bool lockTaken = false;

try
{
    Monitor.Enter(lockObj, ref lockTaken);
    DoWork();
}
finally
{
    if (lockTaken)
    {
        Monitor.Exit(lockObj);
    }
}
```

`Monitor.Enter` blocks until the monitor is acquired; the `finally` ensures it is always released. `lock` therefore also blocks if another thread holds the lock already.

https://learn.microsoft.com/en-us/dotnet/api/system.threading.monitor

src/Core/Authentication/MicrosoftAuthentication.cs
mjcheetham: Fixed!

src/Core/Commands/GetCommand.cs
dscho: Ah, okay!

src/Atlassian.Bitbucket.Tests/BitbucketAuthenticationTest.cs
dscho: Makes sense to me now!

src/Core/Interop/Windows/WindowsAnsiConsoleInput.cs
dscho: Thanks for the explanations!

PR #2378, opened 2026-07-13
build: fix AOT and IDE integration

mjcheetham:
## Summary

Address follow-up issues from the Native AOT and build-system changes. Distribution builds remain AOT by default, while normal and design-time builds produce IL that IDE tooling such as the Avalonia previewer can load.

## What changes

- Move `PublishAot` from the main project into the platform publish scripts.
- Retain AOT analyzer coverage with `IsAotCompatible`.
- Remove an AOT warning by using generic `Marshal.SizeOf<T>()`.
- Add shared build scripts to the solution.
- Correct the Avalonia code-behind filename for IDE nesting.

There is no intended runtime behavior change.

## Testing

- `dotnet build git-credential-manager.slnx`
- macOS AOT publish
- macOS non-AOT publish

Follow-up to #2366 and #2367.


dscho (approved):
Great!

PR #2379, opened 2026-07-13
entra: evolve Microsoft authentication

mjcheetham:
## Summary

This series evolves the existing Microsoft authentication component into the reusable Microsoft Entra authentication API used by Azure Repos.

Rather than constructing a second implementation, migrating the caller, and deleting the original, this history renames and reorganizes the existing component first. It then changes its API and behavior incrementally, updating callers and tests with each breaking change.

## Why

The existing implementation already contains the core public-client and confidential-client authentication flows. Treating the replacement as a separate component obscures that continuity and leaves reviewers comparing two large implementations before a final migration and deletion.

This history instead makes the architectural changes explicit:

- adopt Microsoft Entra naming in place;
- establish a dedicated namespace and file layout;
- split public-client, confidential-client, and cache responsibilities;
- replace string-shaped identities with stable account and managed identity types; and
- evolve configuration, interaction, broker, and cache behavior in dependency order.

The result is the same implementation as `entraauth`, but each commit describes how the existing component becomes that implementation.

## What changes

- Rename `MicrosoftAuthentication` and its public contracts for Microsoft Entra.
- Move the component into `Authentication/Entra` and extract its public contracts and supporting identity types.
- Split the implementation into core, public-client, confidential-client, and token-cache partial-class files.
- Add cached account enumeration and removal, return stable accounts from authentication, and authenticate a selected account.
- Update Azure Repos alongside each API change instead of through a final migration commit.
- Model managed identities explicitly and simplify the confidential authentication APIs.
- Make the default-account setting tri-state.
- Preserve escaping when handing browser URLs to platform launchers.
- Expose interaction-mode selection and cancellation.
- Move application settings into `PublicClientConfig`.
- Add per-application broker capabilities and update the Microsoft identity packages required by them.
- Make the shared Microsoft developer token cache opt-in while preserving Azure Repos compatibility.

## Review guide

1. **Rename:** adopt the Entra names without changing behavior.
2. **Organize:** move the types into their dedicated namespace and files.
3. **Split:** separate public-client, confidential-client, and cache responsibilities while preserving the old API.
4. **Evolve the contracts:** introduce stable account and managed identity types, then update each API and its Azure Repos caller.
5. **Evolve user authentication:** add tri-state default-account handling, preserve browser URLs, and expose interaction modes.
6. **Finish application integration:** centralize public-client configuration, add broker capabilities, and make shared-cache use explicit.

There is no parallel implementation and no final migration-and-delete step: the production component and its caller remain usable throughout the series.

## Validation

- `dotnet build src/git-credential-manager/git-credential-manager.csproj`


dscho (approved):
This was the hardest PR yet, in this `vnext` endeavor. I had to use `--color-moved --color-moved-ws=allow-indentation-change` liberally, and still there was so much that at times, I had to hide the removed lines and only focus on the added lines to make sense of them. I reviewed the commits to the best of my abilities, and they look good to me! If there is something both you, Copilot _and_ me have overlooked, I am quite confident that it won't be critical. So: good to go from my side!

mjcheetham:
> I had to hide the removed lines and only focus on the added lines to make sense of them

Yeah, it's better to view this as a rewrite of the existing `MicrosoftAuthentication` and have the old and new (`EntraAuthentication`) side by side to match up the logically identical pieces, and see where structurally they differ.

Thanks for the review!

PR #2380, opened 2026-07-14
windows: use Inno Setup 7.0.2 and x64 native installer

mjcheetham:
Update to Inno Setup 7.0.2, and use the new 64-bit installer for win-x64 and win-arm64 targets, falling back to the classic 32-bit installer for win-x86 targets.

https://jrsoftware.org/ishelp/topic_setup_setuparchitecture.htm
https://jrsoftware.org/ishelp/topic_64bit.htm


dscho (approved):

mjcheetham:
Confirmed in task manager for x86 vs x64 (cannot easily test ARM64)

image

PR #2381, opened 2026-07-14
entra: ensure we only configure broker when app supports it

mjcheetham:
Only correctly configure the broker when the application config states that it is configured correctly for the broker.

Without this check it is possible that we configure a public client application that has the broker enabled, but the application config does NOT have the correct redirect URI configured.

Easiest to review without whitespace changes highlighted.


dscho (approved):
With `git show -w`, this commit looks obviously correct to me.

PR #2408, opened 2026-08-13
trace2: improve lifecycle and instrumentation

mjcheetham:
GCM already emits Trace2 events, but the implementation is owned by `CommandContext` even though Trace2 represents a single process-wide event stream. This prevents tracing from starting before context construction and requires the tracer to be passed through services that cannot meaningfully replace it.

The existing implementation also attributes asynchronous work to physical managed threads, reports some errors through tracing-specific exception types, and leaves major command, UI, and authentication boundaries uninstrumented.

This PR:

- reorganises and simplifies the Trace2 message and writer code;
- initialises one process-wide Trace2 session at program entry;
- removes tracer dependencies from contexts, services, and OAuth clients;
- centralises exception reporting at the application boundary;
- gives child-process events stable correlation IDs and accurate timing;
- tracks logical execution contexts across asynchronous continuations;
- adds command-name, scalar-data, and structured-data events;
- instruments command dispatch, Git configuration, credential storage, UI startup, and authentication flows.

Trace2 continues to honour the existing Git environment variables and configuration settings, and supports event, normal, and performance output formats. The added instrumentation records control-flow and configuration choices without recording credentials.

The series is organised as follows:

- commits 1-2 reorganise and refactor the tracing implementation;
- commits 3-8 migrate Trace2 to process-wide ownership;
- commits 9-14 improve event lifecycle, context, and event types;
- commits 15-18 add focused instrumentation to application workflows.

To keep the cross-cutting ownership migration reviewable, commits 3-7 are intentionally not independently buildable. Commit 8 completes the migration and restores a buildable tree. The remaining commits build normally.


src/Core/Tracing/Trace2.cs
```diff
@@ -16,203 +16,72 @@ public class Trace2Settings
...
+        Trace2.CompleteRegion(_nesting);
     }
 }
 
 /// <summary>
-/// Represents the application's TRACE2 tracing system.
+/// The application's process-wide TRACE2 tracing system.
 /// </summary>
-public interface ITrace2 : IDisposable
+public class Trace2 : DisposableObject
```
dscho, line 64: Gotta say: This diff wasn't helpful (it's not like it modified `ITrace2` to become `Trace2`, it's more like a complete deletion of that interface and an addition of the class).

src/Core/Git.cs
```diff
@@ -97,23 +94,18 @@ public GitVersion Version
...
-                        }
-                        else
-                        {
-                            _version = new GitVersion();
-                        }
                     }
+                    Match match = Regex.Match(data, @"^git version (?'value'.*)");
+                    _version = match.Success
+                        ? new GitVersion(match.Groups["value"].Value)
+                        : new GitVersion();
```
dscho, line 108: Technically, this refactoring has nothing to do with the concern "trace2: decouple core services". But since I don't want to review another iteration of such a large PR, I don't want you to split out that refactoring into its own commit 😆

src/Core/Tracing/Trace2.cs
```diff
@@ -19,88 +22,71 @@ internal class Trace2Settings
...
-        double relativeTime = (DateTimeOffset.UtcNow - _startTime).TotalSeconds;
-        Trace2.WriteRegionLeave(
-            relativeTime, _category, _label, _message, _thread, _nesting, _filePath, _lineNumber);
-        Trace2.CompleteRegion(_nesting);
-    }
-}
-
 /// <summary>
-/// The application's process-wide TRACE2 tracing system.
+/// The application's process-wide Trace2 tracing system.
```
dscho, line 26: Heh. It's really not an acronym. But fixing that in a comment is _technically_ unrelated to the concern "trace2: track logical thread context"...

Having said that, I really don't want to review another iteration of this large PR, so it's fine with me to leave this change in the unrelated commit ;-)

src/Core/CredentialStore.cs
```diff
@@ -75,11 +79,17 @@ private void EnsureBackingStore()
...
-                                ?? GetDefaultStore();
+            string credStoreName = _context.Settings.CredentialBackingStore?.ToLowerInvariant();
+            string defaultStore = GetDefaultStore();
+
+            Trace2.WriteData("cred_store", "store/configured", credStoreName);
+            Trace2.WriteData("cred_store", "store/default", defaultStore);
+            Trace2.WriteData("cred_store", "store/namespace", ns);
 
-            switch (credStoreName)
+            switch (credStoreName ?? defaultStore)
```
dscho, line 92: I'm unsure whether that change was intended? It does not strike me as increasing Trace2 coverage, but as changing behavior (even if I suspect that falling back to `defaultStore` is an improvement over the previous logic?).

dscho (commented):
Oh my, what a big PR!

But it's worth it. I'm particularly excited about proper tracing of async call paths, and _especially_ about a much more complete Trace2 coverage!

dscho (approved):
Oops, I meant to approve the PR 😊

src/Core/CredentialStore.cs
mjcheetham: A slight behaviour change yes, since we now always call `GetDefaultStore()` above, regardless if `_context.Settings.CredentialBackingStore` is `null` or not.. but I had to split it so we can log the values of both the default and selected stores to Trace2.

src/Core/Tracing/Trace2.cs
mjcheetham: Yeah. I was going around updating shouty `TRACE2` to `Trace2` as we I was going.. `TRACE2` is the environment variable name :)

src/Core/Git.cs
mjcheetham: This may be a product of the LLM doing their usual "let's rewrite this anyway". Sorry!

PR #2413, opened 2026-08-18
msauth: isolate macOS user token cache

### LINKED ISSUE #2079: Constantly being asked to authenticate keychain on MacOS against Azure Devops
### Version

2.6.1+786ab03440ddc82e807a97c0e540f5247e44cec6

### Operating system

macOS

### OS version or distribution

26.0.1 (25A362)

### Git hosting provider(s)

Azure DevOps

### Other hosting provider

_No response_

### (Azure DevOps only) What format is your remote URL?

https://{org}.visualstudio.com

### Can you access the remote repository directly in the browser?

Yes, I can access the repository

### Expected behavior

Authentication at a reasonable frequency (no more often than once a day)

### Actual behavior

My issue is the same as #435, which was closed due to inactivity.

For reference, I am a data scientist working at Microsoft.

I have to authenticate ~once an hour when I do git pull or push, which includes having to type my password for keychain. "Always Allow" does not help.

Image

The other thread asked for the result of `security find-generic-password -gs git:dev.azure.com/XXXXXXXXhq` where the last bit is the url to the ADO org. Ours is e.g. `https://orgX.visualstudio.com`.

```
❯ security find-generic-password -gs git:https://<redacted>.visualstudio.com                                                                            (notesUtilityMetrics-3.12-2025-10-16)
keychain: "/Users/thomas/Library/Keychains/login.keychain-db"
version: 512
class: "genp"
attributes:
    0x00000007 <blob>="git:https://<redacted>.visualstudio.com"
    0x00000008 <blob>=<NULL>
    "acct"<blob>="<redacted>@microsoft.com"
    "cdat"<timedate>=0x32303235303931373135323834345A00  "20250917152844Z\000"
    "crtr"<uint32>="aapl"
    "cusi"<sint32>=<NULL>
    "desc"<blob>=<NULL>
    "gena"<blob>=<NULL>
    "icmt"<blob>=<NULL>
    "invi"<sint32>=<NULL>
    "mdat"<timedate>=0x32303235303931373135323834345A00  "20250917152844Z\000"
    "nega"<sint32>=<NULL>
    "prot"<blob>=<NULL>
    "scrp"<sint32>=<NULL>
    "svce"<blob>="git:https://<redacted>.visualstudio.com"
    "type"<uint32>=<NULL>
password: "<redacted>"
```

### Logs

I'd prefer to avoid pasting sensitive logs, but can provide specific things on-demand.

thomasaarholt:
## The problem
On macOS, GCM shares the `Microsoft.Developer.IdentityService` MSAL Keychain item with other Microsoft developer tools. When another tool recreates this item, its access-control list can lose GCM's entry, causing repeated Keychain password prompts even after selecting **Always Allow**.

## The solution
Use a GCM-owned Keychain item and cache file for the macOS user-token cache. This prevents other applications from changing GCM's Keychain permissions while retaining persistent authentication between GCM processes.

Windows and Linux continue using the shared Microsoft developer cache. macOS users may need to authenticate once after upgrading to populate the new cache.

Fixes #2079.

## Testing
- `dotnet test src/shared/Core.Tests/Core.Tests.csproj --no-restore --nologo`
- `dotnet build -c MacDebug --no-restore --nologo`
- Installed the generated arm64 package and verified the installed binary version and checksum.


mjcheetham:
Hello @thomasaarholt! Thanks for you pull request here.

Rather than just always opting out GCM on Mac from using the shared Microsoft developer tools MSAL cache, we would rather prefer to make this an option that users could configure, with the default still being 'true'.

If we could figure out what was happening with the Keychain ACLs that is resulting in these prompts that would be the preferred option. Until then, disabling the cache integration via option is reasonable.

Please see this PR that should hopefully provide you with the required option to avoid this issue on macOS: https://github.com/git-ecosystem/git-credential-manager/pull/2415

thomasaarholt:
Awesome! Thanks, I'll take a look at the PR!

PR #2415, opened 2026-08-24
azrepos: make shared token cache configurable

mjcheetham:
Azure Repos OAuth tokens currently always use the shared Microsoft developer tooling cache. While this enables authentication reuse across GCM, Visual Studio, and Azure CLI, users have no way to isolate GCM's authentication state.

Allow users to select a GCM-specific token cache through Git configuration or an environment variable. Keep the shared cache as the default to preserve existing behaviour and document the interoperability and opt-out semantics.


src/Microsoft.AzureRepos/AzureReposHostProvider.cs
```diff
@@ -406,14 +406,30 @@ private PublicClientConfig GetEntraConfig()
             {
                 ClientId = GetClientId(),
                 IsMsaPassthroughEnabled = true,
-                UseSharedCache = true,
+                UseSharedCache = GetUseSharedCache(),
```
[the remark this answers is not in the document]
mjcheetham: Added tests.

docs/environment.md
```diff
@@ -920,6 +920,33 @@ export GCM_MSAUTH_USEDEFAULTACCOUNT="false"
...
+allows those tools to reuse authentication performed by GCM.
+
+Set this value to `false` to use a GCM-specific token cache instead.
+
+Defaults to `true`.
+
+#### Windows
+
+```batch
+SET GCM_AZREPOS_USE_MSFT_CACHE="false"
```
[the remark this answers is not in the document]
mjcheetham: Fixed this instance. Also created a new PR to fix up the other existing instances in this file: https://github.com/git-ecosystem/git-credential-manager/pull/2420

src/Core/Diagnostics/EntraAuthenticationDiagnostic.cs
```diff
@@ -15,48 +15,73 @@ public EntraAuthenticationDiagnostic(ICommandContext context)
...
-            });
+            var failures = new List<Exception>();
 
-            log.Append("Gathering MSAL token cache data...");
-            StorageCreationProperties cacheProps = entraAuth.CreateUserTokenCacheProps(true);
-            log.AppendLine(" OK");
-            log.AppendLine($"CacheDirectory: {cacheProps.CacheDirectory}");
-            log.AppendLine($"CacheFileName: {cacheProps.CacheFileName}");
-            log.AppendLine($"CacheFilePath: {cacheProps.CacheFilePath}");
+            await RunCacheDiagnosticAsync("Shared Microsoft developer tools", true, log, failures);
```
mr-otmarg, line 20: The diagnostic now tests both caches and throws if either fails. Consequently, a user who explicitly selects the GCM-specific cache can still get a failed git-credential-manager diagnose result solely because the unused shared cache is inaccessible, and vice versa. This is especially misleading because DiagnoseCommand reports that failures indicate an installation problem.

Consider logging results for both caches but making only the configured cache determine diagnostic success, or registering each cache check as a separate diagnostic so the relevant and optional failures are distinguishable.

docs/configuration.md
```diff
@@ -827,6 +827,27 @@ Credential: "git:https://bob@github.com/example/myrepo" (user = bob)
 
 ---
 
+### credential.azreposUseMicrosoftSharedCache
+
+Use the token cache shared by Microsoft developer tools when authenticating to
+Azure Repos with Microsoft identity OAuth tokens. This allows GCM to reuse
+authentication performed by tools such as Visual Studio and Azure CLI, and
+allows those tools to reuse authentication performed by GCM.
```
mr-otmarg, line 835: These additions say the selected cache allows authentication reuse with Azure CLI. Azure CLI maintains its own cache under $AZURE_CONFIG_DIR/msal_token_cache.{bin,json}, rather than the .IdentityService/msal.cache cache configured here. Any broker-based SSO on supported systems is separate from sharing this cache and may still occur when this option is disabled.

Remove Azure CLI from these examples or describe the narrower set of tools that actually consume the Microsoft developer shared cache.

docs/environment.md
```diff
@@ -920,6 +920,33 @@ export GCM_MSAUTH_USEDEFAULTACCOUNT="false"
 
 ---
 
+### GCM_AZREPOS_USE_MSFT_CACHE
```
mr-otmarg, line 923: Same comment seems to apply here from https://github.com/git-ecosystem/git-credential-manager/pull/2415/changes#r3870142015

src/Core/Diagnostics/EntraAuthenticationDiagnostic.cs
mjcheetham: I've rebased on the new diagnostic infrastructure that now reports only a warning should one of the shared or isolated caches fail the test.

docs/configuration.md
mjcheetham: You are correct. I really mean to say the Azure PowerShell CLI (not the Python-based `az` CLI).. unless they also moved away from that too since it was first added?

mr-otmarg: Azure PowerShell did use that cache previously, but Az.Accounts 2.15.1 renamed its cache to msal.cache.cae / msal.cache.nocae. It only migrates msal.cache when the suffixed cache does not already exist, so current versions do not continuously share authentication state with GCM’s msal.cache. WAM may still provide SSO on Windows, but that is broker-based and independent of this setting.

mjcheetham: Makes sense. I'll drop references to the Azure Pwsh CLI and keep just Visual Studio.

mjcheetham: Done. We now only reference "Visual Studio" as participating with the shared cache.

mr-otmarg (approved):

dscho (approved):
Looks great!

PR #2416, opened 2026-08-24
azrepos: use GCM's own Entra application

### LINKED ISSUE #47: Create new client ID for GCM when speaking to Azure Repos

mjcheetham:
This PR has been a long time in the making! We are giving Git Credential Manager its own first-party Microsoft Entra client application for Azure Repos!

GCM previously authenticated as the Visual Studio application, whose registration could not safely be extended for all of GCM's broker and Microsoft account requirements.

The dedicated application enables broker authentication on Windows, macOS, and Linux. In particular, this avoids unnecessary browser prompts on platforms where the broker can provide an existing signed-in account.

## Changes

- Use GCM's dedicated Entra client application for Azure Repos OAuth.
- Enable broker support for the new application on macOS and Linux, in addition to the existing Windows support. (Note that broker support is still opt-in at this time!)
- Preserve the Visual Studio client as a documented compatibility option through `credential.azreposUseLegacyClientId` and `GCM_AZREPOS_USE_LEGACY_CLIENTID`.
- If user authentication fails, then print a message informing the user they can switch back to the legacy client if MSAL, and asks them to report the failure.
- Replace the unrestricted developer client-ID and authority overrides with the focused legacy-client compatibility setting.

## Compatibility

The new GCM application is the default. Users who need the previous behaviour can select the legacy client explicitly. An MSAL failure in the new application triggers an advice message to retry with the old client when obtaining a user Entra token, either directly for OAuth credentials or as part of PAT creation.

Managed identity, workload federation, and service-principal flows are unaffected.

Fixes git-ecosystem/git-credential-manager#47

### DIFF
--- a/src/Core/Authentication/Entra/EntraAuthentication.PublicClient.cs
+++ b/src/Core/Authentication/Entra/EntraAuthentication.PublicClient.cs
@@ -68,7 +68,8 @@ public record PublicClientConfig
 public partial class EntraAuthentication
 {
     private const string MacBrokerRedirectUrl = "msauth.com.msauth.unsignedapp://auth";
-    private readonly PublicClientConfig _publicClientConfig;
+
+    public PublicClientConfig PublicClientConfig { get; }
 
     public async Task<InteractionMode> GetInteractionModeAsync(CancellationToken ct = default)
     {
@@ -219,7 +220,7 @@ public partial class EntraAuthentication
         try
         {
             return await app.AcquireTokenSilent(scopes, msalAccount)
-                .WithMsaPassthroughTransfer(_publicClientConfig.IsMsaPassthroughEnabled, msalAccount)
+                .WithMsaPassthroughTransfer(PublicClientConfig.IsMsaPassthroughEnabled, msalAccount)
                 .ExecuteAsync(ct);
         }
         catch (MsalUiRequiredException)
@@ -439,7 +440,7 @@ public partial class EntraAuthentication
     /// <param name="useBroker">True if the broker will be used for this applications build using this builder.</param>
     private PublicClientApplicationBuilder GetPublicAppBuilder(out bool useBroker)
     {
-        if (_publicClientConfig is null)
+        if (PublicClientConfig is null)
         {
             throw new InvalidOperationException(
                 "Public client configuration is required for user authentication.");
@@ -448,7 +449,7 @@ public partial class EntraAuthentication
         if (_publicBuilder is null)
         {
             Context.Trace.WriteLine("Creating public client application builder...");
-            var builder = PublicClientApplicationBuilder.Create(_publicClientConfig.ClientId)
+            var builder = PublicClientApplicationBuilder.Create(PublicClientConfig.ClientId)
                 .WithHttpClientFactory(_httpFactory)
                 .WithTraceLogging(Context)
                 .WithLegacyCacheCompatibility(false)
@@ -459,9 +460,9 @@ public partial class EntraAuthentication
             if (Context.SessionManager.IsDesktopSession && IsBrokerEnabled())
             {
                 // Check that the app config supports the broker on this platform
-                if (_publicClientConfig.SupportsWindowsBroker && PlatformUtils.IsWindows() ||
-                    _publicClientConfig.SupportsMacBroker && PlatformUtils.IsMacOS() ||
-                    _publicClientConfig.SupportsLinuxBroker && PlatformUtils.IsLinux())
+                if (PublicClientConfig.SupportsWindowsBroker && PlatformUtils.IsWindows() ||
+                    PublicClientConfig.SupportsMacBroker && PlatformUtils.IsMacOS() ||
+                    PublicClientConfig.SupportsLinuxBroker && PlatformUtils.IsLinux())
                 {
                     Context.Trace.WriteLine("Broker is supported by the app and enabled by the user.");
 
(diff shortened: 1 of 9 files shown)


src/Microsoft.AzureRepos/AzureReposHostProvider.cs
```diff
@@ -116,6 +119,31 @@ public async Task<GitResponse> GetCredentialAsync(GitRequest request)
...
             }
 
+            try
+            {
+                return await GetUserCredentialAsync(request);
+            }
+            // If we are using the new client ID, and we get an MSAL exception, let's retry using the
+            // legacy client ID and log a warning; asking the user to report the problem.
+            catch (MsalException msalEx) when (_entraAuth.Value.PublicClientConfig.ClientId !=
+                                               AzureDevOpsConstants.LegacyClientId)
```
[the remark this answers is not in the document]
mjcheetham: Done.

src/Microsoft.AzureRepos/AzureReposHostProvider.cs
```diff
@@ -116,6 +119,31 @@ public async Task<GitResponse> GetCredentialAsync(GitRequest request)
...
+            {
+                _context.Console.WriteError($"Failed to acquire Entra access token: {msalEx.Message}");
+                _context.Console.WriteInfo($"Please report this failure at {Constants.HelpUrls.GcmNewIssue}");
+                _context.Console.WriteInfo("Trying again with the legacy Entra client app...");
+
+                // Recreate the lazy entra auth component forcing the legacy client ID
+                _forceLegacyClientId = true;
+                _entraAuth = new Lazy<IEntraAuthentication>(
+                    () => new EntraAuthentication(_context, GetEntraConfig()));
```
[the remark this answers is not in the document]
mjcheetham: Added tests.

src/Microsoft.AzureRepos/AzureReposHostProvider.cs
```diff
@@ -116,6 +123,30 @@ public async Task<GitResponse> GetCredentialAsync(GitRequest request)
...
+                _context.Console.WriteError($"Failed to acquire Entra access token: {msalEx.Message}");
+                _context.Console.WriteInfo($"Please report this failure at {Constants.HelpUrls.GcmNewIssue}");
+                _context.Console.WriteInfo("Trying again with the legacy Entra client app...");
+
+                // Recreate the lazy entra auth component forcing the legacy client ID
+                ResetEntraAuth(forceLegacyClientId: true);
+
+                // Try again!
+                return await GetUserCredentialAsync(request);
+            }
```
mr-otmarg, line 145: Every MsalException except authentication_canceled triggers legacy authentication. MSAL exposes authorization refusal separately as MsalError.AccessDenied (access_denied). If a user denies the first authorization request, GCM will immediately initiate another sign-in using the legacy application and describe the refusal as a reportable application failure. Exclude MsalError.AccessDenied as well, or restrict fallback to errors that can plausibly be resolved by changing client applications. Add a focused test alongside the new fallback tests.

src/Core/Authentication/Entra/EntraAuthentication.PublicClient.cs
```diff
@@ -459,9 +460,9 @@ private PublicClientApplicationBuilder GetPublicAppBuilder(out bool useBroker)
             if (Context.SessionManager.IsDesktopSession && IsBrokerEnabled())
```
KevinKruse: `himmelblau-broker` provides a broker on linux without desktop session. In combination with broker support on gcm on linux, the IsDesktopSession check could be removed (for himmelblau-brokered linux sessions at least?)

src/Microsoft.AzureRepos/AzureReposHostProvider.cs
mjcheetham: Good call .. `authentication_canceled` is not the only cancellation/denial outcome. There is `access_denied`, and device-code flow can return `authorization_declined`; there are also interaction-required, policy and throttling failures where switching client applications is not appropriate.

I think this is better to change this to warn rather than fall back automatically.

A helper now checks only recognised client-configuration failures from the new GCM app, and we just print guidance to retry explicitly with the legacy client. The original exception is then rethrown. This feels safer since it avoids users unknowingly relying on automatic fallback, and avoids repeating authentication when another call is not valid or useful.

src/Microsoft.AzureRepos/AzureReposHostProvider.cs
```diff
@@ -21,37 +22,35 @@ public class AzureReposHostProvider : DisposableObject, IHostProvider, IConfigur
         private readonly IAzureReposBindingManager _bindingManager;
         private readonly Lazy<IEntraAuthentication> _entraAuth;
 
-        public AzureReposHostProvider(ICommandContext context)
-            : this(context, new AzureDevOpsRestApi(context),
+        public AzureReposHostProvider(ICommandContext context) :
+            this(context, new AzureDevOpsRestApi(context), config => new EntraAuthentication(context, config),
                 new AzureDevOpsAuthorityCache(context), new AzureReposBindingManager(context))
         {
         }
```
mr-otmarg, line 29: This changes both collaborator-injection constructors that are `public` on `main` to `internal` (and removes the four-argument overload), so consumers compiled against `Microsoft.AzureRepos` can hit `MissingMethodException`, while source consumers and out-of-tree tests can no longer construct the provider with their own REST API/cache/binding/auth implementations. That compatibility break is unrelated to changing the Entra client. MIght want to retain the existing public overloads and have them delegate to the new internal factory overload; the factory itself can remain internal.

mr-otmarg (dismissed):
One compatibility issue remains: the authentication-factory refactor removes existing public `AzureReposHostProvider` constructor APIs. The new client selection and targeted failure guidance otherwise look well covered, and all 27 CI checks are green.

src/Microsoft.AzureRepos/AzureReposHostProvider.cs
mjcheetham: We don't offer any sort of API compatibility guarantees to those outside of the tree for the providers themselves. The 'Core' project's APIs we are more careful with since they are used by others that may be working on their own providers, but we still reserve the right to break changes here.

Plus this is targeting the first GCM 3.x release that is a major change - perfect time to make any behaviour or source breaking changes ;)

mr-otmarg (approved):

dscho (approved):
Very nice!

It probably makes sense to keep this change in mind in particular, once the time to write release notes for the next version comes around.

PR #2425, opened 2026-08-28
diagnose: improve result reporting and output

mjcheetham:
Update `git-credential-manager diagnose` to report structured diagnostic outcomes instead of collapsing each diagnostic into a true/false result and an unstructured log.

- Distinguish successful, warning, error, and skipped diagnostic outcomes.

- Use Spectre.Console to show live diagnostic status, colour-coded results, and a consolidated summary with generated log file paths.

- Add `--strict` so automated checks can treat warnings as a non-zero exit, while preserving the default behaviour of failing only on errors.

- Collect exceptions and additional files through the diagnostic reporter so presentation, logging, and file copying are handled consistently.

- Allow diagnostics to be registered explicitly, making the command infrastructure easier to test independently of the standard diagnostics.

- Add coverage for command exit behaviour and structured networking reports.

This ultimately forms a better base for expanding our diagnostic tests.

**Example Terminal Output**

image

image

### DIFF
--- a/src/Core/Commands/DiagnoseCommand.cs
+++ b/src/Core/Commands/DiagnoseCommand.cs
@@ -61,10 +61,6 @@ namespace GitCredentialManager.Commands
                 return 0;
             }
 
-            int numFailed = 0;
-            int numSkipped = 0;
-            int numWarned = 0;
-
             string currentDir = Directory.GetCurrentDirectory();
             string outputDir;
             if (string.IsNullOrWhiteSpace(output))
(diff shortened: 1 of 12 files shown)


src/Core/Commands/DiagnoseCommand.cs
```diff
@@ -99,13 +101,17 @@ private async Task<int> ExecuteAsync(string output)
...
+                    fullLog.Write("Outcome: Skipped");
+                    if (!string.IsNullOrWhiteSpace(skipReason))
+                    {
+                        fullLog.Write($" ({skipReason})");
+                    }
                     fullLog.WriteLine();
 
                     Console.Write(" ");
-                    ConsoleEx.WriteColor("[SKIP]", ConsoleColor.Gray);
+                    ConsoleEx.WriteColor("[SKIP]", ConsoleColor.DarkGray);
```
dscho, line 114: Not sure that this fits the bill of "diagnose: report structured outcomes" ;-) But it does not really matter, I think, it's not like this introduces a regression.

src/Core/Commands/DiagnoseCommand.cs
```diff
@@ -23,10 +23,13 @@ public DiagnoseCommand(ICommandContext context)
 
             _context = context;
 
-            var output = new Option<string>(new[] { "--output", "-o" }, "Output directory for diagnostic logs.");
+            var output = new Option<string>(["--output", "-o"], "Output directory for diagnostic logs.");
```
dscho, line 26: In general, I dislike those drive-by changes by AI, they increase the cognitive load for reviewers in unnecessary ways. But in this instance, it's a net readability improvement, so I don't mind _as_ much.

dscho (approved):
In particular the "sprucing up" of the output should make those among us happy who insist on Quarter-by-Quarter Incremental Delivery, i.e. those who demand that every initiative/epic must come with a user-visible improvement 😉

PR #2429, opened 2026-09-01
fix: Refactor Bitbucket Cloud's OAuth2 host provider to support rotating refresh tokens

### LINKED ISSUE #2428: An upcoming Bitbucket OAuth2 change will break the GCM on Windows when using the default wincredman credentialStore
Hi - I am an Atlassian engineer who works on Bitbucket Cloud's OAuth infrastructure.

An impending OAuth2 change in Bitbucket Cloud will cause the OAuth2 access tokens (and refresh tokens) generated by the client used by the GCM to regularly exceed the limits of the **default credentialStore** implementation on windows `wincredman`. This causes a complete failure during the `refresh_token` flow and initial `authorization_code` grant flows.

The error is as follows:
```
fatal: Failed to write item to store. [0x6f7]
fatal: The stub received bad data
```

The reason for the increase of token size is a change in our backend OAuth implementation. This change brings in rotating refresh tokens, and a swathe of other security uplifts. The downside is the tokens are slightly over the character limits imposed by the  `wincredman` which is 2,560 bytes.

This can be remedied by swapping to use the `dpapi` **credentialStore** which doesn't have such trivial limits on credential limits. This obviously requires a command line option to be run to swap, but is a pretty minimal change to keep OAuth working. I am not sure if swapping the `credentialStore` would copy across the `refresh_token` though, but I would hope there would be some chain of commands that will enable users to keep their authorizations.

An alternative could be splitting and storing the refresh / access tokens in smaller parts, stitching them back together. However this seems undesirable as it would require a version upgrade to continue using the GCM to auth.

Just wondering if any of the other OAuth providers have run into this - or if you have any other thoughts / remedies. We're currently exempting the OAuth client from the rollout. Another thing I found in my travels is that the GCM is re-using the Source Tree app OAuth client.

I'd love to change this over to use a unique OAuth client for GCM - but that's probably a problem for another day - and would require users to go through a full 3LO authorization flow.

Best regards,
Daniel Fraser

More details about the OAuth changes here: See: https://developer.atlassian.com/cloud/bitbucket/changelog/#CHANGE-3052)

danielrfraser:
Updates the `BitbucketHostProvider` to always store the refresh credentials using the resolved **username** from bitbucket.

## Background
Bitbucket Cloud's OAuth implementation is moving to rotating refresh tokens. These refresh tokens have a TTL of 3 months, but are immediately expired (~10 minutes) after their first usage. This means after each `refresh_token` grant flow it's expected that clients store the newly issued `refresh_token` from the OAuth token set.

Another change is that the `access_token` and `refresh_token` are far larger than the previous tokens. Regularly exceeding 2500 characters. This will cause issues with `CredentialStore` which have limits on the credential length. The default for windows `wincredman` is one such store which has issues. We will be asking customers using GCM on windows (likely via **git output** and a deprecation notice) that they'll need to change credential store that git credential manager is configured to use to `dpapi` which doesn't have the size constraints.

### Why does the Bitbucket OAuth code in GCM need changing?
When the initial `authorization_code` grant flow is run, the `access_token` and `refresh_token` returned are stored in the configured `CredentialStore` against the validated Bitbucket `username` (resolved by calling `api.bitbucket.org/2.0/user` with the `access_token`).

When the `access_token` stored expires, the `refresh_token` grant flow automatically runs to generate fresh credentials. This flow looks for a `refresh_token` in the Credential Store stored under the `account` provided by the `request.UserName` property.

In some cases this property is null (if the git remote URI configured specifies no username). When the account is null, the Credential Store will return back the first **Credential** that matches the git host. This will find the `refresh_token` stored by the original authorization flow. For git remotes configured with a `username` the CredentialManager will look for a secret belonging exactly to that account.

When, the `refresh_token` grant flow has concluded and the `access_token` has been validated the new `refresh_token` is stored in CredentialStore against the `account` provided by the `remoteUri.GetUserName()` function. However, because `remoteUri` is built with `includeUser=False` (which is the default) the `refresh_token` credential will always be stored against a `null` account in the credential store.

Because of the discrepency between the locations where we store the `refresh_token` and retrieve it, The `refresh_token` will only works the first time the `refresh_token` grant flow is executed. Subsequent `refresh_token` grant flows will retrieve the original (and now expired) `refresh_token` - which will be blocked. This causes the auth code will fall into a new authorization grant flow. The broken behaviour would then loop in this manner indefinitely requiring constant reauthorization.

### The change
This fix brings the `refresh_token` grant flow inline with the implementation of the `authorization_code` grant flow. This stores the `refresh_token` against the `account` which is the resolved Bitbucket `username`. This should mean that regardless of whether you've set your `username` in the git remote URI, the latest `refresh_token` will be used to initiate the `refresh_token` flow.

### Rollout
Once the version containing this release is shipped / bundled into Git for Windows, we will be announcing a deprecation and publicly documenting the steps required to continue authing using the OAuth credentials. We will likely run a series of brownouts against the older versions of GCM prior to the complete disablement of the non-rotating refresh tokens.

We will also highlight in our deprecation / announcement (In official Atlassian channels) that any Windows users need to move from using the default `CredentialStore` of `wincredman` onto `dpapi`. 

###
I sparred with the problem and fix with my colleague @mminns who worked on some of the original code some 8+ years ago.

Fixes: https://github.com/git-ecosystem/git-credential-manager/issues/2428

### DIFF
--- a/src/Atlassian.Bitbucket/BitbucketHostProvider.cs
+++ b/src/Atlassian.Bitbucket/BitbucketHostProvider.cs
@@ -139,8 +139,16 @@ namespace Atlassian.Bitbucket
             // Check for presence of refresh_token entry in credential store
             Uri remoteUri = request.GetRemoteUri();
             var refreshTokenService = GetRefreshTokenServiceName(remoteUri);
-
-            _context.Trace.WriteLine("Checking for refresh token...");
+            
+            _context.Trace.WriteLine($"Checking for refresh token stored against user: '{request.UserName}'...");
+            // request.UserName can either be be null or the registered Bitbucket username of the user attempting the
+            // git operation.
+            // * In the case of null the CredentialStore.Get will find the *first* refresh token stored for Bitbucket in
+            // the credential store - The storing of the refresh token is stored against the resolved Bitbucket username
+            // as part of the access token validation flow (calls 2.0/user)
+            // * When request.UserName is set the refresh token must exist under the provided username in the credential
+            // store.
+            // If a refresh token is unable to be found a full OAuth authorization flow is initiated.
             ICredential refreshToken = SupportsOAuth(authModes)
                 ? _context.CredentialStore.Get(refreshTokenService, request.UserName)
                 : null;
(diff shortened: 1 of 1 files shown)


LabTopProductions (commented):
@brecht

PR #2431, opened 2026-09-02
oauth: use native Base64Url encoder

becm:
Targeting `.Net 10+` allows use of native `System.Buffers.Text.Base64Url` instead of manual `Base64`/char-replace.
Padding has always been suppressed (with intent) in existing calls.


mjcheetham (approved):
Thank you!

PR #2433, opened 2026-09-02
NONE: Refactor Bitbucket Cloud's OAuth2 host provider to support rotating refresh tokens

### LINKED ISSUE #2428: An upcoming Bitbucket OAuth2 change will break the GCM on Windows when using the default wincredman credentialStore

danielrfraser-atlassian:
Updates the `BitbucketHostProvider` to always store the refresh credentials using the resolved **username** from bitbucket.

## Background
Bitbucket Cloud's OAuth implementation is moving to rotating refresh tokens. These refresh tokens have a TTL of 3 months, but are immediately expired (~10 minutes) after their first usage. This means after each `refresh_token` grant flow it's expected that clients store the newly issued `refresh_token` from the OAuth token set.

Another change is that the `access_token` and `refresh_token` are far larger than the previous tokens. Regularly exceeding 2500 characters. This will cause issues with `CredentialStore` which have limits on the credential length. The default for windows `wincredman` is one such store which has issues. We will be asking customers using GCM on windows (likely via **git output** and a deprecation notice) that they'll need to change credential store that git credential manager is configured to use to `dpapi` which doesn't have the size constraints.

### Why does the Bitbucket OAuth code in GCM need changing?
When the initial `authorization_code` grant flow is run, the `access_token` and `refresh_token` returned are stored in the configured `CredentialStore` against the validated Bitbucket `username` (resolved by calling `api.bitbucket.org/2.0/user` with the `access_token`).

When the `access_token` stored expires, the `refresh_token` grant flow automatically runs to generate fresh credentials. This flow looks for a `refresh_token` in the Credential Store stored under the `account` provided by the `request.UserName` property.

In some cases this property is null (if the git remote URI configured specifies no username). When the account is null, the Credential Store will return back the first **Credential** that matches the git host. This will find the `refresh_token` stored by the original authorization flow. For git remotes configured with a `username` the CredentialManager will look for a secret belonging exactly to that account.

When, the `refresh_token` grant flow has concluded and the `access_token` has been validated the new `refresh_token` is stored in CredentialStore against the `account` provided by the `remoteUri.GetUserName()` function. However, because `remoteUri` is built with `includeUser=False` (which is the default) the `refresh_token` credential will always be stored against a `null` account in the credential store.

Now on some `credentialStores` this will find the first credential configured (as the `account` is null and that's the behaviour documented by `AddOrUpdate`) however on file system based backends (Which we will be requiring windows users to move to with `dpapi`) it appears as though it will create a new file for the credential:
image
Notice from this image above the "latest" refresh token is store in a file called `.credential`. However, when the next git operation needs to auth, the `CredentialStore.get` appears to take the `<username>.credential` file over the account-less `.credential` file. This behaviour appears consistent between both Windows and MacOSX when using a file system based `CredentialStore` (tried both `dpapi` and `plaintext`). Note that if the user has specified a `username` in their remote git URI the backend will always return back the `<username>.credential` file and will not ever fall back to the `.credential` file - in this case a reauthorization flow would be triggered.

Because of the discrepency between the locations where we store the `refresh_token` and retrieve it, The `refresh_token` will only works the first time the `refresh_token` grant flow is executed. Subsequent `refresh_token` grant flows will retrieve the original (and now expired) `refresh_token` - which will be blocked. This causes the auth code will fall into a new authorization grant flow. The broken behaviour would then loop in this manner indefinitely requiring constant reauthorization.

### The change
This fix brings the `refresh_token` grant flow inline with the implementation of the `authorization_code` grant flow. This stores the `refresh_token` against the `account` which is the resolved Bitbucket `username`. This should mean that regardless of whether you've set your `username` in the git remote URI, the latest `refresh_token` will be used to initiate the `refresh_token` flow.

### Rollout
Once the version containing this release is shipped / bundled into Git for Windows, we will be announcing a deprecation and publicly documenting the steps required to continue authing using the OAuth credentials. We will likely run a series of brownouts against the older versions of GCM prior to the complete disablement of the non-rotating refresh tokens.

We will also highlight in our deprecation / announcement (In official Atlassian channels) that any Windows users need to move from using the default `CredentialStore` of `wincredman` onto `dpapi`. 

###
I sparred with the problem and fix with my colleague @mminns who worked on some of the original code some 8+ years ago.

Fixes: https://github.com/git-ecosystem/git-credential-manager/issues/2428

### DIFF
--- a/src/Atlassian.Bitbucket/BitbucketHostProvider.cs
+++ b/src/Atlassian.Bitbucket/BitbucketHostProvider.cs
@@ -139,8 +139,16 @@ namespace Atlassian.Bitbucket
             // Check for presence of refresh_token entry in credential store
             Uri remoteUri = request.GetRemoteUri();
             var refreshTokenService = GetRefreshTokenServiceName(remoteUri);
-
-            _context.Trace.WriteLine("Checking for refresh token...");
+            
+            _context.Trace.WriteLine($"Checking for refresh token stored against user: '{request.UserName}'...");
+            // request.UserName can either be be null or the registered Bitbucket username of the user attempting the
+            // git operation.
+            // * In the case of null the CredentialStore.Get will find the *first* refresh token stored for Bitbucket in
+            // the credential store - The storing of the refresh token is stored against the resolved Bitbucket username
+            // as part of the access token validation flow (calls 2.0/user)
+            // * When request.UserName is set the refresh token must exist under the provided username in the credential
+            // store.
+            // If a refresh token is unable to be found a full OAuth authorization flow is initiated.
             ICredential refreshToken = SupportsOAuth(authModes)
                 ? _context.CredentialStore.Get(refreshTokenService, request.UserName)
                 : null;
@@ -203,51 +211,39 @@ namespace Atlassian.Bitbucket
 
         private async Task<ICredential> GetOAuthCredentialsViaRefreshFlow(GitRequest request, ICredential refreshToken)
         {
-            Uri remoteUri = request.GetRemoteUri();
-
-            var refreshTokenService = GetRefreshTokenServiceName(remoteUri);
             _context.Trace.WriteLine("Refreshing OAuth credentials using refresh token...");
-
             OAuth2TokenResult refreshResult = await _bitbucketAuth.RefreshOAuthCredentialsAsync(request, refreshToken.Password);
-
-            // Resolve the username
-            _context.Trace.WriteLine("Resolving username for refreshed OAuth credential...");
-            string refreshUserName = await ResolveOAuthUserNameAsync(request, refreshResult.AccessToken);
-            _context.Trace.WriteLine($"Username for refreshed OAuth credential is '{refreshUserName}'");
-
-            // Store the refreshed RT
-            _context.Trace.WriteLine("Storing new refresh token...");
-            _context.CredentialStore.AddOrUpdate(refreshTokenService, remoteUri.GetUserName(), refreshResult.RefreshToken);
-
-            // Return new access token
-            return new GitCredential(refreshUserName, refreshResult.AccessToken);
+            return await ResolveCredsAndStoreRefreshToken(request, refreshResult);
         }
 
         private async Task<ICredential> GetOAuthCredentialsViaInteractiveBrowserFlow(GitRequest request)
         {
-            Uri remoteUri = request.GetRemoteUri();
-
-            var refreshTokenService = GetRefreshTokenServiceName(remoteUri);
-
             // We failed to use the refresh token either because it didn't exist, or because the refresh token is no
             // longer valid. Either way we must now try authenticating using OAuth interactively.
 
             // Start OAuth authentication flow
             _context.Trace.WriteLine("Starting OAuth authentication flow...");
             OAuth2TokenResult oauthResult = await _bitbucketAuth.CreateOAuthCredentialsAsync(request);
+            return await ResolveCredsAndStoreRefreshToken(request, oauthResult);
+        }
 
-            // Resolve the username
+        private async Task<ICredential> ResolveCredsAndStoreRefreshToken(GitRequest request, OAuth2TokenResult tokenSet)
+        {
+            Uri remoteUri = request.GetRemoteUri();
+
+            var refreshTokenService = GetRefreshTokenServiceName(remoteUri);
+            // Resolve the username to authenticate the git call
             _context.Trace.WriteLine("Resolving username for OAuth credential...");
-            string newUserName = await ResolveOAuthUserNameAsync(request, oauthResult.AccessToken);
-            _context.Trace.WriteLine($"Username for OAuth credential is '{newUserName}'");
+            string bitbucketUsername = await ResolveOAuthUserNameAsync(request, tokenSet.AccessToken);
+            _context.Trace.WriteLine($"Username for OAuth credential is '{bitbucketUsername}'");
 
-            // Store the new RT
-            _context.Trace.WriteLine("Storing new refresh token...");
-            _context.CredentialStore.AddOrUpdate(refreshTokenService, newUserName, oauthResult.RefreshToken);
+            // Store the new refresh token in the credential store against the resolved Bitbucket username
+            _context.Trace.WriteLine($"Storing new refresh token against user: '{bitbucketUsername}'...");
+            _context.CredentialStore.AddOrUpdate(refreshTokenService, bitbucketUsername, tokenSet.RefreshToken);    
             _context.Trace.WriteLine("Refresh token was successfully stored.");
 
             // Return the new AT as the credential
-            return new GitCredential(newUserName, oauthResult.AccessToken);
+            return new GitCredential(bitbucketUsername, tokenSet.AccessToken);
         }
 
         private static bool SupportsOAuth(AuthenticationModes authModes)
--- a/src/Atlassian.Bitbucket.Tests/BitbucketHostProviderTest.cs
+++ b/src/Atlassian.Bitbucket.Tests/BitbucketHostProviderTest.cs
@@ -155,7 +155,7 @@ namespace Atlassian.Bitbucket.Tests
                 MockDCSSOEnabled();
             }
             MockStoredAccount(context, request, token);
-            MockRemoteAccessTokenValid(request, token);
+            MockRemoteAccessTokenValid(username, token);
 
             var provider = new BitbucketHostProvider(context, bitbucketAuthentication.Object, MockRestApiRegistry(request, bitbucketApi).Object);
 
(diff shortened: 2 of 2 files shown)


mminns (commented):
A small change with limited impact, looks good to me.

becm:
The support for `authtype` in [Credential](/git-ecosystem/git-credential-manager/blob/main/src/Core/Credential.cs) is not present yet.
All requests _will_ use a `Basic` header combined of [username:token](https://github.com/danielrfraser-atlassian/git-credential-manager/blob/c4972b72a488c9ef783e778c4d955518d1595e4f/src/Core/Commands/GetCommand.cs#L99).
Is there a transition plan/period on BitBucket **server** to reject this variant and only accept `Bearer`?

DPAPI (as an implementation of `PlaintextCredentialStore`) has no convention of storing additional properties.
So the Provider will have to augment/replace the returned Credential to represent the correct token type.
→ general issue if token type is part of Credential (not supported on _primitive_ stores) or the Provider!

danielrfraser-atlassian:
@becm Bitbucket Cloud's credentials have not changed for either git operations (basic auth) or for our APIs (Bearer token). It's just that we're going to be generating far larger tokens (JWT now instead of opaque). This causes issues with Windows `wincredman`.

However - the fix in this PR is fixing a tangential issue related to our move to rotating refresh tokens. The fix ensures the `account` we're storing the `refresh_token` credential against is consistent between the Original authorization flow and refresh_token flow.

src/Atlassian.Bitbucket/BitbucketHostProvider.cs
```diff
@@ -203,51 +211,39 @@ private async Task<ICredential> GetRefreshedCredentials(GitRequest request, Auth
...
-            _context.Trace.WriteLine("Storing new refresh token...");
-            _context.CredentialStore.AddOrUpdate(refreshTokenService, newUserName, oauthResult.RefreshToken);
+            // Store the new refresh token in the credential store against the resolved Bitbucket username
+            _context.Trace.WriteLine($"Storing new refresh token against user: '{bitbucketUsername}'...");
+            _context.CredentialStore.AddOrUpdate(refreshTokenService, bitbucketUsername, tokenSet.RefreshToken);    
             _context.Trace.WriteLine("Refresh token was successfully stored.");
 
             // Return the new AT as the credential
-            return new GitCredential(newUserName, oauthResult.AccessToken);
+            return new GitCredential(bitbucketUsername, tokenSet.AccessToken);
```
becm, line 246: [Both](https://github.com/git-ecosystem/git-credential-manager/blob/c268a7b63687d53381d2b3e6a18f077127e8b73c/src/Atlassian.Bitbucket/BitbucketHostProvider.cs#L116) [tokens](https://github.com/git-ecosystem/git-credential-manager/blob/c268a7b63687d53381d2b3e6a18f077127e8b73c/src/Atlassian.Bitbucket/BitbucketHostProvider.cs#L144) are looked up with a username from `GitRequest` (URL or config) if set.
Irrelevant if not set → uses first stored credential for _service_. 
But can lead to (persistent) mismatch if _real_ BitBucket user name deviates.

_This is not a new issue, but would require an **additional** breaking change to adjust later._

danielrfraser-atlassian: Yeah - this is a subtle point which i'd missed. In the case of no `GitRequest` username being set, we'd store the `access_token` and `refresh_token` against `user_name1`. Retrieval would be fine, as there is only a singular credential stored. However, upon refresh, if the username changes (reported from the Bitbucket API call), we'd go to store the new refresh token (and access token) against `user_name2`.

At this point the next lookup will choose whatever the `CredentialStore` decides is first - which I'd guess would be deterministic, but in the case of a file system based store is likely lexiographical - which is a little undesirable.

With an out of date `GitRequest` username being set, we'd request the "old" credentials / refresh token which would cause a 401 with the refresh flow, lead to a full 3LO re-authorization flow, which would store against the "new" bitbucket registered username in the `CredentialStore`. But without the user updating their username we'd lead to this behaviour repeating.

But as you said - I think this is not a new issue. In most cases I'd imagine that users don't specify a mismatch (or rename their bitbucket username) and in both cases expecting them to either clean out their credentials / re-auth isn't exactly huge effort given how much of an edge case the scenario is?

mjcheetham: > But can lead to (persistent) mismatch if real BitBucket user name deviates.

> But as you said - I think this is not a new issue. In most cases I'd imagine that users don't specify a mismatch (or rename their bitbucket username) and in both cases expecting them to either clean out their credentials / re-auth isn't exactly huge effort given how much of an edge case the scenario is?

This is an gap in the credential protocol behaviour, and not a Bitbucket-specific or GCM-specific issue.

If Git gives a helper a `username` (from the remote URL userinfo) in a `get` request but the helper returns a _different_ username then the subsequent `store` (or `erase`) call will return that _new_ username and the helper be expected to store against that new username. **But now** there is nothing Git does to reconcile the mismatch meaning any following `get` requests will be for the original `username` from the remote URL userinfo.

This is nothing to be concerned about in this PR :smile:

mjcheetham (approved):
Thanks for the PR!

mjcheetham:
> DPAPI (as an implementation of PlaintextCredentialStore) has no convention of storing additional properties.

One of these days we should probably wire up extra metadata that we can (token type, expiry, etc) against credentials we store. The issue is that some credential stores do not support metadata.

The file-based store formats could be easily expanded to add extra attributes, as can the macOS Keychain (encoding as `key=value\n*` strings in the '[comment](https://developer.apple.com/documentation/security/ksecattrcomment)' attribute) and freedesktop.org Secret Service (as [generic attribute(s)](https://specifications.freedesktop.org/secret-service/latest/lookup-attributes.html)). Windows Credential Manager also [supports an `attributes` collection](https://learn.microsoft.com/en-us/windows/win32/api/wincred/ns-wincred-credentialw).

The ['cache' store](https://github.com/git-ecosystem/git-credential-manager/blob/main/docs/credstores.md#gits-built-in-credential-cache) (git-credential-cache) and the [passthrough pseudo-store](https://github.com/git-ecosystem/git-credential-manager/blob/main/docs/credstores.md#passthroughno-op-no-credential-store) however pose issues. The former could probably be taught to store and recall `state[]` attributes if given and we could use that.

**OR** we just store a sidecar of metadata in the `~/.gcm` directory.

danielrfraser-atlassian:
@mjcheetham thanks for the review.

I actually have a more in depth fix (in a branch on my fork ATM) to mean we don't need users to swap credential stores.

The access tokens (and refresh tokens) Bitbucket is about to move too are 2000+ characters long. I've made a change (isolated to the Bitbucket Host provider) to chunk the token when running against the Windows Credential Manager credential store.

That will mean that there is no user action required outside of upgrading the GCM version.

I was sparring with @mminns on it this evening, but didn't get a chance to raise the PR. The overlap with the UK is terrible from Australia 🦘 

I'll raise that tomorrow - that also relied on this fix, so probably better to have them separated regardless. That maybe more contentious a PR than this was though.

PR #2440, opened 2026-09-10
[Bitbucket Cloud] Implement credential chunking for the wincredman credentialStore to ensure seamless upgrade path for Bitbucket Cloud OAuth changes

danielrfraser-atlassian:
# Problem
As stated here: https://github.com/git-ecosystem/git-credential-manager/issues/2428

Bitbucket Cloud will start generating tokens which are in excess of 2,000 character wide strings. This causes problems for our Windows users which have the `wincredman` CredentialStore configured (the default for GCM on windows). Windows underlying credential manager is only able to store 2,560 bytes per secret. 

See: https://learn.microsoft.com/en-us/windows/win32/api/wincred/ns-wincred-credentiala

> CredentialBlobSize
> 
> The size, in bytes, of the CredentialBlob member. This member cannot be larger than CRED_MAX_CREDENTIAL_BLOB_SIZE (5*512) bytes.

This is a 1280 character long unicode string (Validated on my windows 10 laptop).

## Solutions to this Problem
After some back and forth, I decided there was was two options:
1. Force windows users to change their configured `credentialStore` to `dpapi`, so that there isn't a limitation on the size of tokens. This required user action, and meant that any other stored credentials would be lost, and would require re-entry / re-authorization (I know that it's possible to configure different credential stores for different hosts but still)

2. Implement some form of chunking algorithm that allows Windows users to continue to use the same default `CredentialStore` but would get around the size constraints on the Windows Credential Manager. This solution is massively preferable as it means the only action we'd need users to take for us to migrate the OAuth token format would will be to upgrade to the GCM version containing the changes.

# This PR
This PR implements the second option above. This implements the changes isolated within the `BitbucketHostProvider`. It will chunk both `access_tokens` and `refresh_tokens` which are larger than the available space in the underlying `CredentialStore` into the appropriate number of chunks. 

The implementation defines a new overridable property on the `ICredentialStore` which defines the `CredentialStores` maximum credential (when stored as a string) size. The default value for this is `0` which means that the `CredentialStore` does not impose any restrictions. The only `CredentialStore` implementation which sets a max size is the `WindowsCredentialManager`.

## The chunking implementation
The actual chunking implementation is relatively trivial and maintains backwards compatibility for existing stored credentials. This allows backwards compatibility with existing stored secrets.

Consider the following credential to store (length 13)

```
service: bitbucket.org
Account: danielrfraser
Password: a-01234567890
```
Normally this would be stored like this:
```
bitbucket.org/danielrfraser => a-01234567890
```
But let's say we wish to store it in a `CredentialStore` with a limit of **5 characters per credential**. We'd have to break it into 3 chunks (2 complete and 1 partial) which would be stored like this:
```
bitbucket.org/chunks/danielrfraser_0 => a-012
bitbucket.org/chunks/danielrfraser_1 => 34567
bitbucket.org/chunks/danielrfraser_2 => 890
bitbucket.org/danielrfraser => chunks=3
```
We store a chunk descriptor (for lack of better name) against the `account` in the `service`. This means that when we read an existing secret for a given `account`, we would either a **chunk descriptor** or a **regular secret** (which is not chunked).

We can then use regex to detect if the credential is "chunked" or a regular secret. The format of the chunk descriptor: `chunks=<NUMBER_OF_CHUNKS>` is impossible to collide with a real Bitbucket issued PAT or OAuth token - `=` is not an allowed character in our token set, and the width is smaller than we'd ever mint.

If we encounter a chunked descriptor (on Get) we then use the stored credential's `Account`, coupled with the extracted `chunk count` to retrieve the `n` chunks and stitch the full secret together.

Note: If we encounter a missing chunk, we return `null` which would treat it the same as if there was no credential (or an expired credential) and the auth behaviour would continue on.

## Why this solution is good?
* It functions exactly the same as it used to for all other distros / backends.
* It also only chunks if it needs to, so for PAT authentication (even on Windows), the chunking code is not exercised as the tokens are not wide enough to trigger the chunking mechanism.
* It doesn't require any user intervention (outside of updating to this version of GCM) to support our new Access token / refresh tokens.
* It's contained to where it's needed (for now)

# What could be better?
* Potentially this chunking could be lifted (and hidden) into the underlying `WindowsCredentialManager` implementation. This would mean all `HostProviders` could use the behaviour. This seemed like a fair chunk more effort than I can realistically spend. If there is a long term desire that this could be the direction we need to go - we should ensure that we're happy with the "shape" of the implementation so that a future change could be rolled out without impacting the Bitbucket stored secrets.

# Testing
* Tested the build (using a different OAuth client which generates large access tokens / refresh tokens) on Windows 10 / MacOS

### DIFF
--- a/src/Atlassian.Bitbucket/BitbucketHostProvider.cs
+++ b/src/Atlassian.Bitbucket/BitbucketHostProvider.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Net.Http;
+using System.Text;
 using System.Threading.Tasks;
 using Atlassian.Bitbucket.Cloud;
 using GitCredentialManager;
@@ -10,13 +11,16 @@ using System.Text.RegularExpressions;
 
 namespace Atlassian.Bitbucket
 {
-    public class BitbucketHostProvider : IHostProvider
+    public partial class BitbucketHostProvider : IHostProvider
     {
         private readonly ICommandContext _context;
         private readonly IBitbucketAuthentication _bitbucketAuth;
         private readonly IRegistry<IBitbucketRestApi> _restApiRegistry;
-        private const string ChunkRegex = @"^chunks=(?'count'\d+)$";
-
+        
+        private const string ChunkRegexPattern = @"^chunks=(?'count'\d+)$";
+        [GeneratedRegex(ChunkRegexPattern)]
+        private static partial Regex ChunkRegex();
+        
         public BitbucketHostProvider(ICommandContext context)
             : this(context, new BitbucketAuthentication(context), new BitbucketRestApiRegistry(context)) { }
 
@@ -449,17 +453,15 @@ namespace Atlassian.Bitbucket
 
         private void AddOrUpdateCredential(string service, string account, string secret)
         {
-            var chunkConfig = _context.CredentialStore.MaxCredentialSize;
-            
-            if (chunkConfig.HasValue && secret?.Length > chunkConfig.Value)
+            if (SecretRequiresChunking(secret, out var chunkSize))
             {
                 var chunks = secret
-                    .Chunk(chunkConfig.Value)
+                    .Chunk((int)chunkSize)
                     .Select(x => new string(x)).ToList();
                 var chunkCount = chunks.Count;
                 _context.Trace.WriteLine(
-                    $"Storing credential of length {secret.Length} in {chunkCount} chunks for service: {service} and " +
-                    $"account: {account}"
+                    $"Storing credential of length {secret.Length} in {chunkCount} chunks of size {chunkSize} for " +
+                    $"service: {service} and account: {account}"
                 );
                 // We're storing the chunks in a separate service (nested under the provided service). This ensures that
                 // retrieving credentials for the provided service cannot trip up on the chunks when looking up with a
@@ -490,20 +492,32 @@ namespace Atlassian.Bitbucket
             }
         }
 
-        private int? GetChunkCount(ICredential credential)
+        private bool SecretRequiresDechunking(String secret, out int chunkCount)
         {
-            if (_context.CredentialStore.SupportsChunking() && credential.Password is not null)
+            var chunkSize = _context.CredentialStore.MaxCredentialSize;
+            // If the credentialStore has a valid configured chunk size, we have to support reading a chunked secret.
+            if (chunkSize > 0 && secret is not null)
             {
-                var isChunked = Regex.Match(credential.Password, ChunkRegex);
+                var isChunked = ChunkRegex().Match(secret);
                 if (isChunked.Success)
                 {
-                    return int.Parse(isChunked.Groups["count"].Value);
+                    chunkCount = int.Parse(isChunked.Groups["count"].Value);
+                    return true;
                 }
             }
             // If the credentialStore does not support chunking, or the secret wasn't a chunk descriptor, we are
             // returning null to indicate we shouldn't read using the chunking implementation
-            return null;
+            chunkCount = 0;
+            return false;
         }
+        
+        private bool SecretRequiresChunking(string secret, out int chunkSize)
+        {
+            // If a secret is bigger than the credentialStore's valid configured chunk size - we have to chunk it.
+            chunkSize = _context.CredentialStore.MaxCredentialSize;
+            return chunkSize > 0 && secret?.Length > chunkSize;
+        }
+        
         
         private ICredential GetCredential(string service, string account)
         {   
@@ -512,11 +526,9 @@ namespace Atlassian.Bitbucket
             {
                 return null;
             }
-
-            var optionalChunkCount = GetChunkCount(credential);
-            if (optionalChunkCount.HasValue)
+            
+            if (SecretRequiresDechunking(credential.Password, out var chunkCount))
             {
-                var chunkCount = optionalChunkCount.Value;
                 // We will be using the Account stored in the located credential to build the chunk identifiers and
                 // return the stitched together final Credential.
                 var credentialAccount = credential.Account;
@@ -525,7 +537,7 @@ namespace Atlassian.Bitbucket
                     $" stored against account: {credentialAccount}"
                 );
                 
-                var chunkedPassword = "";
+                var chunkedPasswordBuilder = new StringBuilder();
                 var chunkedService = GetChunkServiceName(service);
                 for (var index = 0; index < chunkCount; index++)
                 {  
@@ -540,10 +552,10 @@ namespace Atlassian.Bitbucket
                         $"Found chunk for service: {chunkedService} and account: {chunkAccount}: {{0}}", 
                         new object[] { chunk.Password }
                     );
-                    chunkedPassword += chunk.Password;
+                    chunkedPasswordBuilder.Append(chunk.Password);
                 }
                 // We've stitched together all the chunks into a singular secret - return it with the original account.
-                return new GitCredential(credentialAccount, chunkedPassword);
+                return new GitCredential(credentialAccount, chunkedPasswordBuilder.ToString());
             }
             _context.Trace.WriteLine(
                 $"Found non-chunked credential for service: {service} and account: {account} stored against " +
@@ -588,13 +600,6 @@ namespace Atlassian.Bitbucket
         {
             _restApiRegistry.Dispose();
             _bitbucketAuth.Dispose();
-        }
-    }
-    static class CredentialStoreExtensions
-    {
-        public static bool SupportsChunking(this ICredentialStore store)
-        {
-            return store.MaxCredentialSize.HasValue;
         }
     }
 }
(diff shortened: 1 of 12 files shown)


src/Atlassian.Bitbucket/BitbucketHostProvider.cs
```diff
@@ -6,7 +6,8 @@
 using Atlassian.Bitbucket.Cloud;
 using GitCredentialManager;
 using GitCredentialManager.Authentication.OAuth;
-using Spectre.Console;
+using System.Text.RegularExpressions;
+using GitCredentialManager.Interop.Windows;
```
joshua-tang-atlas: Don't think we need this anymore?

src/Atlassian.Bitbucket/BitbucketHostProvider.cs
```diff
@@ -448,7 +448,126 @@ private async Task<bool> ValidateCredentialsWork(GitRequest request, ICredential
...
+                }
+                // When using the chunking implementation we store a chunk descriptor against the account in the usual
+                // service. This allows us to identify (on get) that we've stored chunks, and the number of chunks to
+                // query.
+                _context.CredentialStore.AddOrUpdate(service, account, GetChunkSecret(chunkCount));
+            }
+            else
+            {
+                _context.Trace.WriteLine(
+                    $"Storing credential of length {(secret ?? "").Length} for service: {service} and " +
```
joshua-tang-atlas, line 486: Is `secret` nullable? Should we just early return without doing anything in the function?

src/Atlassian.Bitbucket/BitbucketHostProvider.cs
```diff
@@ -448,7 +448,126 @@ private async Task<bool> ValidateCredentialsWork(GitRequest request, ICredential
...
+                );
+                
+                var chunkedPassword = "";
+                var chunkedService = GetChunkServiceName(service);
+                for (var index = 0; index < chunkCount; index++)
+                {  
+                    // Build the chunk identifier using the username stored in the chunk descriptor we looked up.
+                    var chunkAccount = GetChunkAccount(credentialAccount, index);
+                    var chunk = _context.CredentialStore.Get(chunkedService, chunkAccount);
+                    if(chunk == null){
```
joshua-tang-atlas: nit:

```suggestion
                    if (chunk == null){
```

src/Atlassian.Bitbucket/BitbucketHostProvider.cs
danielrfraser-atlassian: There was some tests which had a nullable here. I can also just drop the length from the message. The old code didn't handle null at all. The secret store does "something" with it, without breaking tests at least.

danielrfraser-atlassian:
@mjcheetham  - Out of question - what would the next version of GCM (if this PR gets merged and released?)

src/Core/ICredentialStore.cs
```diff
@@ -11,7 +11,13 @@ public interface ICredentialStore
         /// Get the name of the credential store.
         /// </summary>
         string Name { get; }
-
+        
+        /// <summary>
+        /// Gets the maximum width of a secret that can be stored by this CredentialStore.
+        /// A value of null (the default) indicates there is no limits on the size of credential storage
+        /// </summary>
+        public int? MaxCredentialSize => null;
```
mjcheetham: Can we clarify in the xmldocs here if this is a byte count, or a Unicode/wstr/UTF-16 character count?

Also, I think we should use negative values here to signify 'unlimited' rather than `null`, to save all the awkward `.HasValue` checks vs a `>= 0` (or `<0` inverse). Default to `-1`, and update the XML doc to explain this.

```suggestion
        /// <summary>
        /// Gets the maximum size of a secret that can be stored by this CredentialStore.
        /// A negative value indicates there is no limits on the size of credential storage.
        /// </summary>
        /// <remarks>
        /// The size represented here is the number of characters (<see langword="char"/>) of the string.
        /// </remarks>
        public int MaxCredentialSize => -1;
```

src/Atlassian.Bitbucket/BitbucketHostProvider.cs
```diff
@@ -472,4 +590,11 @@ public void Dispose()
             _bitbucketAuth.Dispose();
         }
     }
+    static class CredentialStoreExtensions
+    {
+        public static bool SupportsChunking(this ICredentialStore store)
+        {
+            return store.MaxCredentialSize.HasValue;
+        }
```
mjcheetham: nit: isn't this more aptly named `RequiresChunking`? Any store _could_ be used to chunk if you wanted.

src/Atlassian.Bitbucket/BitbucketHostProvider.cs
```diff
@@ -448,7 +447,126 @@ private async Task<bool> ValidateCredentialsWork(GitRequest request, ICredential
...
+                    $"Storing credential of length {(secret ?? "").Length} for service: {service} and " +
+                    $"account: {account}"
+                );
+                // In the case we only have a single chunk (which will be the case on all credential managers except
+                // `wincredman`) we can store the secret as the old code used to.
+                _context.CredentialStore.AddOrUpdate(service, account, secret);
+            }
+        }
+
+        private int? GetChunkCount(ICredential credential)
```
mjcheetham: Could make this a bool+out style method:


```suggestion
        private bool RequiresChunking(ICredential credential, out int chunkCount)
```

```csharp
if (RequiresChunking(credential, out int chunkCount))
{
    // ..
}
else
{
    // ..
}
```

src/Atlassian.Bitbucket/BitbucketHostProvider.cs
```diff
@@ -15,6 +15,7 @@ public class BitbucketHostProvider : IHostProvider
         private readonly ICommandContext _context;
         private readonly IBitbucketAuthentication _bitbucketAuth;
         private readonly IRegistry<IBitbucketRestApi> _restApiRegistry;
+        private const string ChunkRegex = @"^chunks=(?'count'\d+)$";
```
mjcheetham: To optimise (since we now publish Native AOT with GCM 3.x) we could use [source generated regex](https://learn.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-source-generators#source-generation) here:

```csharp
public partial class BitbucketHostProvider
{
    private const string ChunkRegexPattern = @"^chunks=(?'count'\d+)$";

    [GeneratedRegex(ChunkRegexPattern)]
    private static partial Regex ChunkRegex();

    example_usage()
    {
        var chunkMatch = ChunkRegex().Match(credential.Password);
        if (chunkMatch.IsSuccess)
        {
            return int.Parse(chunkMatch.Groups["count"].Value);
        }
    }

}
```

src/TestInfrastructure/Objects/TestCredentialStore.cs
```diff
@@ -15,6 +15,11 @@ public TestCredentialStore()
         #region ICredentialStore
 
         public string Name => "Test Credential Store";
+        
+        public int? MaxCredentialSize
+        {
+            get;set;
+        }
```
mjcheetham: After switching to use `int` and negative values for 'infinite':

```suggestion
        public int MaxCredentialSize { get; set; } = -1;
```

src/Atlassian.Bitbucket/BitbucketHostProvider.cs
```diff
@@ -448,7 +447,126 @@ private async Task<bool> ValidateCredentialsWork(GitRequest request, ICredential
...
+                    var chunk = _context.CredentialStore.Get(chunkedService, chunkAccount);
+                    if (chunk == null){
+                        _context.Trace.WriteLine($"Chunk {index} was unexpectedly null");
+                        return null;
+                    }
+                    _context.Trace.WriteLineSecrets(
+                        $"Found chunk for service: {chunkedService} and account: {chunkAccount}: {{0}}", 
+                        new object[] { chunk.Password }
+                    );
+                    chunkedPassword += chunk.Password;
```
mjcheetham: To avoid reallocating all the time, could we use a `StringBuilder` instead here?

```csharp
var chunkedPassword = new StringBuilder();

..

chunkedPassword.Append(chunk.Password);

..

return new GitCredential(credentialAccount, chunkedPassword.ToString());
```

mjcheetham (changes requested):
Thanks for the contribution and support! Overall this looks like a reasonable solution to a sad limitation of Windows 😢 

I agree that chunking should probably _not_ live in the core implementation of the credential stores - this will keep things scoped and safer (as you said, "=" is not a valid Bitbucket account name so you _can_ rely on the `chunks=N` pattern, whereas in general this is not true).

I've left a few comments for some minor tweaks.

src/Atlassian.Bitbucket/BitbucketHostProvider.cs
danielrfraser-atlassian: Ah - this is a bit like a walrus operator in python. Sorry this is my first time coding in C# - so learning lots of new things :D

mjcheetham (approved):
Thanks for addressing my feedback!

danielrfraser-atlassian:
@mjcheetham thanks for the feedback and review What's the release process look like for GCM?

mjcheetham (approved):

mjcheetham:
> What's the release process look like for GCM?

We aim to make releases approximately in line with the Git project, which is every 12 weeks.

Git 2.56.0 is currently at RC0, so we're right around the point where a new GCM release is pending to be cut and go out. However we are currently in a transition point to release GCM 3.0, which means the next GCM may be a little offset. I hope in the next week or two we can get a release out.

PR #1956, opened 2025-06-14
Support generic OAuth for non-HTTP hosts

pl4nty:
Non-HTTP hosts can support OAuth using absolute `authorize` and `token` endpoints, like EXO/Outlook.com SMTP. Example:

```
[sendemail]
smtpServer = smtp-mail.outlook.com
smtpUser = example@outlook.com
smtpEncryption = tls
smtpServerPort = 587
smtpAuth = XOAUTH2

[credential "smtp://smtp-mail.outlook.com:587"]
oauthClientId = a95fbac5-cbbb-4032-b10d-fb756c457ba1
oauthAuthorizeEndpoint = https://login.microsoftonline.com/common/oauth2/v2.0/authorize
oauthTokenEndpoint = https://login.microsoftonline.com/common/oauth2/v2.0/token
oauthScopes = offline_access https://outlook.office.com/SMTP.Send
```

PR #2445, opened 2026-09-15
generic: permit any protocol for generic OAuth config

mjcheetham:
We currently do not consider either Windows Integrated or generic OAuth authentication in the generic provider, for a non-HTTP based remote URL.

Since GCM can be called as a credential helper by Git for things such as an SMTP server (see `git sendmail`), where the SMTP server may use OAuth authentication, we should relax this restriction.

Note that we still require HTTP(S) remotes for Windows Integrated auth since the probe for that requires an HTTP(S) call.

Replaces #1956

> [!TIP]
> Ignore whitespace when reviewing the diff in the `GenericHostProvider.cs` file - we are adding a wrapping `if () { }` block to existing code.


dscho (approved):
LGTM!

PR #2449, opened 2026-09-17
Switch to using OAuth tokens (Azure Repos) and the MS auth broker by default

mjcheetham:
Update the defaults for the Azure Repos credential type and Microsoft authentication broker enabled to `oauth` and `true` respectively. With the new GCM major version bump to 3.x, along with the new Entra 1st party application and Mac & Linux broker support, we are ready to adopt these by default.

The settings remain for those that need to revert back to `pat` and broker-less (`false`) auth.

Finally we update the documentation for the broker with new images, behaviour, and updated sections for Mac and Linux:

[Rendered broker docs](https://github.com/mjcheetham/git-credential-manager/blob/azrepos-newauth-default/docs/msauth-broker.md)


dscho (approved):
Looks great, in particular the _extensive_ documentation!

PR #2451, opened 2026-09-18
dispatcher: fix macOS UI and broker auth hangs

mjcheetham:
Supersedes [git-ecosystem/git-credential-manager#2448][previous-pr]. This version uses a single platform work queue and includes the execution-context and Trace2 changes as separate, reviewable commits.

On macOS, GCM hangs if it shows UI before an interactive Entra sign-in that uses the broker. A credential prompt followed by broker authentication can leave Git waiting on a process that never finishes. Fixing that ordering must not break the reverse ordering, or make every GCM invocation pay to start a UI framework.

## Why it happens

GCM runs the application on an `AppMain` thread and keeps the process entry thread available for platform APIs that need it. Two features depend on the main thread and its main loop, for different reasons:

- macOS requires UI controls to be created on the entry thread.
- The macOS MSAL broker path depends on whether an `NSApplication` is running.

Avalonia was started by posting its main loop to GCM's dispatcher as a job. That job never returns until shutdown, so the dispatcher's own queue stops being drained as soon as a window is shown. A subsequent broker call marshalled to the main thread queues behind work that will never finish.

There is a second half to the problem. MSAL decides **once per process** whether it is a console application, and caches the answer:

```csharp
// DesktopOsHelper.cs, MSAL 4.85.2
private static readonly Lazy<bool> _isMacConsoleApp =
    new Lazy<bool>(() => !LibObjc.IsNsApplicationRunning());
```

Without a running `NSApplication`, MSAL requires interactive broker calls to run on managed thread 1 and takes that thread over with its own polling loop. That cannot coexist with the UI main loop. With `NSApplication` running, it delegates threading to the broker instead.

Fixing queue starvation alone is therefore not enough. If the broker first sees a console application, starting UI later does not change that cached decision. The "broker, then UI, then broker again" ordering would still freeze the UI.

## The fix

The dispatcher owns the platform main loop instead of treating it as a job. It starts that loop **lazily, on the first request for main-thread work**. Avalonia's startup and pumping move behind an `IMainLoop` contract, implemented by `AvaloniaMainLoop`.

On successful startup, posting blocks until `IMainLoop.Initialize()` has made the platform queue available. Each caller then hands its own job directly to that queue. There is no dispatcher-owned pending-work queue to drain or transfer during startup.

The distinction between accepting and executing work matters here: initialization makes posting safe, but a queued job cannot execute until `IMainLoop.Run()` is pumping. On macOS, that means the job runs with `NSApplication` already up. The first interactive broker call consequently sees a GUI application, regardless of whether UI or broker authentication was requested first. Callers do not need to remember an explicit startup step.

Two one-shot gates coordinate this: `_workRequested` wakes the parked main thread, and `_mainLoopReady` releases callers waiting to post. Lifecycle, failure, and outstanding-job state are protected separately by `System.Threading.Lock`. Shutdown opens both gates, and failure also releases callers waiting for readiness. A caller must inspect the outcome rather than assume that waking means initialization succeeded.

**The common fast path still avoids starting Avalonia.** An invocation that posts no main-thread work parks and shuts down without initializing the UI framework. Silent and default-account authentication attempts deliberately remain off the dispatcher: they do not make MSAL's interactive broker console/GUI decision, so starting Avalonia for them would buy nothing.

## Context and tracing

Handing a job to another thread should not discard the posting caller's ambient state. The dispatcher captures and restores that caller's `ExecutionContext`, preserving values such as `AsyncLocal<T>` and `Activity.Current`.

Trace2 deliberately needs different attribution. The entry thread now starts the main loop and emits events of its own, so `AppMain` gets a distinct Trace2 thread context. Dispatched work uses the dispatcher's captured Trace2 context, applied **inside** the restored execution context so it is not overwritten by the caller's `AsyncLocal` state. The caller is also recorded as a data event to preserve the link between the two.

This makes the startup and main-thread work distinguishable from the application work that requested it, rather than reporting both as the same thread or losing the caller's other ambient state.

## Series

The 12 commits are arranged for commit-by-commit review:

| Commits | Purpose |
| --- | --- |
| 1 | Fold the untyped job into the generic implementation so the following fixes have one execution path to maintain. |
| 2-6 | Fix job exception propagation, inline task continuations, premature completion of asynchronous work, shutdown before `Run()`, and the distinction between stopping and stopped. |
| 7 | Introduce lazy main-loop ownership, the single-queue posting model, startup gates, and failure handling. |
| 8 | Correct the explanation of the macOS broker requirement and its cached console/GUI decision. |
| 9-11 | Separate the application's Trace2 context, expose a context handle, and preserve the caller's execution context while attributing dispatched work correctly. |
| 12 | Document the dispatcher design, thread interactions, lifecycle, and caller rules. |

The preparatory fixes address existing dispatcher problems, rather than introducing them as part of the main-loop change and repairing them later. In particular, a job that throws must not leave its caller waiting forever, and an asynchronous delegate must return a task representing the whole operation, not just the part before its first `await`.

## Coverage

The dispatcher cases in `src/Core.Tests/UI/DispatcherTests.cs` use a fake main loop so startup, failure, and shutdown can be controlled without initializing Avalonia or requiring a real process entry thread. They cover:

- Work requested before `Run()`, concurrent first callers, and later or re-entrant posts once initialization completes.
- Accepting a job before the main loop pumps, without executing it early.
- Initialization, main-loop, and posting failures, including preserving the first failure and skipping callbacks whose jobs have already been faulted.
- Shutdown before `Run()`, with no work, during initialization, while running, and with work still pending.
- Preserving the posting caller's ambient state on both cold and warm paths.

These cases exercise the dispatcher contract; they do not stand in for the native UI/broker integration on macOS.

## Notes for reviewers

> [!TIP]
> Read the [dispatcher design guide][dispatcher-guide] first. It includes diagrams of the thread interaction, posting path, startup, and lifecycle.

> [!IMPORTANT]
> The MSAL behavior described here is version-specific. The broker comment is based on MSAL 4.85.2; `DesktopOsHelper.IsMacConsoleApp` and the paths that evaluate it should be revisited on an MSAL upgrade.

- **Posting can block during startup**, including calls named `InvokeAsync`. It is the work's completion that is asynchronous; callers must first wait for a queue that can accept it.
- **Shutdown abandons pending work rather than draining it.** The application must await anything it cares about before shutting the dispatcher down. Draining would hang when a window is left open with nobody awaiting or closing it.
- **The motivating failure is macOS-specific, but the shared changes are cross-platform.** Windows and Linux broker calls still do not marshal through this dispatcher; the dispatcher fixes, UI main-loop ownership, and Trace2 attribution changes also apply on those platforms.

[previous-pr]: https://github.com/git-ecosystem/git-credential-manager/pull/2448
[dispatcher-guide]: https://github.com/mjcheetham/git-credential-manager/blob/dispatcher-v2/docs/dispatcher.md


dscho (approved):
Wow, what a big one. It looks good to me, and I think it will dramatically improve the user experience as well as make debugging easier (with traces that allow following the call path). Great job!

